PR #133, opened 2014-03-19
ENH: Grid object

kirknorth:
Adding the ability to exclude fields during the creation of a `Grid` object, as well as the ability to add a new field with the `add_field` attribute similar to the `Radar` object. Furthermore, I am adding some extra axes and dimensions to check for when creating a Grid object. These axes and dimensions are compatible with older versions of MMCG files. Finally, this corrects a bug during the creation of a Grid when a field has the wrong shape.

### DIFF
--- a/pyart/graph/gridmapdisplay.py
+++ b/pyart/graph/gridmapdisplay.py
@@ -29,6 +29,7 @@ class GridMapDisplay:
     grid : Grid
         Grid with data which will be used to create plots.
     proj : str
+        The projection to be used.
     datum : str
     debug : bool
         True to print debugging information.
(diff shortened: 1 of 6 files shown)


jjhelmus:
Is there still interesting in keeping this PR open?  We are still missing a **GridDisplay** class and a de-speckeling function would be useful.  I can make issues for both of these if there is interest.

kirknorth:
@jjhelmus seeing as there's been lots of discussion recently about remodelling the `Grid` class, I think it's safe to close this PR but at the same time recognizing that a salt/pepper noise removal function (despeckling) can be useful for radar grids. As of writing this I don't have a nice packaged function that does this, but one way to do it (I think we both mentioned in a previous email) is to use `ndimage.label` and remove labels (features) that are "small".

jjhelmus:
I created issue #345 to track work and discussion on grid and radar despeckling routines.  Closing this PR.

PR #158, opened 2014-05-29
ENH: objective analysis

kirknorth:
This is my initial attempt at enhancing certain components of the mapping routine. The improvements in this PR targets two areas:

1. A non-uniform grid can now specified, e.g. a grid that has higher resolution in the ABL and thins out aloft. This was an easy fix, and involved removing the `grid_shape` and `grid_limits` variables and allowing a user to input his/her own `grid_dimensions` variable.
2. Changing the Barnes weight to reflect the literature, primarily Trapp and Doswell (2000). The free parameter for a Barnes weight is the smoothing parameter, not the radius of influence like that of the Cressman weight.

As of this PR, only a constant smoothing parameter can be specified for the Barnes weight. This smoothing parameter depends on the non-dimensional smoothing parameter and the data spacing. They argue that "in terms of preservation of the phase and amplitude of the input data, predictability of the resultant smoothing and filtering, and relative insensitivity to input data unsteadiness or spatial characteristics, the isotropic Barnes weight function with constant smoothing parameter appears to be the most desirable of the schemes considered." See Trapp and Doswell (2000) for more information.

### DIFF
--- a/pyart/map/grid_mapper.py
+++ b/pyart/map/grid_mapper.py
@@ -435,7 +435,7 @@ def map_to_grid(radars, grid_shape, grid_dimensions, grid_origin=None,
     # create arrays to hold the gate locations and indicators if the gate
     # should be included in the interpolation
     gate_locations = np.empty((total_gates, 3), dtype=np.float64)
-    include_gate = np.ones((total_gates, 1), dtype=np.bool)
+    include_gate = np.ones(total_gates, dtype=np.bool)
     
     # offsets from the grid origin in meters for each radar
     offsets = []
--- a/pyart/map/grid_mapper.py
+++ b/pyart/map/grid_mapper.py
@@ -37,21 +37,21 @@ from .ckdtree import cKDTree
 from .ball_tree import BallTree
 
 
-def grid_from_radars(radars, grid_shape, grid_limits, **kwargs):
+def grid_from_radars(radars, grid_shape, grid_dimensions, **kwargs):
     """
-    Map one or more radars to a Cartesian grid returning a Grid object.
+    Map one or more radars to a Cartesian grid, returning a Grid object.
 
     Additional arguments are passed to :py:func:`map_to_grid`
 
     Parameters
     ----------
-    radars : tuple of Radar objects.
-        Radar objects which will be mapped to the Cartesian grid.
-    grid_shape : 3-tuple of floats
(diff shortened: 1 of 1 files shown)


jjhelmus:
@kirknorth A lot has changed in Py-ART since this was submitted which makes merging this difficult.  Sorry I did not get to this earlier.

Looking at the code, I'm seeing three major changes that this PR introduces.

1. `grid_from_radars` can produce non-uniform grids.
2. Trapp and Doswell (2000) weighting.
3. Use pyproj to determine a more accurate offset between radars.

I think, 1 needs to wait until the **Grid** object is refactored to support both uniform and inon-uniform grids.  Being able to generate non-uniform grids would be great but if you cannot do anything with them... then not so great.

Number 2 should get included in Py-ART in both the `map_to_grid` and `map_gates_to_grid` function.  Perhaps having three build in weighting functions (Cressman, "Traditional" Barnes, and "Trapp and Doswell" Barnes) would be reasonable.

Using pyproj for better mapping between lat/lon and Cartesian space is an big bug in Py-ART and should be fixed, but I think the issues needs to be address in a more comprehensive manner than is done here.  This is especially true if we wish to keep basemap and pyproj as optional dependencies.  Issue #294 for more discussion on this topic.  

My recommendation would be that two issues get created which refer to this PR; one to add Trapp and Doswell type weighting to the gridding methods.  The second requesting support the creation and use of non-uniform grid.  Issue #294 already covers the projection issues.  With these two issues created I think this PR can be closed.  Does that sound reasonable?

kirknorth:
Yup sounds reasonable. I would just add that as it stands now the `Grid` object does support non-uniform grids, it's just the mapping routines currently do not. However, non-uniform grids are basically a trivial addition when viewed from a k-d tree standpoint. Finally, I agree that we need more comprehensive, robust projection methods, but my addition of using `pyproj` in order to more accurately place (project) a radar's location within the grid is a valid one. The bigger issue is in properly projecting the radar polar coordinates onto the grid itself, where we also have to deal with atmospheric refraction. Ultimately though there needs to be consistency between these two projections. @gamaanderson has already brought up several of these issues.

jjhelmus:
Issues #349  and #350 created to track work on Trapp and Doswell weighting and non-uniform grids.

Closing this PR.  Thanks @kirknorth for this work!

PR #191, opened 2014-09-24
Added module to core to allow the import of enumerated echo classes 

scollis:
in any module use: from pyart.core.echo_classes import * (or relative import)
This will populate namespace with CLASS_DESCRIPTOR
and CLASS_METADATA
CLASS_METADATA contains verbose descriptors intended for use in things like field['notes'] etc.. 

This module is added with the intent to grow as users/contributors add more classes.. But of course module developers do not need to use ALL classes.. eg a classification algorithm may only use classes 1,2 and 10 (eg clutter, non-meteo and a yet to be named class, let us call it "ships") 

Use of this module will unify enumerated classes.. it is expected future PRs will add to this.. The metadata string can be trimmed to use using searches between \n

### DIFF
--- a/pyart/core/__init__.py
+++ b/pyart/core/__init__.py
@@ -19,6 +19,6 @@ Core Py-ART classes and function for interacting with weather radar data.
 
 from .radar import Radar, is_vpt, to_vpt
 from .grid import Grid
-from .echo_classes import *
+import echo_classes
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
Closing, will open new PR when we have merge this with a use case.

PR #223, opened 2015-01-14
ENH: GateFilter class

jjhelmus:
Add a GateFilter class in the pyart.correct namespace which is intended to be used for creating boolean array for selecting gates for processing.


jjhelmus:
I have a few additional commit that should be included in this PR but they are on top of a branch which add an improved dealiasing routine to the pyart.correct namespace.  Merging so the forth coming PR can be merged cleanly.

PR #227, opened 2015-01-19
BUG: Non-collected gates in Sigmet files masked

jjhelmus:
Non-collected gates in Sigmer files are masked for rays where the number of bins in the ray is smaller than the
number of gates recorded in the product header.


jjhelmus:
Should mention that since masking of non-collected gates was added as a Cython cdef function the overhead on file with no non-collected gates is negligible (< 0.01 sec).

PR #232, opened 2015-01-22
Fail softly when volume number is unknown

akleeman:
While attempting to read in [NPOL DATA](ftp://gpm.nsstc.nasa.gov/gpm_validation/ifloods/NPOL/data) I encountered a bug in which reading the UF file failed because a volume number was not recognized.

```
  File "pyart/pyart/io/rsl.py", line 122, in read_rsl
    rsl_field_name = VOLUMENUM2RSLNAME[volume_num]
KeyError: 46
```

This hints that the VOLUMENUM2RSLNAME dict is out of date.  This change doesn't update the dict, but instead lets reading continue after logging a warning:

```
2015-01-22 10:54:04,253 WARNING root: Unknown Volume Number 46
2015-01-22 10:54:04,253 WARNING root: Unknown Volume Number 46
```

### DIFF
--- a/pyart/io/rsl.py
+++ b/pyart/io/rsl.py
@@ -14,12 +14,20 @@ Python wrapper around the RSL library.
 """
 
 # Nothing from this module is imported into pyart.io if RSL is not installed.
+import sys
 import numpy as np
+import logging
 
 from ..config import FileMetadata, get_fillvalue
 from . import _rsl_interface
 from ..core.radar import Radar
 from .common import dms_to_d, make_time_unit_str
+
+logger = logging.getLogger()
+handler = logging.StreamHandler(sys.stderr)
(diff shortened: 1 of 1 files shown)


jjhelmus:
@akleeman Thanks for finding this and for the fix.  I'd prefer that a warning be issued rather than logging the problem.  This can be done with the [warnings](https://docs.python.org/2/library/warnings.html) modules in the standard library.  The syntax is basically:

``` 
import warnings

# code
# something bad occurs
warnings.warn("Something bad occurred")
```

This makes it easy to turn these into exceptions, complete ignore them, or capture them with a user specified logging interface.

I'll look into updating the **VOLUMENUM2RSLNAME** dictionary.  I think a few volume numbers where added in some of the new released.

akleeman:
Anything else you want added to this?

jjhelmus:
Nope looks good.  Sorry I didn't merge this earlier, I didn't catch the second commit.  Thanks for the contribution!

PR #249, opened 2015-03-03
ENH: Enable option of not plotting features on basemap for radarmapdispl...

scollis:
...ay


jjhelmus:
Looks good, thanks Scott.

Issue #251, opened 2015-03-12
Incorrect range when reading files with variable gate spacing using read_rsl


jjhelmus:
When a file in which some rays do not have an identical gate spacing as the first ray of the first sweep the range attribute of the created **Radar** object does not contain the correct gate locations for all rays but rather the ranges for the first ray.  Currently the CfRadial convention on which the **Radar** class is based has not manner for expressing a radar volume in which the gate locations vary by ray.  Until a preferred method of expressing such volumes is determine it would be reasonable for Py-ART to raise a error or warning when data in this form is encountered.

This issue was reported by Anderson Gama.

jjhelmus:
With PR #256 reading files with variable gate spacing will raise an IOError unless the **skip_range_check** parameter is set to True:

```
$ python -c "import pyart; pyart.io.read('volume.uf')"

gzip: stdout: Broken pipe
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/jhelmus/python/pyart/pyart/io/auto_read.py", line 129, in read
    return read_rsl(filename, **kwargs)
  File "/Users/jhelmus/python/pyart/pyart/io/rsl.py", line 125, in read_rsl
    raise IOError(message)
IOError: Range bin locations change between rays. File cannot be read with with correct range locations for all rays. To read in data reporting the ranges from the first ray set the **skip_range_check** parameter to True.
```

jjhelmus:
Closed by PR #256

PR #256, opened 2015-03-16
ENH: Added check for uniform range bin locations

jjhelmus:
Added optional parameter **skip_range_check** to read_rsl function which
controls if a check for uniform range gate locations for all rays is performed.
When this check is performed and a the gate locations change a IOError is
raised.


jjhelmus:
Closes #251

PR #259, opened 2015-03-26
Wradlib texture

scollis:
This PR is the first in the "Bridge" series. This one starts with wradlib [1] using their implementation of JJ Gourley's texture calculation 

[1] http://wradlib.bitbucket.org/

### DIFF
--- a/doc/source/dev_reference/bridge.rst
+++ b/doc/source/dev_reference/bridge.rst
@@ -1,0 +1,7 @@
+============
+pyart.bridge
+============
+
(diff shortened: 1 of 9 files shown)


jjhelmus:
This passes CI because the `--no skip` option of nosetests has been broken for some time [nose issues #512](https://github.com/nose-devs/nose/issues/512).  I'll add wradlib to the CI and check again.

jjhelmus:
Merging. I have a commit which adds wradlib to the CI, will make a new PR for that work.

PR #260, opened 2015-03-27
ADD: Backing out SNR from reflectivity

scollis:
Super simple calculation of SNR from reflectivity.


scollis:
Doh! Forgot Init file..

PR #263, opened 2015-03-31
experiments with parellel

gamaanderson:
This is a parellel version of map_to_grid. Its almost linear with the number of cores. It is a littel messy, if you want to merge we could explain a littel


gamaanderson:
@jjhelmus I would like to close this pull request, as now I understand parallelization is really application dependent, so this should not be merge any way. I also don't like to see it in the PR list, getting mixed with important ones.

jjhelmus:
@gamaanderson Closing this sounds like a good idea.

PR #264, opened 2015-03-31
Latlon for points in grid

gamaanderson:
add a function to create in grid object from the axes x_disp,y_disp,lat and lon, 2-d axes for the latitude or longitude of every point the the xy plane. The math is "simple" azimuthal equidistant projetion to latlon. I map it into pyart so I can call as a user, but one should decide how to implement, if as method in grid or something else

### DIFF
--- a/pyart/io/__init__.py
+++ b/pyart/io/__init__.py
@@ -22,6 +22,7 @@ from and write data to a number of file formats.
     write_cfradial
     read_grid
     write_grid
+    add_2d_latlon_axis
 
 """
 
@@ -38,5 +39,6 @@ from .nexrad_archive import read_nexrad_archive
 from .nexrad_cdm import read_nexrad_cdm
 from .grid_io import read_grid, write_grid
 from .auto_read import read
+from .common import add_2d_latlon_axis
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/io/common.py
+++ b/pyart/io/common.py
@@ -120,16 +120,21 @@ def make_time_unit_str(dtobj):
     return "seconds since " + dtobj.strftime("%Y-%m-%dT%H:%M:%SZ")
 
 
-def add_2d_latlon_axis(grid):
+def add_2d_latlon_axis(grid,**kwargs):
     """
     Add to Grid (in place) a 2-dimensional axes for latitude and longitude 
-    of every point in the y,x plane
+    of every point in the y,x plane. If available conversion is done using Basemap,
+    extra arguments in **kwargs are passed to that function. If not available
+    internal implementation is used
 
     Parameters
     ----------
     grid: grid object
         Cartesian grid object containing the 1d axes "x_disp","y_disp" and 
         scalar axes 'lat','lon'.
+    **kwargs: Basemap options
+        Options to be passed to Basemap. If projection is not specified here it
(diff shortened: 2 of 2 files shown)


kirknorth:
I think the most generic approach here would be to utilize the `pyproj` package to compute the locations of each (x, y) point in a `Grid`. The idea here is that the user would be able to more accurately compute these locations  anywhere on the Earth through the ability to change the `proj` kwarg for the map projection, and to a lesser extent the `datum` and `ellps` kwargs. Just my two cents.

gamaanderson:
That a good idea @kirknorth , this however will create a new dependency (or is it already in use some else?), but if not one has a problem with that I will be implementing it.  I wounder if Grid should have a internal proj or more extensive projection description, but that is not for me to decide, I'm just using kwarg as you suggested.

kirknorth:
@gamaanderson yes it's implicitly required to use any of the map display routines (e.g. `GridMapDisplay`). I say implicit because `pyproj` is available from `basemap`.

scollis:
Yep.. We can call pyproj from Matplotlib..

However Matplotlib is an optional dependancy.  So I suggest we build on 
Anderson's PR with a check for matplotlib.. if matplotlib exists it 
should call pyproj.. if not it can use the simple but effective 
technique he has proposed.

Great discussion all.


On 4/1/15 6:32 PM, Kirk North wrote:
>
> @gamaanderson <https://github.com/gamaanderson> yes it's implicitly 
> required to use any of the map display routines (e.g. 
> |GridMapDisplay|). I say implicit because |pyproj| is available from 
> |basemap|.
>
> —
> <https://github.com/ARM-DOE/pyart/pull/264#issuecomment-88663465>.
>

-- 

gamaanderson:
I try to implement Basemap, but my version (Debian wheezy) seem to have a bug. Please some one try the following:

from mpl_toolkits.basemap import Basemap
b = Basemap(lat_0=0,lon_0=0,projection='aeqd')
b(2.5e+5, 2.5e+5, inverse=True)

This should return (2.24, 2.24). Mine return (1e+30, 1e+30)!

Anyway I let Basemap version commented and put pyproj version instead
Error from the internal implementation is around 10e-3, probably because of earth radius

tjlang:
I get 1e+30 too. Ananconda on Mac and Linux.

Timothy Lang

kirknorth:
@gamaanderson when I do

```python
from mpl_toolkits.basemap import pyproj

pj = pyproj.Proj(lat_0=0.0, lon_0=0.0, proj='aeqd')
print pj(2.5e5, 2.5e5, inverse=True) -> (2.24694..., 2.26033...)
```

gamaanderson:
In fact it does, i didn't realize that , but documentation indicates Basemap should work just as good.

Anyway, should I just import pyproj from mpl_toolkits.basemap and put Basemap to rest?

kirknorth:
This is what we already do in **gridmapdisplay.py** so I wouldn't worry about being inconsistent or anything of that sort.

gamaanderson:
Done and tested! Thank you for the help @kirknorth

jjhelmus:
In the future this functionality should be added as a method to the **Grid** method, but this can wait for the refactoring of the class as discusses in #285.  At the same time the metadata of the latitude and longitude dictionaries should moved to the `default_config.py` file.  

There are a few style issues that I will fix post-merge.

PR #265, opened 2015-03-31
Fast grid mapper

gamaanderson:
This is a different version of map_to_grid made to be as fast as possible. It works like TITAN, that is, instead of pushing radar points to Cartesian and then interpolate a cloud of points it search for points in the radar coordinates. This has some limitation but it makes 1 millions grid points in 1.3s. It still only with nearest_point interpolation, but one could add multi-linear. I needed to use a lot of Cython to get that velocity and that make me think how much a full Cython version of the original map_to_grid could make.

P.S. to call it use map.grid_from_radars(radars, grid_shape, grid_limits ,grid_mapping_func='fast_map_to_grid' ,fields=[...])


scollis:
Very nice work @gamaanderson ! We will definitely take a good look at this one. We have been talking about doing similar.

gamaanderson:
Just add some comments, if some one is going to revise function cart_to_radar_coords in cython_fast_grid_mapper.pyx this picture could be helpful :
cart_to_radar

jjhelmus:
@gamaanderson Great work, these look to be some very useful function to add to Py-ART.! 

I'll try to provide feedback on all of these pull requests in the few weeks as my schedule permits so that they can be updated and merged into the mainline.  I have a number of projects I'm working on at the moment so feedback might take a while.

Chapter 2 of Doviak and Zrnic "Dopper Radar and Weather Observation" also contains a good discussion relevant to the  `cart_to_radar_coords` function and has a similar figure.  But the one you provided is much more readable.

gamaanderson:
@jjhelmus Thank you and please take your time. Also do not feel overwhelmed, as we are kind of on the run with implementing pyart we may produce a lot of code, but we do not expect these pull requests to be worked on quickly, on the contrary we expect them to hang there for some time. Also if you consider that in some point we are taking pyart to a direction that you don't want to follow, that is also ok, we will still work to maintain our fork as close as possible to the main one and to contribute in as much points as we can.

In respect to  cart_to_radar_coords, we based ourselves in Doviak and Zrnic, but they didn't deduce those formulas there, just the inverse ones, so we needed to do it (just some trigonometry). Just want to say we are using this formulas for vertical cuts for quite some time now and they seem pretty good, showing the same kind of images they show in the book.

gamaanderson:
Just added some improvements, mainly linear interpolation in the elevation coordinate, which is the most critical. Unfortunately this come with a lost in readability, especially in the cython side. I tried a lot, but it is a really critical part of the code, even small alteration result in severe performance lost. That been said, I still consider it to be a important inclusion to pyart, but future work should concentrate in the map_to_grid function.

jjhelmus:
Was it decided that this could be closed without merging?  The `map_gates_to_grid` method is likely faster than this method.  I still would like to get back to gridding at some point but not soon.

jjhelmus:
I'll close this for now as I do not have the time to fix this up.  It would be great to have fast nearest-neighbor and linear interpolation mapping methods in Py-ART but I think the Grid class needs some clean up first.  I'll create an issue for the new interpolation methods and reference this PR as a guide.

PR #266, opened 2015-03-31
Binarize

gamaanderson:
add function that binarize the data of a ncvar dict. This is done by hand, if one could add to this dics the data type on wants in the netcdf it could be done automatically by NetCDF-4


jjhelmus:
Once this passed CI this should be good to go.  At some point a description of the special keys should be added to the docstring of the write_cfradial function.

gamaanderson:
I would prefer a manual entry describing the dictionary itself, beginning by the `data` key. This is because other functions also use this special keys.

jjhelmus:
Yes, making these notes in the Radar/Grid class descriptions does make more sense.

PR #272, opened 2015-04-01
ENH: Faster region based dealiasing

jjhelmus:


jjhelmus:
With this enhancement the time required to dealias an entire CSAPR volume using the default gatefilter and values of 100 for the gaps decreased by 60% from 1 minute and 34 seconds to 37 seconds.

Issue #100, opened 2013-11-13
Algorithm output inconsistency


kirknorth:
I have noticed the following inconsistencies when using some of the Py-ART algorithms. As an example, consider reading a C-SAPR file with `radar = read_mdv()`

1. All fields in the radar object are dictionaries and each field has a `_FillValue` key, and the data arrays are all NumPy masked arrays with `fill_value = -9999.0`

2. Running `vr_corr = dealias_fourdd()` outputs the dictionary `vr_corr` which is <i>missing</i> a `_FillValue` key, but its data is still a NumPy masked array with `fill_value = -9999.0`

3. Running `phidp_corr, kdp_corr = phase_proc_lp()` outputs the two dictionaries `phidp_corr` and `kdp_corr`. The data in these dictionaries are NumPy arrays, <i>not</i> NumPy masked arrays. However each dictionary does have a `_FillValue` key which is set to -9999.0

4. Running `spec_at, ze_corr = calculate_attenuation()` outputs the two dictionaries `spec_at` and `ze_corr`. The `spec_at` dictionary <i>doesn't</i> have a `_FillValue` key, its data is a NumPy array <i>not</i> a masked array, its data has `NaN` in it, and there are no -9999.0 values in it. The `ze_corr` dictionary <i>does</i> have a `_FillValue` key set to -9999.0, its data is a NumPy masked array with `fill_value = -9999.0`, but its data also has `NaN` in it.

jjhelmus:
@kirknorth Are you still seeing inconsistencies in the output of algorithms?  In general, retrievals and corrections in Py-ART should return one or more dictionaries of field data and metadata with the data stored as a NumPy ndarray or masked array.  Adding this data to the radar.fields attribute dictionary is also acceptable but not recommended.

gamaanderson:
I just notice this inconsistency:
grid_shape order in 
pyart.testing.make_empty_grid is (x,y,z)
but in
pyart.map.grid_from_radars is (z,y,x)

jjhelmus:
@gamaanderson Looks to be a remnant from PR #115 where the order of the grid_shape was changed.  I'll put together a PR to address it this morning.

jjhelmus:
Anyone following this issue want this kept open?  If not I'll close it later this week.

gamaanderson:
From my part, this is ok to close.

jjhelmus:
Never got around to closing this but am now.

PR #274, opened 2015-04-03
MAINT: Fixed inconsistency in grid shape ordering

jjhelmus:
All grid related functions and class should use a z, y, x ordering
for shape and size.  Radar related functions may still use x, y, z.


jjhelmus:
Address inconsistency highlighted by @gamaanderson in issue #100

PR #271, opened 2015-03-31
Read KAZR ingest

kirknorth:
Adding the ability to read KAZR ingest (e.g., a1 data). These file types are characterized as NetCDF files that are not fully CF/Radial in terms of their convention. Nonetheless it borrows heavily from the existing CF/Radial module. I haven't looked at other data streams (e.g., 915-MHz RWP), but if they have an ingest similar to the KAZR, then read_kazr can likely be renamed to something more generic...

In terms of testing, I have only confirmed that I am indeed able to create a radar object from KAZR data. I have not yet tried to use it in any of the other modules (e.g., `correct` or `graph`).

PR #275, opened 2015-04-03
BUG: Support for dealiasing vpt scans with dealias_unwrap_phase.

jjhelmus:
Specifying the unwrap_unit as 'sweep' or 'volume' supported when
unfolding VPT scans with the dealias_unwrap_phase provided the data
is stored as a single sweep.


jjhelmus:
Address issue with use of **unwrap_unit='sweep'** in `dealias_unwrap_phase` when unfolding VPT volumes as pointed out by @kirknorth in #271

PR #277, opened 2015-04-06
Write mdv

gamaanderson:
Major alterations in class MdfFile so that it can write mdv files. Another PR will follow this one to convert a grid object to mdv.

This are some big change, but they should not change anything in the user side, so the best way would be to make tests to guaranty existing functionalities were not affected. If some one were kind of instructing me in that matter I would be happy to perform those.

There are also some small style change, e.g. change 'nrays' to 'ny' as this is the original name in mdv documentation. I would also be happy to discuss them.


scollis:
This will be a useful feature for integration with TITAN.

gamaanderson:
@scollis This is exactly why I'm doing it! We are intending to skip TITAN interpolation and do forecasting using py-art mdv's. But I'll not know if it work till I complete the next PR.

gamaanderson:
Travis CI is falling doe to the style change, I didn't realize plot used MdvFile directly. I could also perform change there, but I feel I would need some approval. 
My point is that MdvFile should be as close as possible to the file structure, so all problems are restricted to the putting and retrieving that from the structure, no in the read/write operation. Of course this is will create some inconvenient style, but this is a result of mdv being originaly ideialised in

jjhelmus:
@gamaanderson Agreed, the **MdvFile** class should reflect the structure of MDV files as closely as possible.  **MdvFile** has not be updated recently and could use a fresh set of eyes, thanks for taking this on!  Have you seen how the unpacking of structures is accomplished in the **SigmetFile** and **NEXRADLevel2File** classes with the tuple-of-tuples and ` _unpack_structure` function?  Following this layout might simplify dealing with the MDV specific structures.

There is a [**MdvDisplay**](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.graph.MdvDisplay.html#pyart.graph.MdvDisplay) class which can be used to plot MdvFile objects in the same manner as the RadarDisplay class is used to plot Radar objects.  I've in favor of depreciating this object along with the other file specific Display classes, **RslDisplay** and **CFRadialDisplay**, rather than supporting them in the long run.  Offering a single procedure for plotting data is simpler and will be easier to maintain in the long run and I think the loss of functionality would be acceptable.  In addition adding a **Display** class for each supported file format will become cumbersome, the **Radar** abstraction should be used instead.  Anyone who wanted to plot data without generating a **Radar** object would still be able to but they would need to write their own code to do the plotting.  Implementing this depreciation would also be a good time to refactor some of the modules in the graph directory.

gamaanderson:
@jjhelmus I haven't seen this files. In fact, this is better solution than mine, I shall implement it. 

I agree these display classes should be deprecated, it is just to much work to little benefit. But answer something for me, when you say "would still be able to", you are implying the user would have access to the MdvFile class, if so, one may need to improve the code, not that it don't work, but it is vulnerable to misuse.

One final thing, even if MdvDisplay is to be deprecated, my alteration are not deprecating it, they are broking it. I still believe I should correct it and add a deprecated tag, so future alterations may ignore it. .

jjhelmus:
The convention is Python is that private or internal portions of the code are indicated by prefixing the class, function, method or module with an [underscore](https://www.python.org/dev/peps/pep-0008/#public-and-internal-interfaces).  Since the **MdvFile** class does not have this prefix it should be considered in the public API which users have access to.  That said, the class is not imported into the pyart.io namespace nor included in the [User Reference Manual](http://arm-doe.github.io/pyart-docs-travis/user_reference/io.html) and so it not as public as say the `read_mdv` function.  

I've always treated the file specific classes, **MdvFile**, **SigmetFile**, **NEXRADLevel2File**, as accessible to "power users" who want to more complete control over the specific file and do not need the other functionality and niceties that Py-ART provides with the **Radar**class.

All that is a long way of saying; I think it is fine to change the implementation details of the **MdvFile** class as is done here as long as `read_mdv` is update to use the new implementation.  Power users can still use the class for their own use, such as plotting MDV files without creating a **Radar** object, but should not be surprised or upset at this or future changes as the class as it is not documented in the User Manual.  

As for depreciating **MdvDisplay** you are correct, the proper method is to add a depreciation warning, wait a release and then remove the class.  Although I doubt the class is widely used, since it is documented in the User Manual and used in the Example this is the path we will follow.  Once you have finished up the modification to the **MdvFile** I can make the necessary changes to **MdvDisplay** and add the depreciation bits.

gamaanderson:
Ok, thank you @jjhelmus , I will make the needed updates "soon".

scollis:
Great work all. Anderson: These additions are appreciated and I look 
forward to seeing how this integrates in TITAN.

On 4/9/15 11:39 AM, Anderson wrote:
>
> Ok, thank you @jjhelmus <https://github.com/jjhelmus> , I will make 
> the needed updates "soon".
>
> —
> <https://github.com/ARM-DOE/pyart/pull/277#issuecomment-91285944>.
>

-- 

gamaanderson:
I improved how packing and unpacking handle strings. However struct does not work with arrays, and those are used a lot (more than sigmet), mainly in vlevels_header there is 122 floats for the elevation of every level, so I can not implement the structure as in sigmet, I still need to keep track of array size.
I've opened an issue in python tracker (http://bugs.python.org/issue23933), but of course this will not change any time soon.

I think I could still make alteration, but they are not that much of an improvement and would still be inconsistent with sigmet. That is, I would vote to let as it is.

About the change in style that shall affect `MdvDisplay`, they are:
1. In master_header change "nsweeps","nrays","ngates" to "max_nz","max_ny","max_nx"
2. In field_header change "nsweeps","nrays","ngates" to "nz","ny","nx"
3. Fields are no longer saved using "field_name" as attributes of `MdvFile`, but by index in `field_data` (list), however as calling `read_a_field` already used fnum, this does not change the calling of the function
4. No longer calls `_calc_geometry()`, `_make_carts_dict()` in `_init_`, as these are specific for polar mdv, so to get the attributes az_deg, range_km, el_deg and carts, one must call this functions directly.

For last, just a question. The definitions in capital letter made just before class MdvFile (e.g. PROJ_POLAR_RADAR = 9), I'm using those in writing grid as mdv and this is in a file mdv_io.py, so I need to import the whole mdv.py file, and not just the class MdvFile. Is this the desirable behavior?

gamaanderson:
I've corrected some errors and updated the test routines to fit the new style. Will now start working in issue #286

gamaanderson:
Done. As for me, this is ready to merge. Then I can add the read and write grid mdv routines that @scollis is waiting for.

gamaanderson:
result of the speed test for lazy read MDV
------------------normal----------------------------lazy
open file-------0.00467586517334------------0.00313305854797
read field------0.000511169433594----------0.00318813323975
total------------0.00518703460693------------0.00632119178772

jjhelmus:
@gamaanderson I'll merge this tomorrow once I can have some time to add adjust the MDVDisplay class as needed.  Thanks for all the work in getting this updated!

jjhelmus:
@gamaanderson Looks like this is good to go.  There are a few PEP8 issues in mdv.py, is it alright with you if I fix these post-merge?  I'd also like to add a few additional unit tests to get the coverage up.

gamaanderson:
ok, make all fixes you want. I know about some of the PEP8 issues, but I let them, as I was not sure how to solve.

jjhelmus:
Sounds good, merging!  Thanks for all your work on this should open some exciting new uses.

PR #280, opened 2015-04-11
ENH: inclusive kwarg to filters

kirknorth:
Adding an `inclusive` keyword to the relevant filter methods (e.g., exclude_below). This gives the user more control over the specifics and reduces ambiguity, for example, it's ambiguous as to whether the actual minimum and maximum values specified in exclude_inside (include_inside) will be excluded (included) or not unless the user dives into the actual code. One final point is that the default setting of the inclusive kwarg is consistent with what @jjhelmus originally intended with the filter methods.

### DIFF
--- a/pyart/correct/filters.py
+++ b/pyart/correct/filters.py
@@ -227,8 +227,8 @@ class GateFilter(object):
             Name of field compared against the value.
         value : float
             Gates with a value below this value in the specified field will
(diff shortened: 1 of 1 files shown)


jjhelmus:
@kirknorth Looks goods!  Thanks for the added syntax and for fixing the documentation.

Issue #138, opened 2014-03-27
No method for specifying the altitude of a grid in the map_to_grid function


jjhelmus:
The map_to_grid function does not provide a method for specifying the starting altitude of a grid.  The default is to set it to the altitude of the first radar which can then be adjusted by setting the limits of the z grid_limits.  This is far from ideal.

gamaanderson:
I would like to revive this topic, since it is not really clear to me and there may to be some inconsistence in file io.grid_mapper.py. 
Function `map_to_grid` consider in lines 399 grid origin altitude to be always zero, but the reference is not clear, since it does not correct radar.altitude when calculating Cartesian coords (line 455). There is also a comment in line 445
On the other hand, function `grid_from_radars` in line 136 consider grid origin to be the first radar altitude.

I believe we should clarify this once for all and add this information in the function description.  As for me I would prefer altitude to be (always) absolute (meters above sea level).

jjhelmus:
I agree that the user should be able to specify an altitude above sea level for the starting level of the grid and that the algorithm should correct for the difference between this altitude and the altitudes of the Radar objects which are being gridded.  I still think that the default value should be the altitude of the first radar.

A bit more explanation of what the code current does.  In the Grid classes axes attribute there are two dictioaries, **alt** and **z_disp** which record the altitude of the origin of the grid and the displacement from this origin.  To get the altitude above mean sea level for each point in the grid one must add the value found in **alt** dictionary to the values found in **z_disp** dictionary.  This is why line 136 in the `grid_from_radars` function sets **alt** to the altitude of the first radar, this is considered the origin of the grid.

In the gridding algorithm the altitudes of all the radars are assumed to be the same, this assumption is not correct and any displacements in altitude between the radars should be taken into account which is why there is a XXX comment on line 445.  If/when the option to specify the altitude of the radar is added then offsets for all radars altitudes to this value should be calculated and the z gate locations should be corrected in line 455 in the same manner as is done for the y and x gate locations.

gamaanderson:
Being that the case, I believe is just a question of implementing. Unfortunately, because of the z,y,x order, transforming the grid_origin in 3-truplet has the possibility of creating bugs. Nevertheless, I think this should be the way of doing it.
How are you dealing with somethings like this, updates that may create bugs in user-scripts? How connected are you to your users?

scollis:
FYI The z,x,y order is there to mirror CF standards


On 4/23/15 11:54 AM, Anderson wrote:
[...]

PR #284, opened 2015-04-23
solve issue #138

gamaanderson:


jjhelmus:
@gamaanderson Thanks for the quick turn around on this.  I'm glad to have this issues addressed.  

Two very minor fixes before this is merged.  There is a violation of the [pep8](https://www.python.org/dev/peps/pep-0008/) style guidelines on line 262 which we are following in Py-ART, line 588 also should have an extra space if you want to fix that at the same time.  Line 456 contains tailing whitespace which is a no-no, I have my editor set to strip trailing whitespace on .py files to prevent this.

I can fix these two issues after merging if you would prefer, just let me know.

pyart/map/grid_mapper.py
```diff
@@ -256,6 +259,7 @@ def find_neighbors_and_dists(self, q, r):
 
 
 def map_to_grid(radars, grid_shape, grid_limits, grid_origin=None,
+                grid_origin_alt = None,
```
jjhelmus: There should not be spaces around the = sign, grid_origin_alt=None

pyart/map/grid_mapper.py
```diff
@@ -442,7 +452,8 @@ def map_to_grid(radars, grid_shape, grid_limits, grid_origin=None,
         radar_lat = float(radar.latitude['data'])
         radar_lon = float(radar.longitude['data'])
         x_disp, y_disp = corner_to_point(grid_origin, (radar_lat, radar_lon))
-        offsets.append((0, y_disp, x_disp))  # XXX should include z
+        z_disp = float(radar.altitude['data']) - grid_origin_alt
+        offsets.append((z_disp, y_disp, x_disp))
```
jjhelmus: Tailing whitespace, should be removed.

gamaanderson:
I will solve this. I'm not used with playing that much attention to style and I also follow a some what different style. So I'm sorry, I will try to avoid in the future, but this will probably happen again.

jjhelmus:
Do not worry about style too much, I fix up a lot of style issues in pull requests and an happy to do it.  There is a command line tool that I used to check for style prior to checking in files that you might find useful, [pep8](https://pypi.python.org/pypi/pep8).

gamaanderson:
what is the problem in line 588? I can't see it

jjhelmus:
There should be a space between # and the first letter.  It has been in there for quite some time, pep8 only started picked up the error a few months back.

gamaanderson:
in line 262 I've put grid_origin_alt in a line of its own, breaking the overall block, should I also correct that?

jjhelmus:
Would be good, it is not a technical violation it just looks out of place.

jjhelmus:
Merged, FYI there was a long line that I fixed up pre-merge ca611ddd9e

Issue #278, opened 2015-04-07
Encoding Value as int


gamaanderson:
I would like to open a discussion in this topic, since I already found it twice (netcdf and mdv). It's about when writing objects, in order to reduce file size, encoding variables as int (8 or 16 bits), instead of using float. The problem is how to decide the type, I see three possibilities:
1. the user encodes the data before calling write and them we use dtype.
2. the user's adds an attribute to the variable (like "write_dtype"="int8") and optionally also add scale and offset.
3. The user gives what type he want while calling write

An utility for number 1. is already in PR #266, but this is far from optimal since it may destroy data

Number 2. is my favorite and would be just a question of defining the attribute name and implement

I am not a fan of number 3. since it would need to add an option in all write routines; however one thing plays in its favor: NetCDF just use signed ints and MDV just UNsigned ones. However I believe we could recalculate the scale and offset to solve this problem without user interference.

What do you think? do you see any problem or other alternatives?

josephhardinee:
When you say writing objects do you mean serializing of the python class, or the writing of file formats? For the filetypes, many specify an encoding ahead of time we would have to comply with. For serialization of the python objects, I guess the question would be how often do people write out the actual objects to disk?

gamaanderson:
I mean writing file formats. In fact filetypes specify an encoding, but some of them also give some choice. CfRadial for instance allow for fields: ncbyte, short, int, float and double (all signed). So the question is how to allow the user to choose, comply with filetypes and at the same time don't fill up function call with options.

jjhelmus:
Agreed, for file formats which specify an encoding (MDV), the only option is to cast the values in the arrays into the appropriate type after attempting to find reasonable scale/offset parameters if these are not provided by the user.  

For file types which do allow for a choice of encoding the question of what to do becomes more complicated.  I believe that with the [compression of variables available with NetCDF4](https://www.unidata.ucar.edu/software/netcdf/docs/netcdf/NetCDF_002d4-Format.html) there is little benefit to encoding values into integer types to save space while maintaining the original data precision, that is not "destroying data".  In addition when a loss of precision is acceptable there are more efficient manners of reducing the size of data than encoding data into other data types.

[...]

PR #292, opened 2015-05-15
Write grid mdv

gamaanderson:
One PR closes, an other opens. But I do think all issue that may emerge are easy fix, I simplify add functions `write_grid_mdv` and `read_grid_mdv`.

There is this thing with encoding, I forget to write in the documentation but float64 is being casted to float32, there is a hack to allow other encoding available in MDV file, but a much better solution can be done following PR #266 and issue #278, when that is merged.

Pay attention that I'm importing *mdv.py* as  `import mdv as MDV`, I know this is bad, but I needed some constants defined there. Please change that to what you consider better.

You may also consider moving `read_mdv` from *mdv.py* to *mdv_io.py*.
 
Finally, the git history is a mess, you may rebase if you want, I'm not familiar in doing that.


jjhelmus:
This will make another great addition and will make adding unit tests for the writing function of **MdvFile** much easier.  I need to see to a few other tasks before I can review this but hopefully should be later today or Monday at the latest.

jjhelmus:
Also ignore the failing coverage test, I have no idea why that started popping up and need to figure out how to disable it.

gamaanderson:
sounds good.

jjhelmus:
I reverted (undid the commits) in PR.  When writing unit tests for the new code there were a number of bugs that need to be address before this is merged into the master.  Sorry for not checking these before merging.  @gamaanderson If you can create a new PR I'll outline the issues and submit a PR with some failing tests to highlight some of the issues.

gamaanderson:
done, PR #297. I had to do it by hand, hope is everything right

PR #295, opened 2015-05-18
MAINT: Clean up of mdv modules.

jjhelmus:


jjhelmus:
Check list of task in the maintenance from PR #292 

[ ] pep8 on mdv_io.py
[ ] pylint check on mdv_io.py
[ ] move MdvFile object to mdvfile.py module.
[ ] split functionality of mdv.py and mdv_io.py into mdv_grid.py and mdv_radar.py
[ ] Adjust docstrings as needed.
[ ] Create unit tests.
[ ] Add depreciation warnings to MdvDisplay, RslDisplay, and CfRadialDisplay classes.
[ ] Remove examples of above Display classes from documentation.

jjhelmus:
@gamaanderson Two question concerning the new functionality of the MdvFIle class.

1. In the `write_grid_mdv` function the documentation indicates that the field data should be compressed prior to the call but this seems inconsistent and broken at the moment.  Having the field data be passed in the packed state before writing is inconsistent with how the data is read.  When reading field data the **fields** attribute contains uncompressed data (where the scale and bias has been applied).  In addition the `_write_a_field` method applies the scale and bias and performs rounding of all types of data including floats.  I think a better option is to treat the field data in the Grid object as uncompressed and apply scaling, offset and compression using field dictionary keys ("scale_factor", "add_offset" and "_Write_as_dtype") as in PR #266.

2. The `if self.fileptr is None` blocks in the following **MdvFile** methods are never called, can they be removed or replace with raise NotImplemented?

    * `_get_chunk_header`
    * `_get_radar_info`
    * `_get_calib`
    * `_get_compression_info`
    * `_get_level_info`

Also the `_time_dict_into_header` method is never called, can this be deleted?  If any of this functionality is needed in the future it can be retrieved from version control, but leaving unused and untested code in the mainline is typically frowned upon.  I added stub tests for these in *test_mdv_io.py* but the do no verification of the results.

Finally, I'm thinking of splitting the modules so that the MdvFile class and file specifications are in **mdv_common.py**, the radar reading functionality in **mdv_radar.py** and the grid functions in **mdv_grid.py**.  Does this seem reasonable?

gamaanderson:
First of all, sorry for the silly mistakes, I probably just tested some cases, not all of them. Thankfully you are vigilant for all of this. 

1.  I started working in this code before issue #278 and I end up never implementing it. Now I notice this resulted in some bugs, in fact `write_grid_mdv` can just properly do float encoding, must implement the use of  "_Write_as_dtype" for the other ones. I also just revised MDV documentation, and line [844](https://github.com/jjhelmus/pyart/blob/pr292_additions/pyart/io/mdv.py#L844) is wrong (obviously!), scaling is just applicable for int encoding.

2. I'm not writing any chunks, if I were to I would need those functions, just like I need `_get_field_headers` in `write_grid_mdv`. The name starting with '_' is probably wrong.

3. In `write_grid_mdv` I'm handling the time by [myself](https://github.com/jjhelmus/pyart/blob/pr292_additions/pyart/io/mdv_io.py#L75-L84) but I could eventually use `_time_dict_into_header`. This is actually the inverse of `_make_time_dict`, so, in some way, it should exist, doesn't it?

I may help with *test_mdv_io.py*, add verifications. You would also need some file for testing the grid, wouldn’t you?

Splitting the modules sounds totally reasonable, may be one day `write_radar_mdv` is added

jjhelmus:
Do not worry not much about the mistakes, getting something working is the first goal which you did great.  With the tests added I think most of the bugs have been found and address, I'll refactor a bit more and then I think this can be merged back in.  

As to the specific points

1. I'll add logic to the field writing method to check for a "_Write_as_dtype" key and handle the conversion as needed.

2. The method names should start with a '_' since these are private.  I'm alright keeping them in the code in anticipation of a `write_radar_mdv` function, I'll add some smoke tests to the unit tests related to these function.

3. I agree that having a inverse of `_make_time_dict` has a place in the code, but if it exists it should be used.  I'll see if the `write_grid_mdv` method can be refactored to use the method and not the `time_dict_to_unixtime` function.

Currently I'm creating MDV grid files by writing out a sample grid, having one or two small sample files would be helpful for testing if you have any that can be included.

jjhelmus:
Also for anyone following this PR, I'll want to rebase this branch to include a revert of the revert (962c85aa) and some of the other recent changes in master so that this can tested and merged more easily.  Speak now if you have branches which reference this branch as these will break.

jjhelmus:
Rebased on top of a revert of 962c85aa5a, now it should be possible for this branch to be automatically merged and pass the CI tests.

jjhelmus:
Depreciation of the MdvDisplay and similar classes do not belong in this PR, this task has been removed from the task list.

jjhelmus:
This is good to be merged back in.

gamaanderson:
Thank you @jjhelmus for all the work. I had a pretty busy week, but the good news is that we will be officially running pyart in some of our data.

jjhelmus:
Great to hear, I'm glad we were able to get these features into Py-ART.

PR #307, opened 2015-05-22
MAINT: Deprecate file specific displays

jjhelmus:
Deprecate the MdvDisplay, RslDisplay and CFRadialDisplay classes.

### DIFF
--- a/examples/plotting/plot_ppi_mdv.py
+++ b/examples/plotting/plot_ppi_mdv.py
@@ -3,8 +3,7 @@
 Create a PPI plot from a MDV file
 =================================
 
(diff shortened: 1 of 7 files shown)


jjhelmus:
@nguy Are you using any of these in ARTView?  The preferred method is the use the RadarDisplay or RadarMapDisplay classes but I wanted to check that deprecating these will not effect ARTView.

nguy:
@jjhelmus Nope, only RadarDisplay and RadarDisplay_Airborne are used in ARTView. 
Thanks!

jjhelmus:
Thanks for the conformation @nguy, merging.

Issue #253, opened 2015-03-12
Document optional parameters in dealias_fourdd function


jjhelmus:
A number of optional parameter can be provided to the `dealias_fourdd` function which override default parameter of the `_fourdd_interface.fourdd_dealias` function, these should be documented in the function docstring.

PR #308, opened 2015-06-01
ENH: GateFilter available for 4DD

kirknorth:
Adding the ability to input a radar `GateFilter` when dealiasing Doppler velocities with the James et al. (2001) 4-D algorithm. Note that certain kwargs are now deprecated (e.g., extra_prep) but I'll let @jjhelmus decide if these kwargs should be permanently removed or kept for backwards compatibility.

### DIFF
--- a/pyart/correct/dealias.py
+++ b/pyart/correct/dealias.py
@@ -20,16 +20,18 @@ import numpy as np
 from ..config import get_field_name, get_fillvalue, get_metadata
 from ..io import _rsl_interface
 from . import _fourdd_interface
+from ._common_dealias import _parse_gatefilter
 from ..util import datetime_utils
 
 
 def dealias_fourdd(radar, last_radar=None, sounding_heights=None,
                    sounding_wind_speeds=None, sounding_wind_direction=None,
-                   prep=1, filt=1, rsl_badval=131072.0, keep_original=False,
-                   extra_prep=False, ncp_min=0.3, rhv_min=0.7,
-                   refl_field=None, vel_field=None, corr_vel_field=None,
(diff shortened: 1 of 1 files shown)


jjhelmus:
This is great @kirknorth!  I think deprecating a number of the parameters in this function makes sense and will make the gatefilter the only method for filtering gates prior to applying the 4-D algorithm.  Marking the following parameters as depreciated and removing them after the next release of Py-ART is reasonable:

* extra_prep
* ncp_min
* rhi_min
* ncp_field
* rhv_field

In additional the following parameters can also be Depreciated if the _fourdd_interface file is adjust as needed.  Reflectivity filtering would then be done when creating the Gatefilter:

* refl_field
* prep
* lowdbz (not explicitly mentioned in the docs)
* highdbz (not explicitly mentioned in the docs)
* rm_missing (not explicitly mentioned in the docs)

Anyone have concerns about these changes?

jjhelmus:
Issue #253 can also be addressed by this PR.

jjhelmus:
@kirknorth Mind if I create a branch off this PR to remove the depreciated parameters as well as the reflectivity filtering in FourDD?   I think having the gatefilter be the only method for specifying what gates are excluded is best.  It will take some mucking with the Cython and C files so I think a separate branch and PR will be easier than trying to merge into this one.

kirknorth:
Sure thing @jjhelmus! And yes I agree, I think removing all hard thresholds (e.g., `min_ncp`, `max_refl`, etc.) from Py-ART functions and instead rely on the more codified `GateFilter` class is the proper way to move forward.

PR #315, opened 2015-06-08
BUG: Issue with sigmet module when detecting discontinity

jjhelmus:


jjhelmus:
This commit should have been included with PR #314

PR #324, opened 2015-06-23
TST: Use TRMM RSL version 1.47 build 0 on binstar on Travis CI

jjhelmus:


jjhelmus:
The conda package [TRMM RSL version 1.47 build 1](https://binstar.org/jjhelmus/trmm_rsl/files) seems to be broken, this should be checked in the future.

PR #332, opened 2015-07-13
Handle SQI in Sigmet reader

deeplycloudy:
1-byte SQI data (type 18) was not being read. Add that capability. Based on old handling for 1-byte RHOHV, but adds handling for similarly-scaled 1-byte formats like SQI, RHOH and RHOV.  

Also fixes a string comparison bug that would always evaluate to True.

In order to use this, I had to manually delete and regenerate _sigmetfile.c with cython _sigmetfile.pyx. Removing the build directory and running python setup.py build wasn't picking up my changes.  I wanted to call that to your attention in case you wanted to automate that step when rebuilding.


jjhelmus:
@deeplycloudy this looks great.  I'm currently traveling but will try to have a more in depth look at this by the end of the week.

jjhelmus:
@deeplycloudy Thanks for the bug fix and adding support for reading 1-byte SQI data.Two fixes before this is merged.  

First, you need to check in and commit the regenerated _sigmetfile.c file.  Some projects use the cythonize command to create the .c files during the build step but in Py-ART we choose to keep the .c files in the repository to simplify the build.

Second, your editor seem to be adding or at least keeping trailing whitespace to the _sigmetfile.pyx file, these should be removed.  This is a purely cosmetic change but all other source files in Py-ART have trailing whitespace removed and keeping this consistent would be best.  I have my editor setup to automatically strip trailing whitespace in .py and .pyx files. you should be able to find instructions to add this setup to your editor if you want this behavior.

If you can make these changes I'd be happy to merge this PR.  If you would prefer that I make these changes let me know and I will make them and merge the other commits.  Thanks again for the contributions!

deeplycloudy:
@jjhelmus Fixed! Sorry about the whitespace. Should be set to go assuming Travis checks out.

jjhelmus:
Thanks for the update and the bug fix @deeplycloudy, merging.

Issue #318, opened 2015-06-09
_gate_to_grid_map module does not compile on Windows


jjhelmus:
The _gate_to_grid_map module does not compile on Windows as the Microsoft Visual Studio C++ compiler used to compile Python 2.7 modules does not implement the C99 isfinite function.  The current work around is to disable the check for finite field values.  

In addition Windows does not separate the math library from the standard C library so the "m" library is not need nor can be linked to the module.

jjhelmus:
The second portion of this bug is fixed by PR #319

gamaanderson:
Just for information, I have found the same problem (the first one) while compiling in Linux, but I cann't say now what configuration in was. The problem, I believe, eventually solved itself.

jjhelmus:
Linking against a libc/libm library on Linux which does not implement the C99 isfinite function would cause such an error but you would need to use a very old version of the GNU C library or a non-standard C library for that.  I think long term all filtering of gates for gridding should be handles by the gatefilter object.  Then isfinite would not be needed in the _gate_to_grid_map module.

jjhelmus:
I was able to build Py-ART successfully on Windows with PR #360 merged. Closing this issue as it has been resolved.

PR #334, opened 2015-07-20
BUG: refl_filter_flag=False not working in map_gate_to_grid

gamaanderson:
Just a small bug. Please rebuild cython, I have problems doing in my computer because of isfinite


jjhelmus:
Good catch @gamaanderson I regenerated the .c file and will merge momentarily.  The isfinite issue may be related to #318, I need to find a good cross platform alternative to isfinite.

PR #338, opened 2015-07-23
MAINT: _rsl_interface.pyx can be compiled with modern Cython version

jjhelmus:


jjhelmus:
The fix for this issue in PR #342 is superior to this implementation.  Closing without merging.

Issue #248, opened 2015-03-03
Remove _parse_fig_ax() and similar methods from RadarDisplay, make them independent & public


tjlang:
I find myself often reproducing outside of Py-ART the functionality of _parse_fig_ax() and similar useful methods in the RadarDisplay family of classes. Since these methods do not rely on any class attributes, it is not necessary for them to be part of those classes. If they were pulled out of these specialized classes, and made independent and public, then one could just import them into other codebases (whether Py-ART-related or not) and not have to actually code them up separately (e.g., from pyart.graph import parse_fig_ax).

jjhelmus:
@tjlang Great suggesting.  I'll label this and look into implementing it in the near future.

tjlang:
Not sure I care enough about that. It's enough that I can import them if I have Py-ART installed ...

jjhelmus:
Keeping the function in the `pyart.graph.common` namespace sounds like a plan for now, we can always revisit this in the future if needed.  Closing this issue as parse_ax_fig is now available.

PR #339, opened 2015-07-27
externalise methods from display

gamaanderson:
In the spirit of issue #248; to help a little in artview and in anticipation of improving `GridDisplay`, I made some independent methods based in what is being done in  `RadarDisplay`. I would just like to get an approval before applying to `RadarDisplay` (removing some private methods and redirecting public ones)

Also, do you have any suggestion on how to proceed (define names and stuff) while creating similar ones for grid?

### DIFF
--- a/pyart/graph/common.py
+++ b/pyart/graph/common.py
@@ -152,3 +152,178 @@ def _interpolate_azimuth_edges(azimuths):
     edges[-1] = np.angle(az[-1] - (az[-2] - az[-1]) / 2., deg=True)
     edges[edges < 0] += 360     # range from [-180, 180] to [0, 360]
     return edges
+
+########################
+# Common radar methods #
+########################
+
+def parse_ax(ax):
+    """ Parse and return ax parameter. """
+    if ax is None:
+        ax = plt.gca()
+    return ax
+
+def parse_ax_fig(ax, fig):
+    """ Parse and return ax and fig parameters. """
+    if ax is None:
+        ax = plt.gca()
+    if fig is None:
+        fig = plt.gcf()
+    return ax, fig
+
+def generate_colorbar_label(standard_name, units):
+    """ Generate and return a label for a colorbar. """
+    return standard_name.replace('_', ' ') + ' (' + units + ')'
+
+def generate_field_name(radar, field):
+    """ Return a nice field name for a particular field. """
+    if 'standard_name' in radar.fields[field]:
+        field_name = radar.fields[field]['standard_name']
+    elif 'long_name' in radar.fields[field]:
+        field_name = radar.fields[field]['long_name']
+    else:
+        field_name = str(field)
+    field_name = field_name.replace('_', ' ')
+    field_name = field_name[0].upper() + field_name[1:]
+    return field_name
+
+def generate_radar_name(radar):
+    """ Return radar name. """
+    if 'instrument_name' in radar.metadata:
+        return radar.metadata['instrument_name']
+    else:
+        return ''
+
+def generate_time_begin(radar):
+    """ Return time begin in datetime instance. """
+    # datetime object describing first sweep time
+    times = radar.time['data'][0]
+    units = radar.time['units']
+    calendar = radar.time['calendar']
+    return netCDF4.num2date(times, units, calendar)
+
+
+def generate_filename(radar, field, sweep, ext='png'):
+    """
+    Generate a filename for a plot.
+
+    Generated filename has form:
+        radar_name_field_sweep_time.ext
+
+    Parameters
+    ----------
+    radar : Radar
+        Radar structure.
+    field : str
+        Field plotted.
+    sweep : int
+        Sweep plotted.
+    ext : str
+        Filename extension.
+
+    Returns
+    -------
+    filename : str
+        Filename suitable for saving a plot.
+
+    """
+    name_s = generate_radar_name(radar).replace(' ', '_')
+    field_s = field.replace(' ', '_')
+    time_s = generate_time_begin(radar).strftime('%Y%m%d%H%M%S')
+    sweep_s = str(sweep).zfill(2)
+    return '%s_%s_%s_%s.%s' % (name_s, field_s, sweep_s, time_s, ext)
+
+def generate_title(radar, field, sweep):
+    """
+    Generate a title for a plot.
+
+    Parameters
+    ----------
+    radar : Radar
+        Radar structure.
+    field : str
+        Field plotted.
+    sweep : int
+        Sweep plotted.
+
+    Returns
+    -------
+    title : str
+        Plot title.
+
+    """
+    time_str = generate_time_begin(radar).isoformat() + 'Z'
+    fixed_angle = radar.fixed_angle['data'][sweep]
+    l1 = "%s %.1f Deg. %s " % (generate_radar_name(radar), fixed_angle, time_str)
+    field_name = generate_field_name(radar, field)
+    return l1 + '\n' + field_name
+
+def generate_vpt_title(radar, field):
+    """
+    Generate a title for a VPT plot.
+
+    Parameters
+    ----------
+    radar : Radar
+        Radar structure.
+    field : str
+        Field plotted.
+
+    Returns
+    -------
+    title : str
+        Plot title.
+
+    """
+    time_str = generate_time_begin(radar).isoformat() + 'Z'
+    l1 = "%s %s " % (generate_radar_name(radar), time_str)
+    field_name = generate_field_name(radar, field)
+    return l1 + '\n' + field_name
+
+def generate_ray_title(radar, field, ray):
+    """
+    Generate a title for a ray plot.
+
+    Parameters
+    ----------
+    radar : Radar
+        Radar structure.
+    field : str
+        Field plotted.
+    ray : int
+        Ray plotted.
+
+    Returns
+    -------
+    title : str
+        Plot title.
+
+    """
+    time_str = generate_time_begin(radar).isoformat() + 'Z'
+    l1 = "%s %s" % (generate_radar_name(radar), time_str)
+    azim = radar.azimuth['data'][ray]
+    elev = radar.elevation['data'][ray]
+    l2 = "Ray: %i  Elevation: %.1f Azimuth: %.1f" % (ray, azim, elev)
+    field_name = generate_field_name(radar, field)
+    return l1 + '\n' + l2 + '\n' + field_name
+
+def set_limits(self, xlim=None, ylim=None, ax=None):
+    """
+    Set the display limits.
+
+    Parameters
+    ----------
+    xlim : tuple, optional
+        2-Tuple containing y-axis limits in km. None uses default limits.
+    ylim : tuple, optional
+        2-Tuple containing x-axis limits in km. None uses default limits.
+    ax : Axis
+        Axis to adjust.  None will adjust the current axis.
+
+    """
+    if ax is None:
+        ax = plt.gca()
+    if ylim is not None:
+        ax.set_ylim(ylim)
+    if xlim is not None:
+        ax.set_xlim(xlim)
--- a/pyart/graph/common.py
+++ b/pyart/graph/common.py
@@ -32,11 +32,13 @@ Common graphing routines.
 """
 
 import numpy as np
+import matplotlib.pyplot as plt
+from netCDF4 import num2date
+
 
 PI = 3.141592653589793
 
 from ..io.common import dms_to_d, radar_coords_to_cart
-from netCDF4 import num2date
 
 
 def corner_to_point(corner, point):
--- a/pyart/graph/common.py
+++ b/pyart/graph/common.py
@@ -12,6 +12,19 @@ Common graphing routines.
     corner_to_point
     ax_radius
     sweep_coords_to_cart
+    parse_ax
+    parse_ax_fig
+    parse_vmin_vmax
+    generate_colorbar_label
+    generate_field_name
+    generate_radar_name
+    generate_time_begin
+    generate_filename
+    generate_title
+    generate_vpt_title
+    generate_ray_title
+    generate_colorbar_label
+    set_limits
     _interpolate_range_edges
     _interpolate_elevation_edges
     _interpolate_azimuth_edges
@@ -23,6 +36,7 @@ import numpy as np
 PI = 3.141592653589793
 
 from ..io.common import dms_to_d, radar_coords_to_cart
+from netCDF4 import num2date
 
 
 def corner_to_point(corner, point):
@@ -46,7 +60,7 @@ def corner_to_point(corner, point):
     """
     Re = 6371.0 * 1000.0
     Rc = ax_radius(point[0], units='degrees')
-    #print Rc/Re
+    # print Rc/Re
     y = ((point[0] - corner[0]) / 360.0) * PI * 2.0 * Re
     x = ((point[1] - corner[1]) / 360.0) * PI * 2.0 * Rc
     return x, y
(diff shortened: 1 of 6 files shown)


jjhelmus:
I'm split between making these functions verses creating a **BaseDisplay** class which encapsulates these methods.  Then **RadarDisplay**, **GridDisplay**, etc would use this base class.  I suppose both options could be accomplished using a staticmethod decorator to defined these methods on the **BaseDisplay** class.

pyart/graph/common.py
```diff
@@ -152,3 +152,178 @@ def _interpolate_azimuth_edges(azimuths):
...
+    """
+    time_str = generate_time_begin(radar).isoformat() + 'Z'
+    l1 = "%s %s" % (generate_radar_name(radar), time_str)
+    azim = radar.azimuth['data'][ray]
+    elev = radar.elevation['data'][ray]
+    l2 = "Ray: %i  Elevation: %.1f Azimuth: %.1f" % (ray, azim, elev)
+    field_name = generate_field_name(radar, field)
+    return l1 + '\n' + l2 + '\n' + field_name
+
+def set_limits(self, xlim=None, ylim=None, ax=None):
```
jjhelmus: No need for the self parameter.

gamaanderson: ok, I missed that

gamaanderson:
I'm not inclined to the **BaseDisplay** as most of this functions are specific to radar, it not sure a base class would save much coding. Anyway I'm in favor of static methods, since artview is not build around `RadarDisplay`, but rather around the Radar structure. That is, sometimes we don't have a initialized display to use;

jjhelmus:
After thinking about this for a while I'm in agreement with you @gamaanderson.  Having the function free of a class will offer the most flexibility.  Can you attach these functions as method on **RadarDisplay**, **RadarMapDisplay**, etc.  With that I think this PR can be merged.

gamaanderson:
Done, but if you don't mind, I would like to take the opportunity and make `RadarDisplay_Airborne` subclass of `RadarDisplay`

jjhelmus:
Sounds good, you may want to check with @nguy on this as he is the expert on the **RadarDisplay_Airborne** class.  

Also, [looks like](https://travis-ci.org/ARM-DOE/pyart/jobs/73080738) the line `import matplotlib.pyplot as plt` needs to be added to the command module.  I think that should fix the unit tests.

gamaanderson:
Done, this caused several minor alteration and update some functions. One point that I would like to comment however is that `RadarDisplay_airborne` was considering radar as sweep-less, I disagree with that, I prefer to consider them as radars with one single sweep. Some fixed radar some times also choose to save sweeps in separated files.

[Take a look](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/radardisplay_airborne.py) in how conciser **radardisplay_airborne.py** became

gamaanderson:
ok, I just tested this and got different images (using artview) with the previews and the new version of pyart, I will continue to debug this, but may be someone can I identify it. First the old one and them the new, for me the new one makes more sense, what do you think @nguy?
old
new

nguy:
Sorry for the delay in response. Okay first off @gamaanderson can you tell me which is the "old" and "new"? The first plot above looks more like the airborne radar I would expect.

To comment on the structure of the program. There is some back story (as always) with airborne radar data. Before about 2012, the data were saved differently, in blocks of code containing ray data for a block of time. A lot of hardware changes were made at that time and after much discussion, we decided that saving one *sweep* might be the best way forward. However, many legacy programs deal with data on a time period basis. This mean that a number of sweeps - or more properly a continuously rotating antenna would form a sort of conical volume. Remember that this scanning is intrinsically different than ground-based radar as the platform is moving at non-negligilbe speed. New methods were not (and I would argue still ARE not developed) and these data *look* different than you would expect from a nice stationary ground system. 

I have not dove back in to the data recently. But will try to speak with some people soon about this.
The data go through more geolocation steps due to movement. In addition, aircraft sampling angles constantly change (turns, turbulence, etc.). So the data turns out to be more complicated than the average radar in geometric terms.

IF plans continue as expected a second antenna will be added to this type of radar system. Other airborne platforms (not active scanners) already have up to 5 potential antennas. This complicates matters well beyond anything PyART can currently handle and I don't have a clue at the moment how to handle this. So the airborne class may become outdated.

nguy:
I also failed to mention in the above that I modified ```RadarDisplay``` to create ```RadarDisplay_Airborne``` before final decisions were made to save just one sweep and was working with converting older data into cfradial format. It seems that they (NOAA Aircraft Operations Center) have settled on the individual sweep approach, but at the time there was talk of saving many sweeps into a pseudo volume as mentioned above.

gamaanderson:
Thank you @nguy, you are correct, I found a bug (jet to be fixed) and the first one (the old one) is correct. Based in you explanation, I think you agree in maintaining a [sweep argument](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/radardisplay_airborne.py#L140) to the plot functions, it is them up to the input function the gave adequate separation in sweep. Just remember that Radar does treat data as a block of rays, sweeps are just sub-sets of this rays and some rays may not be in a sweep.

One final thing, does any one know how to implement the [edges argument](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/radardisplay.py#L320-L326) in [here](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/radardisplay_airborne.py#L118-L127) just as it is in [here](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/common.py#L127-L133)?

gamaanderson:
@jjhelmus Now I think this is really ready for merge. 

Just one thing concerns me, I found this [lines](https://github.com/gamaanderson/pyart/blob/display/pyart/graph/common.py#L155-L156), I didn't change them, but I don't think positive elevation is a safe assumption even for ground radars. Actually in Brazil there are some radar in mountains over big cities (e.g. Rio de Janeiro), so that to get a good view of the rain over the city some scans are done pointing downwards (-0.5°).

pyart/graph/radardisplay.py
```diff
@@ -17,11 +17,16 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import netCDF4
+from mpl_toolkits.axes_grid1 import make_axes_locatable
 
-from .common import corner_to_point, radar_coords_to_cart, sweep_coords_to_cart
+from .common import (
```
jjhelmus: Would `from . import common` make sense here? Then all of these function could be called from the common namespace.  I think this might help the readability.

pyart/graph/radardisplay.py
```diff
@@ -98,10 +106,30 @@ def __init__(self, radar, shift=(0.0, 0.0)):
...
+        # sweep start and end indices
+        self.starts = radar.sweep_start_ray_index['data']
+        self.ends = radar.sweep_end_ray_index['data']
+
+        # list to hold plots, plotted fields and plotted colorbars
+        self.plots = []
+        self.plot_vars = []
+        self.cbs = []
+
+    def _calculateLocalization(self, radar):
```
jjhelmus: Use _calculate_localization for this function name.  Camel Case is not recommended for functions in Python (and is not use in Py-ART that I know of).

pyart/graph/radardisplay.py
```diff
@@ -122,21 +150,6 @@ def __init__(self, radar, shift=(0.0, 0.0)):
             warnings.warn('RadarDisplay does not correct for moving platforms')
         self.loc = (lat, lon)
```
jjhelmus: Setting class attributes is typically not recommended outside of the __init__ method, but I think it this case it makes sense.  Sometime in the future I want to make calculations of the x, y, and z coordinates be performed on a per sweep basis just prior to plotting.

pyart/graph/radardisplay.py
```diff
@@ -230,7 +243,7 @@ def plot_ray(self, field, ray, format_str='k-', mask_tuple=None,
 
         """
         # parse parameters
-        ax, fig = self._parse_ax_fig(ax, fig)
+        ax, fig = parse_ax_fig(ax, fig)
```
jjhelmus: This would become `ax, fig  = common.parse_ax_fig(ax, fig)`

pyart/graph/radardisplay.py
```diff
@@ -586,14 +617,47 @@ def plot_range_ring(self, range_ring_location_km, npts=100, ax=None):
...
         r = np.ones([npts], dtype=np.float32) * range_ring_location_km
         x = r * np.sin(theta)
         y = r * np.cos(theta)
-        ax.plot(x, y, 'k-')
+        if lw is None:
+            lw = 2
+        if ls is None:
+            ls = '-'
+        if col is None:
+            col = 'k'
```
jjhelmus: Default values should be specified explicitly in the function signature not by checking for None.

pyart/graph/radardisplay.py
```diff
@@ -586,14 +617,47 @@ def plot_range_ring(self, range_ring_location_km, npts=100, ax=None):
...
+            Linestyle to use for grid lines.
+
+        """
+        ax = parse_ax(ax)
+        if lw is None:
+            lw = 2
+        if ls is None:
+            ls = ':'
+        if col is None:
+            col = 'k'
```
jjhelmus: Again put these directly in the function call.

pyart/graph/radardisplay.py
```diff
@@ -708,7 +774,19 @@ def plot_colorbar(self, mappable=None, field=None, label=None, cax=None,
...
+            orient = 'vertical'
+
+        # Find the axes locations to set colorbar
+        if cax is None:
+            ax = parse_ax(ax)
+            box = make_axes_locatable(ax)
+            if orient == 'vertical':
+                cax = box.append_axes("right", size="3%", pad=0.05)
+            if orient == 'horizontal':
+                cax = box.append_axes("bottom", size="6%", pad=0.50)
```
jjhelmus: Should we expose these parameters in the plot_colorbar function?

pyart/graph/radardisplay.py
```diff
@@ -708,7 +774,19 @@ def plot_colorbar(self, mappable=None, field=None, label=None, cax=None,
             if field is None:
                 field = self.plot_vars[-1]
             label = self._get_colorbar_label(field)
-        cb = fig.colorbar(mappable, ax=ax, cax=cax)
+
+        if orient is None:
+            orient = 'vertical'
+
+        # Find the axes locations to set colorbar
+        if cax is None:
```
jjhelmus: What is the benefit of using `make_axes_locatable` and having fixed padding and sizes for the colorbar as opposed to having matplotlib colorbar function do the resizing?

pyart/graph/radardisplay.py
```diff
@@ -730,43 +808,46 @@ def set_limits(self, xlim=None, ylim=None, ax=None):
...
         """ Label the yaxis with the default label for a field units. """
-        ax = self._parse_ax(ax)
+        ax = parse_ax(ax)
         ax.set_ylabel(self._get_colorbar_label(field))
 
+    def set_aspect_ratio(self, aratio=None, ax=None):
+        """ Set the aspect ratio for plot area. """
+        ax = parse_ax(ax)
+        if aratio is None:
+            ax.set_aspect(0.75)
```
jjhelmus: Move to the function call and rename the argument to aspect_ratio

pyart/graph/radardisplay.py
```diff
@@ -800,6 +881,18 @@ def _label_axes_ppi(self, axis_labels, ax):
         else:
             ax.set_ylabel(y_label)
 
+    def _label_axes_ppi(self, axis_labels, ax):
```
jjhelmus: Repeated function definition

pyart/graph/radardisplay.py
gamaanderson: No idea, I just copy what was already in RadarDisplay_airborne, as it was an enhancement of `plot_colorbar` i didn't really question anything

pyart/graph/radardisplay.py
gamaanderson: what do you mean?

jjhelmus:
A few suggestions for clean up then I think this should be good to merge.  Looking forward to a cleaner and more easily maintainable graph subpackage in Py-ART!

pyart/graph/radardisplay.py
jjhelmus: I do not like the idea of having fixed values for size and pad.  plt.colorbar will resize the parent axes if cax is None will a "good" size and padding.  I would rather use matplotlib defaults than make our own.  @nguy Is there a good reason for the other custom padding and sizing of colorbars?

pyart/graph/radardisplay.py
jjhelmus: Make size and pad parameters in the `plot_colorbar` function.  Not important if we revert to the plt.colorbar function.

pyart/graph/radardisplay.py
gamaanderson: It is already being performed on a per sweep basis, self.x, self.y and self.z are not really being used (at least not inside this code)

pyart/graph/radardisplay.py
gamaanderson: Testing, I found out that when not using `make_axes_locatable` we can't delete the colorbar in artview when the plot size changes, that is probably the reason we are using `colorbar_flag=False`. Unfortunately even with  `make_axes_locatable` there is a problem when changing the field. As this is the case, I will follow @jjhelmus and let matplotlib work with that.

gamaanderson:
changes done!

nguy:
@jjhelmus I know that the RadX package has the calculations. I'd have to look back to find them.

pyart/graph/common.py
```diff
@@ -152,3 +168,210 @@ def _interpolate_azimuth_edges(azimuths):
...
+    time_str = generate_time_begin(radar).isoformat() + 'Z'
+    l1 = "%s %s" % (generate_radar_name(radar), time_str)
+    azim = radar.azimuth['data'][ray]
+    elev = radar.elevation['data'][ray]
+    l2 = "Ray: %i  Elevation: %.1f Azimuth: %.1f" % (ray, azim, elev)
+    field_name = generate_field_name(radar, field)
+    return l1 + '\n' + l2 + '\n' + field_name
+
+
+def generate_colorbar_label(standard_name, units):
```
jjhelmus: This is already defined on line 209.  This definition should be removed.

pyart/graph/radardisplay.py
```diff
@@ -17,11 +17,12 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import netCDF4
+from mpl_toolkits.axes_grid1 import make_axes_locatable
```
jjhelmus: This import is not used and can be removed.

pyart/graph/radardisplay_airborne.py
```diff
@@ -17,12 +17,13 @@
 import netCDF4
 from mpl_toolkits.axes_grid1 import make_axes_locatable
```
jjhelmus: Not used, can be removed

pyart/graph/radardisplay_airborne.py
```diff
@@ -17,12 +17,13 @@
 import netCDF4
```
jjhelmus: Not used, can be removed

pyart/graph/radardisplay_airborne.py
```diff
@@ -17,12 +17,13 @@
...
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
-from .common import corner_to_point, radar_coords_to_cart
+from .radardisplay import RadarDisplay
+from . import common
 from .coord_transform import radar_coords_to_cart_track_relative
 from .coord_transform import radar_coords_to_cart_earth_relative
```
jjhelmus: Not used, can be removed

jjhelmus:
Looks good @gamaanderson, just remove the unneeded marked lines and this should be good to merge.

gamaanderson:
done, lets just wait travis

jjhelmus:
Thanks, [pylint](http://www.pylint.org/) is good at catching these lines.

PR #211, opened 2014-11-16
Advection clean

scollis:
This PR will eventually add the ability to:
Determine advection for grids where nx=ny (done)
shift a grid a number of pixels and trim to a sub domain (done)
create a list of grids that using advective interpolation
aggregate grids using these lists to affect a smooth aggregation avoiding the "Chain of Pearls" effect. 

Initial PR for this feature addition. I am checking to see if my unit tests work with Travis so please do not merge as yet.. I still have a few more features I want to add.

PR #344, opened 2015-08-04
Advection clean

scollis:
A rebase of #211 
This includes all the aims of the advection project except for the integration.. I am still working on the best way to add the sub-grids together to ensure there is no bias 
This PR includes a bundle of features including the ability to determine image shift and use that as an advection speed
The ability to calculate advected grids based on two time steps and an advection speed. 
the ability to add grids. 

It also adds several new objects to the test objects.


scollis:
Tests are failing.. Let me track this down..

jjhelmus:
I'm going to close this since it has been more than six months since the last commit.  Feel free to re-open when development on this feature begins again.

Issue #294, opened 2015-05-15
Projection issue for VERY large mosiac


scollis:
So this is one for the GIS crowd. Using Jonathan's new gridding engine I am doing a massive CONUS wide mesh.. But as you can see form the plot, unless we have radars off the eastern seaboard the mesh does not line up with the map.. 

Code: https://github.com/scollis/conus_mosiac/blob/master/scripts/conus_mosiac.py#L76

![diag_20150513_200000](https://cloud.githubusercontent.com/assets/825351/7662451/2d078736-fb22-11e4-9bc9-470711805e0e.png)

tjlang:
Sorry, Scott, can't resist, but I fixed your problem in 7 lines of code. :)
![screen shot 2015-05-15 at 10 22 12 pm](https://cloud.githubusercontent.com/assets/3036387/7664507/0c4c9afc-fb51-11e4-9b14-82747ab6a7ff.png)
On a more serious note, yeah if the map projection is based off a single point you will get huge errors near the edges. I would need to dig up how MRMS fixed this, it's probably in the Zhang et al. (2011) paper or something.

tjlang:
Yeah, reading the paper it sounds like they map each individual radar to its own cylindrical equidistant projection (0.01-deg resolution), then they merge all those individual grids together onto a national projection with the same resolution, and where radars overlap they use an exponential weighting function to merge the data. A lot of intervening QC and other corrections are done as well. 

I did something similar, but far less fancy, for this pre-MRMS/NMQ paper:
http://onlinelibrary.wiley.com/doi/10.1029/2009JA014500/full

So the basic gist is map individual radars locally, then merge after the fact.

scollis:
Thanks @tjlang I admit significant ignorance when it comes to projection issues... It would be nice to resolve this so Py-ART's gridding paridigm and map plotting works..
A fundamental question is are the grid points in the right place? Is the issue in the calculation of the displacement of the radars from the origin of the grid?

jjhelmus:
My guess is that the issue is the projection, or rather lack of a projection, used in calculating displacement of the radar location from the grid origin.  This is done by the [`corner_to_point`](http://arm-doe.github.io/pyart-docs-travis/dev_reference/generated/pyart.graph.common.corner_to_point.html?highlight=corner_to#pyart.graph.common.corner_to_point) function which I believe is not valid over large distances.  I'll investigate on Monday.

scollis:
[...]

Issue #335, opened 2015-07-21
More robust masking procedure for plotting


nguy:
I've added a more robust masking option in the form of a dictionary to [my fork](https://github.com/nguy/pyart). It is only added to the plot_ppi currently, as I'd like to get feedback before issuing a PR. 

An example of functionality can be found in this [notebook ](https://github.com/nguy/Notebooks/blob/master/pyart_masking_example.ipynb).

I've been wanting to add this for a while, and now that ARTView is moving ahead I see this as potentially useful for future development there. This way we can just include a plugin or such on that end for pure masking and I believe it will come in handy when editing.

If it does look good, I'll go ahead and implement in the other plot routines and issue a PR.

nguy:
One other thing: I'm not a fan of having mask_outside turned on by default. It can confuse data analysis if vmin and vmax are copied from an example or such.
Thoughts?

scollis:
I concur.. I was just thinking this today.. I will submit an issue


On 7/21/15 5:53 PM, Nick wrote:
>
> One other thing: I'm not a fan of having mask_outside turned on by 
> default. It can confuse data analysis if vmin and vmax are copied from 
> an example or such.
> Thoughts?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/ARM-DOE/pyart/issues/335#issuecomment-123502386>.
>

-- 
I ride for Parkinsons research http://www.events.org/sponsorship.aspx?id=51573

nguy:
I can make this change with the PR if you guys like the masking.

gamaanderson:
I like the idea of improving the masking process, but it seems a little confusing, may be is just a question of finding the way of explaining it. An other idea would be to use a [gatefilter](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.correct.GateFilter.html#pyart.correct.GateFilter).

scollis:
Having pyart.graph.RadarDisplay* accept gatefilter is an excellent idea


On 7/22/15 9:26 AM, Anderson wrote:
>
> I like the idea of improving the masking process, but it seems a 
> little confusing, may be is just a question of finding the way of 
> explaining it. An other idea would be to use a gatefilter 
> <http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.correct.GateFilter.html#pyart.correct.GateFilter>.
>
> —
> Reply to this email directly or view it on GitHub 
[...]

Issue #285, opened 2015-04-24
Replace the Grid class layout


jjhelmus:
The current layout of the [**Grid** class](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.core.Grid.html#pyart.core.Grid) is historical and ad-hoc, it does not follow a formal standard.  A more formal layout of the class should be used which conforms to the [CF conventions](http://cfconventions.org/).  The section of the CF convention which are relavent are [5.6 Horizontal Coordinate Reference Systems, Grid Mappings, and Projections](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#grid-mappings-and-projections) and [Appendix F. Grid Mappins](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#appendix-grid-mappings).

A specification for [Gridspec](http://www.gfdl.noaa.gov/~vb/gridstd/gridstd.html#gridstdse2.html) standard has been put forward as a possible method for describing gridding data within the CF conventions.  This standard is aimed at storing model data but may be adaptable for storing gridding radar data.

There is also the [UGRID Conventions](https://github.com/ugrid-conventions/ugrid-conventions) for specifying unstructured grids.

gamaanderson:
I've just read that Gridspec documentation, and it seemed for me overly complicated for our simple use. Also it date from 2007 and does not seem to have gained much momentum since them. I don't see how the class Gridspec in matplotlib relate to that, may be is just a coincidence.

CF, on the other hand, is more promising. To make it more friendly, one could add functions that interpret that data and return more useful python variable, for instance a `Proj` instance from `pyproj`.

jjhelmus:
Reading through GridSpec it does seem more complicated than we need.  There seems to be a number example of representing uniform latitude/longitude grids in NetCDF but I have not found one which covers uniform Cartesian grids which is what Py-ART is producting.  Although this does correspond to a lat/lon is is not uniformly spaced.  I'll check into what is required for a `Proj` instance as see if any of the modelers I work with know of any other Grid standards.

gamaanderson:
Have you seen this sites:
http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/reference/StandardCoordinateTransforms.html
http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/tutorial/CoordinateAttributes.html

[...]

PR #346, opened 2015-08-05
ENH: add label, title etc. to gridmapdisplay

gamaanderson:
Several additions to gridmapdisplay, making it more similar to radardisplay. I'm not sure how some things here will be received, so while revising this keep in mind my last comment to issue  #294.

Also check pyart example `plot_grid_three_panel.py` for eventual bugs

### DIFF
--- a/pyart/graph/gridmapdisplay.py
+++ b/pyart/graph/gridmapdisplay.py
@@ -19,6 +19,8 @@ import matplotlib.pyplot as plt
 from mpl_toolkits.basemap import Basemap
 from mpl_toolkits.basemap import pyproj
 
+from . import common
+
 
 class GridMapDisplay():
     """
@@ -62,9 +64,8 @@ class GridMapDisplay():
         # set up the projection
         lat0 = grid.axes['lat']['data'][0]
         lon0 = grid.axes['lon']['data'][0]
-        self.proj = pyproj.Proj(proj='lcc', datum='NAD83',
-                                lat_0=lat0, lon_0=lon0,
-                                x_0=0.0, y_0=0.0)
+        self.proj = pyproj.Proj(proj='aeqd', datum='NAD83',
+                                lat_0=lat0, lon_0=lon0)
 
         # determine grid latitudes and longitudes.
         x_1d = grid.axes['x_disp']['data']
@@ -75,6 +76,7 @@ class GridMapDisplay():
         # set attributes
         self.mappables = []
         self.fields = []
+        self.origin = 'origin'
         self.basemap = None
 
     def plot_basemap(self, lat_lines=None, lon_lines=None,
@@ -107,38 +109,26 @@ class GridMapDisplay():
             Axis to add the basemap to, if None the current axis is used.
 
         """
+        # make basemap
+        self._make_basemap(resolution, area_thresh, auto_range,
+                           min_lon, max_lon, min_lat, max_lat, ax)
+
         # parse the parameters
-        ax = self._parse_ax(ax)
         if lat_lines is None:
             lat_lines = np.arange(30, 46, 1)
         if lon_lines is None:
             lon_lines = np.arange(-110, -75, 1)
 
-        # determine map region
-        if auto_range:
-            max_lat = self.grid_lats.max()
-            max_lon = self.grid_lons.max()
-            min_lat = self.grid_lats.min()
-            min_lon = self.grid_lons.min()
+        self.basemap.drawcoastlines(linewidth=1.25)
+        self.basemap.drawstates()
+        self.basemap.drawparallels(lat_lines, labels=[True, False, False, False])
+        self.basemap.drawmeridians(lon_lines, labels=[False, False, False, True])
 
-        if self.debug:
-            print("Maximum latitude: ", max_lat)
-            print("Maximum longitude: ", max_lon)
-            print("Minimum latitude: ", min_lat)
-            print("Minimum longitute: ", min_lon)
-
-        # plot the basemap
-        basemap = Basemap(llcrnrlon=min_lon, llcrnrlat=min_lat,
-                          urcrnrlon=max_lon, urcrnrlat=max_lat,
-                          projection='mill', area_thresh=area_thresh,
-                          resolution=resolution, ax=ax)
-        basemap.drawcoastlines(linewidth=1.25)
-        basemap.drawstates()
-        basemap.drawparallels(lat_lines, labels=[True, False, False, False])
-        basemap.drawmeridians(lon_lines, labels=[False, False, False, True])
-        self.basemap = basemap
-
-    def plot_grid(self, field, level=0, vmin=None, vmax=None, cmap='jet'):
+    def plot_grid(self, field, level=0, vmin=None, vmax=None, cmap='jet',
+                 mask_outside=True, title=None, title_flag=True,
+                 axislabels=(None, None), axislabels_flag=False,
+                 colorbar_flag=True, colorbar_label=None,
+                 colorbar_orient='vertical', edges=True, ax=None, fig=None):
         """
         Plot the grid onto the current basemap.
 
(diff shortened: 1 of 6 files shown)


pyart/graph/gridmapdisplay.py
```diff
@@ -62,9 +64,8 @@ def __init__(self, grid, debug=False):
         # set up the projection
         lat0 = grid.axes['lat']['data'][0]
         lon0 = grid.axes['lon']['data'][0]
-        self.proj = pyproj.Proj(proj='lcc', datum='NAD83',
-                                lat_0=lat0, lon_0=lon0,
-                                x_0=0.0, y_0=0.0)
+        self.proj = pyproj.Proj(proj='aeqd', datum='NAD83',
```
gamaanderson: I has having "huge" error with lamber, so I changed the data projection to azimuthal equidistant

pyart/graph/gridmapdisplay.py
```diff
@@ -328,11 +653,65 @@ def _check_basemap(self):
...
+
+        # determine plot center
+        lat_0 = self.grid.axes['lat']['data'][0]
+        lon_0 = self.grid.axes['lon']['data'][0]
+
+        # plot the basemap
+        basemap = Basemap(llcrnrlon=min_lon, llcrnrlat=min_lat,
+                          urcrnrlon=max_lon, urcrnrlat=max_lat,
+                          lat_0=lat_0, lon_0=lon_0,
+                          projection='tmerc', area_thresh=area_thresh,
```
gamaanderson: I prefer mercator to miller as most maps (e.g google maps) use it and it also perverse format. `'tmerc'` work better than `'merc'`, setting the scale from origin and it would also set the axes origin if it were not for [this basemap bug](https://github.com/matplotlib/basemap/issues/192)

pyart/graph/gridmapdisplay.py
```diff
@@ -348,31 +727,144 @@ def _find_nearest_grid_indices(self, lon, lat):
...
+            Plot title.
+
+        """
+        return common.generate_latitude_slice_title(self.grid, field, level)
+
+    ##########################
+    #      get methods       #
+    ##########################
+
+    def get_basemap(self):
```
gamaanderson: I will be using basemap in artview to convert from map-projection to data-projection, those this need to be public

gamaanderson:
about the travis error, it is because I'm not raising the "no basemap" error but rather running it: if basemap has no mandatory arguments why no run it for the user if he didn't do it him self?

jjhelmus:
I think that particular test can be modified, it seem reasonable to create a default basemap instance if one does not exist prior to a call to `plot_grid`.  Should the same behavior be followed in `plot_crosshairs`?  I vote yes.  If so then the `_check_basemap` can be removed completely.

pyart/graph/gridmapdisplay.py
```diff
@@ -107,38 +109,28 @@ def plot_basemap(self, lat_lines=None, lon_lines=None,
...
-    def plot_grid(self, field, level=0, vmin=None, vmax=None, cmap='jet'):
+        self.basemap.drawcoastlines(linewidth=1.25)
+        self.basemap.drawstates()
+        self.basemap.drawparallels(lat_lines,
+                                   labels=[True, False, False, False])
+        self.basemap.drawmeridians(lon_lines,
+                                   labels=[False, False, False, True])
+
+    def plot_grid(self, field, level=0, vmin=None, vmax=None, cmap='jet',
+                  mask_outside=True, title=None, title_flag=True,
```
jjhelmus: I think the default for mask_outside should be `False` as in the **RadarDisplay** class.

pyart/graph/gridmapdisplay.py
jjhelmus: Can ARTView access the basemap using attribute access using display.basemap or getattr(display, 'basemap') so that this can be a private function?  If not, that is fine also just want to check.

pyart/graph/gridmapdisplay.py
gamaanderson: the default is `True` in [ppi](https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/radardisplay.py#L265) and [rhi](https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/radardisplay.py#L365) !?! But if it is to be `False` I have no objection

gamaanderson:
I also vote yes for default `Basemap` to `plot_crosshairs`

pyart/graph/gridmapdisplay.py
jjhelmus: We talked about changing it to False in all instances in #335 but no ticket was created for the task.  For the time being keep this as `True` to be consistent with the other plotting function (except plot_ray) which happened to be the first one I found.  Oops

pyart/graph/gridmapdisplay.py
gamaanderson: Yes, we can access the attribute, but as there is no distinction between public and private attributes someone may want in the future to change `basemap` (e.g to `base_map`). The use of a public function is kind of a safety.

pyart/graph/gridmapdisplay.py
jjhelmus: Issue #351 created to track changing mask_outside to `False` in all instances.

jjhelmus:
Should the same projections be used to create the basemap instance be used to project the latitudes and longitude positions?  Also, if the projection name can be made a parameter to the `__init__` function or the `_make_basemap function` that would be great.

I do not think the **grid_lons** and **grid_lats** class attributes are needed, they can be calculated in the `plot_grid` and `_make_basemap` functions.  This required a bit of work, do not feel obligated to refactor these out if it does not interest you.  I think the same can be said for **self.prog** but is a bit more tricky.

gamaanderson:
I didn't understand your first question, could you clarify it.

passing the projection to `plot_basemap` and them to `_make_basemap` would be nice.

I agree with removing **grid_lons** and **grid_lats**.

Ultimately **self.proj** should be provided by the grid it self, while that does not happen we could let as it is.

scollis:
We have a ticket on having a proj class.. I will update this one soon.. 
and I hope to get to fixing the whole projection issue next week


On 8/6/15 3:00 PM, Anderson wrote:
>
> I didn't understand your first question, could you clarify it.
>
> passing the projection to |plot_basemap| and them to |_make_basemap| 
> would be nice.
>
> I agree with removing *grid_lons* and *grid_lats*.
>
> Ultimately *self.proj* should be provided by the grid it self, while 
> that does not happen we could let as it is.
>
> —
> <https://github.com/ARM-DOE/pyart/pull/346#issuecomment-128487046>.
>

-- 

jjhelmus:
In `__init__` the aeqd projection is used to create self.proj and then in `_make_basemap` a tmerc projection is used.  Should these be the same?

gamaanderson:
@scollis, we are kind of going back-wards, we should first solve issue #285, then #294 and at last solve the graphing. Of course, I was the one start with the last, so I can't really complain about that.

scollis:
yep.. but I will make a general proj class that returns an instance to 
do projections..
It will default to a spherical one and if it is available it can use 
proj or gdal


On 8/6/15 3:09 PM, Anderson wrote:
>
> @scollis <https://github.com/scollis>, we are kind of going 
> back-wards, we should first solve issue #285 
> <https://github.com/ARM-DOE/pyart/issues/285>, then #294 
> <https://github.com/ARM-DOE/pyart/issues/294> and at last solve the 
> graphing. Of course, I was the one start with the last, so I can't 
> really complain about that.
>
> —
> <https://github.com/ARM-DOE/pyart/pull/346#issuecomment-128491185>.
>

-- 

gamaanderson:
@jjhelmus  We could make a plot in aeqd, but you will not find maps in aeqd, at maximum coast- and country-lines. We could consider making a plot without map, something like `GridDisplay`, but that is an other issue.

jjhelmus:
I do not think we should use the aeqd projection to make a map, rather we should use tmerc to perform the calculation of the grid point locations in latitude and longitude.  Or in general use the same projection for transforming the grid points as is used in the map.  If these are not the same I would assume that there may be small shifts in the locations of grid points.  I'll do some tests and see if this is true.

gamaanderson:
if the grid as originally created with one projection, using **proj** in an other projection can create a error, a big one if the grid is big. It is for the grid creating function to inform what projection is being considered. And in the lack of a projection we should define which one is the standard (my vote go for aeqd), but any way this is also not for the `GridMapDisplay` to decide.

jjhelmus:
Right, I still am struggling with the idea that a grid must be created with a projections, but I'm sure it must.      With that then the two different projections make sense.  I need to do some more reading on the topic in order to better understand this.

gamaanderson:
Just added the alteration we were discussing:

* add projection options to `plot_basemap` and `_makea_basemap`
* remove pre-calculation of `self.grid_lons` e `self.grid_lats`
* add default basemap to `plot_crosshairs`, remove `_check_basemap` and the travis check for that

also I reconsidered and changed default from tranverse mercator to mercator, to match the traditional global maps, just need to set lat_ts to get the right scale.

jjhelmus:
The projection and kwargs are not being passed to the `_make_basemap` method.  **RadarMapDisplay** does a basic check on some of the parameters to see what arguments should be passed to Basemap (not all projects work with providing corner data), something like this would make sense here, see lines [221-237](https://github.com/ARM-DOE/pyart/blob/54d5f00d5c25db7f61edbcd23b4689b68556daec/pyart/graph/radarmapdisplay.py#L221-L237).  

I'm hesitant to change the default to mercator as the distortions at high latitudes are large.  One of ARM's majors sites is in Alaska and displaying the data on mercator projected maps is awful.  The default in the **RadarMapDisplay** class is Lambert conformal conic ('lcc').

gamaanderson:
The problem with Mercator is that it change the scales, but the lat_ts (latitude of true scale) correct this.

gamaanderson:
will Basemap not just ignore unneeded argument? it is necessary to test `using_corners`?

jjhelmus:
You are right on both points. I'm still learning about Basemap.  I think setting lat_ts to lat_0 makes sense. With that change defaulting to Mercator is fine with me.  Passing extra arguments with the default None value to Basemap also works, so the test for `using_corners` is not needed in the **RadarMapDisplay** class.

gamaanderson:
Does this last commit (2680650) cover this points? I not sure I understood everything correctly.

jjhelmus:
The Basemap defaults should be provide as non-None values for either width and height parameter or llcrnrlon, llcrnlat, urcrnrlon, urcrnrlat parameters but not both.  I cannot tell from the Basemap documentation which set of parameters would be used if both a supplied.  I would suggest calculating height and width if either is None and adding them to the Basemap kwargs is the corner parameters are not supplied.

gamaanderson:
Which parameter is used will depend on the projection, but for `merc` the corners are used (it warns that on execution)

jjhelmus:
This issue with having defaults for both sets is that a user would need to set all six to override the defaults.  I think most users would expect that setting one set would override the defaults of the other set.

gamaanderson:
done

jjhelmus:
Looks good to me!  Anything else needed before this is merged?

gamaanderson:
no, you can merge it.

jjhelmus:
Thanks for all the work @gamaanderson.  Glad to see this one merged.

PR #348, opened 2015-08-06
ENH: GateFilter based filter in dealias_fourdd

jjhelmus:
Add a gatefilter parameter to the dealias_fourdd function which is used to specify which gates should be included or excluded from the dealiasing algorithm.  Remove parameters which specified this filtering by radar fields.


jjhelmus:
This PR supersedes PR #308.

jjhelmus:
Merging this PR closes #308

PR #360, opened 2015-08-24
ENH: Mapping routines accept GateFilter instances to specify gates to include in interpolation

jjhelmus:


jjhelmus:
This should address issue #318.  I will check manually to see if Py-ART compiles on Windows once merged.

PR #363, opened 2015-08-28
Add GateFilter to RadarDisplay

nguy:
Add GateFilter instance (pyart.correct.GateFilter) as a keyword input to ```plot_ppi``` and ```plot_rhi``` in  ```radardisplay.py```.
If acceptable, it could be extended to the other display code. It's a relatively simple change. It will make building the ARTView masking procedure much easier and allow users to see changes of applying a masking filter to data.

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -321,6 +321,9 @@ class RadarDisplay(object):
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             plotted.
+        gatefilter : GateFilter Instance
+            pyart.correct.GateFilter instance. None will result in no
+            gatefilter mask being applied to data.
         filter_transitions : bool
             True to remove rays where the antenna was in transition between
             sweeps from the plot.  False will include these rays in the plot.
@@ -423,6 +426,9 @@ class RadarDisplay(object):
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             not plotted.
+        gatefilter : GateFilter Instance
+            pyart.correct.GateFilter instance. None will result in no
+            gatefilter mask being applied to data.
         filter_transitions : bool
             True to remove rays where the antenna was in transition between
             sweeps from the plot.  False will include these rays in the plot.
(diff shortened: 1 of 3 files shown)


pyart/graph/radardisplay.py
```diff
@@ -320,6 +321,9 @@ def plot_ppi(self, field, sweep=0, mask_tuple=None, vmin=None, vmax=None,
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             plotted.
+        gatefilter : GateFilter Instance
+            pyart.correct.filters.GateFilter instance. None will result in no
```
jjhelmus: Just use "GateFilter instance" here, the namespace is not important

pyart/graph/radardisplay.py
```diff
@@ -320,6 +321,9 @@ def plot_ppi(self, field, sweep=0, mask_tuple=None, vmin=None, vmax=None,
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             plotted.
+        gatefilter : GateFilter Instance
```
jjhelmus: Remove Instance

pyart/graph/radardisplay.py
```diff
@@ -421,6 +427,9 @@ def plot_rhi(self, field, sweep=0, mask_tuple=None, vmin=None, vmax=None,
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             not plotted.
+        gatefilter : GateFilter Instance
+            pyart.correct.filters.GateFilter instance. None will result in no
```
jjhelmus: Some as above

pyart/graph/radardisplay_airborne.py
```diff
@@ -221,6 +221,9 @@ def plot_sweep_grid(self, field, sweep=0, mask_tuple=None, vmin=None,
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             plotted.
+        gatefilter : GateFilter Instance
+            pyart.correct.filters.GateFilter instance. None will result in no
+            gatefilter mask being applied to data.
```
jjhelmus: Same as above

jjhelmus:
Adjust the docstring and I think this should be good to go.

nguy:
Changes also made in the ```radarmapdisplay.py``` and ```radardisplay_airborne```

jjhelmus:
Looks good, will merge once AppVeyor gives the green light.  Thanks for all the work @nguy!

PR #369, opened 2015-09-01
BUG: 1-byte width field from Sigmet files returns in m/s

### LINKED ISSUE #367: Sigmet: Units conversion for 1-byte spectrum width
According to the comments in the source, which are consistent with the IRIS docs, the 1-byte spectrum width values should be multiplied by the Nyquist velocity.

 https://github.com/ARM-DOE/pyart/blob/master/pyart/io/_sigmetfile.pyx#L709

This doesn't happen automatically, although according to line 705, a similar multiplication is done to velocity in get_data (actually, in read_data, line 216).

Would the maintainers be amenable to replicating the VEL method for WIDTH so that the conversion is done for the user?

jjhelmus:
closes #367


jjhelmus:
Closed, PR #372 contains a corrected version of this fix.

PR #377, opened 2015-09-22
Gatefilter direct acess

gamaanderson:
Hello, in artview we have a tool that allows the user select regions with the mouse. One of the uses of this tools could be selecting bad-data to be removed. We will proceed for doing that with a gatefilter, but for such we need to direct say which gates should be filtered, not a value comparisons. For such we propose adding the follow two methods to `GateFilter`.


jjhelmus:
@gamaanderson Would it be possible to add some tests for these new methods?

pyart/filters/gatefilter.py
```diff
@@ -320,6 +320,39 @@ def exclude_invalid(self, field, exclude_masked=True, op='or'):
...
+            excluding gates where the desired effect is to exclude gates which
+            meet any of the conditions. 'and', the default for include
+            methods, is typically desired when building up a set of conditions
+            where the desired effect is to include gates which meet any of the
+            conditions.  Note that the 'and' method MAY results in including
+            gates which have peviously been excluded because they were masked
+            or invalid.
+
+        """
+        fdata = self._radar.fields.values()[0]['data']
```
jjhelmus: This will not work in Python 3 as `values` returns a dict_value object which does not support indexing.   If you cast to a list it will work::
``` Python
fdata = list(self._radar.fields.values())[0]['data']
```

jjhelmus:
CI is working again, restarted the builds.

jjhelmus:
Ping @gamaanderson.  Wanted to check that you saw the comments, for some reason I was not getting updates on this issue and was worried you might be missing them also.  If you prefer I can add unit tests to this tomorrow.

gamaanderson:
Ping Back! Thank you, sorry I saw the comments but was busy at the time and end up forgetting later (although t I knew I was forgetting some thing I had to do!). I will get to this, so problem.

gamaanderson:
That should done, just don't know if I must do anything else to add the tests to travis.

jjhelmus:
Looks good to me, thanks for the code.  Merging.

PR #387, opened 2015-09-30
ENH: write UF files

jjhelmus:
Add the `pyart.io.write_uf` function for creating UF files from **Radar** instances.


scollis:
Nice work Jonathan!

jjhelmus:
rebased this incorrectly, will fix and submit a new PR.

Issue #186, opened 2014-09-11
UF File Writing


josephhardinee:
This is mostly as a placeholder to reduce duplicated work. 
Py-Art needs the ability to write UF files. I have most of a PR ready that implements this, just gotta iron out a few things and ensure RSL compatibility.

jjhelmus:
@josephhardinee Any updates on this work?

josephhardinee:
It kind of got strangled by dissertation writing. Hopefully I should be able to revisit it very soon.

gamaanderson:
Is some one working in this issue? Responding from a request from Simepar, I would like to get to it.

jjhelmus:
@josephhardinee mentioned he had some code for this at the AMS radar meeting.  

Since I added support for reading UF and am now well versed in the format I can try to code something up this afternoon.

jjhelmus:
In progress work on a UF writing appears in the [jjhelmus/pyart/uf_write](https://github.com/jjhelmus/pyart/tree/uf_writer) branch.  This branch will rebased when complete.

jjhelmus:
Closed with the merging of #388

gamaanderson:
Nice work @jjhelmus

@TulipaSilva can you please confirm that TITAN is recognizing this UF-files. If you don't have the time, ask someone else, Beneti is aware of that.

jjhelmus:
Let me know if there are any issue with UF writing that you find.  Code coverage is 100% but it could still use some testing with real world files.  Two other things to keep in mind are:

* Py-ART is writing UF files with 2-byte padding around each record, this is consistent with the files RSL creates but may not be supported by all software.  If you need 1-byte or no padding files open an issue and I should be able to add support for these files without too much work.

* UF writing in Py-ART relatively slow, probably no good way around this without some serious optimization.

gamaanderson:
BUG: in [line 265](https://github.com/ARM-DOE/pyart/blob/master/pyart/io/uf_write.py#L265) it should be ` / 1000` and not ` * 1000`, this is causing overflow and consequent error while packing. Also in line 266 I believe `'range_start_m'` is just the decimal part of the range, otherwise a start range of `33 km` will overflow and I don't see a reason for repeating the information. Sorry it took so long for reporting this bug.

[...]

PR #388, opened 2015-09-30
ENH: write UF files

jjhelmus:
Add a `pyart.io.write_uf` function for writing UF files from **Radar** instances.

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -39,7 +39,7 @@ if [[ $PYTHON_VERSION == '2.7' ]]; then
 
     # wradlib and dependencies
     conda install --yes sphinx gdal numpydoc h5py
-    pip install sphinx-rtd-theme
+    conda install --yes sphinx_rtd_theme
     pip install sphinxcontrib-bibtex
     pip install xmltodict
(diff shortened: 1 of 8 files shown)


jjhelmus:
Much better rebase.  Full commit history can be found in [jjhelmus/pyart/tree/uf_writer](https://github.com/jjhelmus/pyart/tree/uf_writer)

jjhelmus:
closes #186

gamaanderson:
BUG: in [line 265](https://github.com/ARM-DOE/pyart/blob/master/pyart/io/uf_write.py#L265) it should be ` / 1000` and not ` * 1000`, this is causing overflow and consequent error while packing. Also in line 266 I believe `'range_start_m'` is just the decimal part of the range, otherwise a start range of `33 km` will overflow and I don't see a reason for repeating the information. Sorry it took so long for reporting this bug.

Also I would have found more comfortable, fields not present in `uf_field_names` to be excluded from writing, and not causing an KeyError: in [line 150](https://github.com/ARM-DOE/pyart/blob/master/pyart/io/uf_write.py#L265). But you probably have your reasons for not doing so.

PR #396, opened 2015-10-08
A quick fix for single sweep mode type files that slightly violate CF…

josephhardinee:
This is a small fix for files that slightly break cf/convention by having 1 sweep_mode per file. It just catches the exception and grabs the single value of sweep mode instead.

### DIFF
--- a/pyart/io/cfradial.py
+++ b/pyart/io/cfradial.py
@@ -180,7 +180,11 @@ def read_cfradial(filename, field_names=None, additional_metadata=None,
         ray_angle_res = None
 
     # first sweep mode determines scan_type
-    mode = str(netCDF4.chartostring(sweep_mode['data'][0]))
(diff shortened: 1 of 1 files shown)


jjhelmus:
@josephhardinee Did you intend to close this PR?

josephhardinee:
Yes, it added a different bug. I will fix it and then reopen.

jjhelmus:
Good to know, I'll look into the [NumPy 1.10.0 issue with Travis CI](https://travis-ci.org/ARM-DOE/pyart/jobs/84398088).  Looks like a few conda packages might need to be recreated.

jjhelmus:
No idea why the Travis CI test fails with numpy 1.10.0 as it seems to be fine now.  I'll investigate more it keeps reoccurring.

PR #399, opened 2015-10-09
Update chl.py documentation

josephhardinee:
Just removed some documentation artifacts from the mdv reader this was based on.


jjhelmus:
Looks good.  Somewhat surprised that the old docstring didn't cause errors in Sphinx.

PR #402, opened 2015-10-10
add mdv_field_names to write_mdv_grid

gamaanderson:
As this function is founding increasing use, I thought `mdv_field_names` and `field_write_order`  would be an important addition. Since I'm already here, is there something else that could be done?

### DIFF
--- a/pyart/io/mdv_grid.py
+++ b/pyart/io/mdv_grid.py
@@ -33,8 +33,8 @@ def write_grid_mdv(filename, grid, mdv_field_names=None,
     Write grid object to MDV file.
 
     Create a MDV file containing data from the provided grid instance.
-    The MDV file will contain instrument parameters from the following
-    dictionaries if they contained in grid.metadata:
+    The MDV file will contain parameters from the following keys if they are
+    contained in grid.metadata:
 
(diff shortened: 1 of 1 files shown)


jjhelmus:
@gamaanderson these changes look good to me.  Would a `exclude_fields` parameter make sense?  The `write_uf` function has this parameter but it is of limited use.

gamaanderson:
I thought about that, but different then UF, here I can use the original fields names, so don't need for the `radar_field_names` and a consequence `exclude_fields` is redundant, fields can be excluded with the `mdv_field_names` or with the `field_write_order`.
The only good reason to add `exclude_fields` seems to be for consistence with other functions, but my impression is that pyart has non intention to be ABSOLUTE consistence, options are added as long as they are relevant. Am I right?

jjhelmus:
Agreed,  I don't think there is a need for the `exclude_fields` parameter.  There is no need to be absolutely consistent in all reading/writing options.  

Then, I think this is good to merge, yes?

gamaanderson:
yes, it is good to merge.

jjhelmus:
Great, thanks for the contribution!

PR #403, opened 2015-10-11
Made reader more resillient to missing sweep_number parameter

josephhardinee:
Added an error recovery measure for missing sweep_number parameters. I believe the parameter is fully redundant as long as fixed angles or sweep_start_ray_index exists. Also fixed a minor spelling error.

### DIFF
--- a/pyart/io/cfradial.py
+++ b/pyart/io/cfradial.py
@@ -161,11 +161,19 @@ def read_cfradial(filename, field_names=None, additional_metadata=None,
         altitude_agl = None
 
     # 4.7 Sweep variables -> create atrribute dictionaries
-    sweep_number = _ncvar_to_dict(ncvars['sweep_number'])
     sweep_mode = _ncvar_to_dict(ncvars['sweep_mode'])
     fixed_angle = _ncvar_to_dict(ncvars['fixed_angle'])
     sweep_start_ray_index = _ncvar_to_dict(ncvars['sweep_start_ray_index'])
     sweep_end_ray_index = _ncvar_to_dict(ncvars['sweep_end_ray_index'])
+ 
+    if 'sweep_number' in ncvars:
+        sweep_number = _ncvar_to_dict(ncvars['sweep_number'])
+    else:
+        nsweeps = len(sweep_start_ray_index['data'])
+        sweep_number = filemetadata('sweep_number')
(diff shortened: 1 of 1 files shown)


jjhelmus:
Thanks for the code @josephhardinee.  Rather than using static metadata it would be best to retrieve this from the Py-ART configuration file using `filemetadata('sweep_number')`.  That way that users can specify unique metadata if they choose.  Also, can the number of sweeps be determined inside the else statement to handle cases of multi-sweep files which are missing the sweep_number variable?  I don't know if such files exist or are common but it is not hard to support.  I can imaging something like:

``` Python
if 'sweep_number' in ncvars:
    sweep_number = _ncvar_to_dict(ncvars['sweep_number'])
else:
    nsweeps = len(sweep_start_rayt_index['data'])
    sweep_number = filemetadata('sweep_number')
    sweep_number['data'] = np.arange(nsweeps, dtype='float32')
```

jjhelmus:
Also it might be good to issue a warning when the file does not contain a sweep_number variable since such a file does not meet CfRadial conventions.

pyart/io/cfradial.py
```diff
@@ -161,11 +161,19 @@ def read_cfradial(filename, field_names=None, additional_metadata=None,
...
     sweep_start_ray_index = _ncvar_to_dict(ncvars['sweep_start_ray_index'])
     sweep_end_ray_index = _ncvar_to_dict(ncvars['sweep_end_ray_index'])
+ 
+    if 'sweep_number' in ncvars:
+        sweep_number = _ncvar_to_dict(ncvars['sweep_number'])
+    else:
+        nsweeps = len(sweep_start_ray_index['data'])
+        sweep_number = filemetadata('sweep_number')
+        sweep_number['data'] = np.arange(nsweeps, dtype='float32')
+        print("Warning: File violates CF/Radial convention. Missing sweep_number variable")
```
jjhelmus: warnings.warn("Message")

And then add an import for warnings to the top level imports

josephhardinee:
Want me to change the other error reporting in the reader while I'm at it?

josephhardinee:
Line 691 that is.

jjhelmus:
Yes, changing the other print statement to a warning sounds good.

jjhelmus:
Merged, thanks for the contribution!

Issue #398, opened 2015-10-09
Plot azimuthal cross section of ppi


jjhelmus:
As suggested by @nguy on the mailing list it would be convenient if a azimuthal cross section of a PPI could be visualized without needing to first create a RHI-like radar instance using `pyart.util.cross_section_ppi`.  **RadarDisplay** method names of `plot_ppi_azimuth` and `plot_vert_azimuth` were suggested.

nguy:
Just submitted a PR that addresses this issue.

jjhelmus:
Thanks @nguy.  Will close this issue once PR #405 is merged.

PR #405, opened 2015-10-21
Add plot azimuthal cross-section to radardisplay class.

nguy:
This PR is in response to Issue #398

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -321,7 +321,7 @@ class RadarDisplay(object):
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             plotted.
-        gatefilter : GateFilter 
+        gatefilter : GateFilter
             GateFilter instance. None will result in no gatefilter mask being
             applied to data.
         filter_transitions : bool
@@ -427,7 +427,7 @@ class RadarDisplay(object):
             coordinates themselved as the gate edges, resulting in a plot
             in which the last gate in each ray and the entire last ray are not
             not plotted.
-        gatefilter : GateFilter 
+        gatefilter : GateFilter
             GateFilter instance. None will result in no gatefilter mask being
             applied to data.
         filter_transitions : bool
@@ -588,13 +588,13 @@ class RadarDisplay(object):
                                orient=colorbar_orient,
                                field=field, ax=ax, fig=fig)
 
-    def plot_azimuth_to_rhi(self, field, target_azimuth, 
+    def plot_azimuth_to_rhi(self, field, target_azimuth,
                             mask_tuple=None, vmin=None, vmax=None,
-                            cmap='jet', mask_outside=False, 
+                            cmap='jet', mask_outside=False,
                             title=None, title_flag=True,
                             axislabels=(None, None), axislabels_flag=True,
-                            reverse_xaxis=None, colorbar_flag=True, 
-                            colorbar_label=None, colorbar_orient='vertical', 
+                            reverse_xaxis=None, colorbar_flag=True,
+                            colorbar_label=None, colorbar_orient='vertical',
                             edges=True, gatefilter=None,
                             filter_transitions=True, ax=None, fig=None):
         """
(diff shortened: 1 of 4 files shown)


pyart/graph/radardisplay.py
```diff
@@ -1041,6 +1180,35 @@ def _get_ray_data(self, field, ray, mask_tuple, filter_transitions):
 
         return data
 
+    def _get_azimuth_rhi_data_x_y_z(self, field, target_azimuth,
+                               edges,mask_tuple):
```
jjhelmus: Just a style issue:
This line is not properly aligned with "self" in the line above and there should be a space after the comma.

jjhelmus:
This looks really great @nguy, thanks for adding this.  A few comments, ideas:

* It looks like your editor is not stripping trailing whitespace.  [Removing trailing whitespace](http://programmers.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal) is a good practice and recommended by [PEP8](https://www.python.org/dev/peps/pep-0008/), Python's style guide.  A google search should pull up instructions on how to configure your editor to do this automatically.

* Any change you could add a unit test for the new method?  [test_xsect.py](https://github.com/ARM-DOE/pyart/blob/master/pyart/util/tests/test_xsect.py) and  [test_radar_display.py](https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/tests/test_radar_display.py) should give you some examples.  For graphing there is typically no artifacts to tests, the assumption is that if the code run and a plot was generated then everything is good.  Not exactly the best testing mentality but better than nothing.

* The `gatefilter` and `filter_transitions` parameters are not used in the method.  I think they can be passed to the `_get_azimuth_rhi_data_x_y_z` method and used to filter the data in the same manner as is used by the `_get_data` method.  

Again thanks for the code!

pyart/graph/radardisplay.py
```diff
@@ -1041,6 +1180,35 @@ def _get_ray_data(self, field, ray, mask_tuple, filter_transitions):
...
+        """Retrieve and return pseudo-RHI data from a plot function. """
+        # determine which rays from the ppi radar make up the pseudo RHI
+        data = self.fields[field]['data']
+
+        if mask_tuple is not None:
+            mask_field, mask_value = mask_tuple
+            mdata = self.fields[mask_field]['data'][ray]
```
jjhelmus: ray is not defined in this scope and I don't believe a slice is required here at all.

pyart/graph/radardisplay.py
nguy: Changed

pyart/graph/radardisplay.py
nguy: You are correct, good catch. The dangers of copy and paste!

nguy:
@jjhelmus Thanks for the tip, I wondered why I always seemed to have these issue!
I just adjusted my settings, so hopefully this will help. What editor do you use? It seems like you told me vim?

I had ```gatefilter``` and ```filter_transitions``` in there, not sure why I took those out.

I'll take a look at the test files you mention and see what I can do.

nguy:
Let me know if that test works for you. Other comments are added.

jjhelmus:
I use vim/Gvim/Macvim depending on the computer within reach and my mood.

I have the following in my .vimrc for removing whitespace. It does it automatically when .py files are saved and I can fit F5 to do manually.

```
" Functions
function! <SID>StripTrailingWhitespaces()
    " Preparation: save last search, and cursor position.
    let _s=@/
    let l = line(".")
    let c = col(".")
    " Do the business:
    %s/\s\+$//e
    " Clean up: restore previous search history, and cursor position
    let @/=_s
    call cursor(l, c)
endfunction

" call with <F5> and turn on for all Python files
nnoremap <silent> <F5> :call <SID>StripTrailingWhitespaces()<CR>
autocmd BufWritePre *.py :call <SID>StripTrailingWhitespaces()
```

I also put a bunch of other stuff in my .vimrc and .gvimrc files, all of them are in my [dotfiles repository](https://github.com/jjhelmus/dotfiles)

jjhelmus:
Looks good to me.  I'll merge this once the AppVeyor tests pass.  Thanks again @nguy.

nguy:
You bet, I figured I could get this in quickly and give myself a feeling of accomplishment this week. :)

scollis:
Nice work @nguy !!

-sent from a mobile device-

PR #409, opened 2015-10-24
ENH: replace_existing in Grid

kirknorth:
Adding the `replace_existing` kwarg to the `add_field` method of Grid. This is to be consistent with the Radar class and its add_field method. I know we plan on overhauling the Grid class but this method should be portable regardless of the changes to Grid in the future.

### DIFF
--- a/pyart/core/grid.py
+++ b/pyart/core/grid.py
@@ -65,25 +65,34 @@ class Grid(object):
         write_grid(filename, self, format=format,
                    arm_time_variables=arm_time_variables)
 
-    def add_field(self, field_name, field_dict):
-        """ Add field to Grid object.
+    def add_field(self, field_name, field_dict, replace_existing=False):
+        """
+        Add a field to the object.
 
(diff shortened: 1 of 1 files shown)


pyart/core/grid.py
```diff
@@ -65,25 +65,34 @@ def write(self, filename, format='NETCDF4', arm_time_variables=False):
...
             Dictionary containing field data and metadata.
+        replace_existing : bool, optional
+            True to replace the existing field with key field_name if it
+            exists, overwriting the existing data. If False, a ValueError is
+            raised if field_name already exists.
+
         """
 
+        # grid fields should have dimensions of (nz, ny, nx)
         nz, ny, nx = self.fields[self.fields.keys[0]]['data'].shape
```
kirknorth: There's a bug here, missing () after keys.

jjhelmus:
Thanks for the PR @kirknorth.  This looks good to me, merging!

PR #411, opened 2015-10-28
ENH: More informative error messages when optional dependency is missing

jjhelmus:
When a function or class is accessed which requires an optional dependency that is missing a **MissingOptionalDependecy** exception is raised with a helpful message.

### DIFF
--- a/doc/source/dev_reference/index.rst
+++ b/doc/source/dev_reference/index.rst
@@ -27,6 +27,7 @@ Documentation is broken down by directory and module.
     filters
     map
     util
+    pkg_util
     bridge
     testing
     _debug_info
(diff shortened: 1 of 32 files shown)


jjhelmus:
Example of the new messages which are provided when trying to use a function or class which depends on an optional dependency:

``` Python
>>> import pyart
>>> radar = pyart.testing.make_target_radar()
>>> display = pyart.graph.RadarMapDisplay(radar)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pyart/pkg_util/decorators.py", line 49, in wrapper
    _REQUIRES_MESSAGE % (func.__name__, requirement_name))
pyart.pkg_util.exceptions.MissingOptionalDepedency: The use of 'RadarMapDisplay' requires that the package basemap be installed.
```

Issue #414, opened 2015-10-28
Wrapped functions/classes cannot be well documented


jjhelmus:
PR #411 added the `requires` decorator which gives nice error message when an unmet optional dependency is encountered. It also strips the doc string off wrapped functions and class, for example; [read_odim_h5](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.aux_io.read_odim_h5.html#pyart.aux_io.read_odim_h5) and [RadarMapDisplay](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.graph.RadarMapDisplay.html#pyart.graph.RadarMapDisplay).  

The `__doc__` attribute can be copied unto the wrapped function using `functools.update_wrapper`, details in [Section 2.1 of SciPy-lecture](http://www.scipy-lectures.org/advanced/advanced_python/), but the function signature is still lost.  There are some [methods for preserving the function signature](http://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions) but from what I read none of these work reliably across Python 2 and 3 without adding an additional dependency.  

I'm thinking of reverting back the PR since it seems to cause more harm than good but wanted to see if other have another solution.  The other option is to make a module like [decorator](https://pypi.python.org/pypi/decorator) a required dependency for Py-ART.

jjhelmus:
After thinking a bit on this topic, I think removing the `requires` decorator makes sense.  In its place a check can be make for the optional dependency within the function or class.  If the check indicates that the package or extension is missing a **MissingOptionalDependency** will be raised with an appropriate message.  This requires a small amount of code repetition but allows the messages to be customized to the specific case and the documentations, call signature, etc can be access without complicated inspection of the function.

jjhelmus:
Will put in a PR to change this behavior shortly.

PR #416, opened 2015-10-29
MAINT: Replace requires decorator with if/raise block

jjhelmus:
* Move the exceptions module to the pyart namespace
* Use a if/raise block in classes and function that require optional dependencies.
* Remove raises decorator
* Remove the pkg_util subpackage
* Documentation of previously wrapped functions and classes should once against be available.

See discussion in #414


jjhelmus:
Merging before AppVeyor checks complete to trigger rebuilding of the docs.  Will monitor AppVeyor to verify that the builds are passing.

PR #426, opened 2015-11-07
Update join_radar function.

nguy:
Enabled join_radar in __init__.py

### DIFF
--- a/pyart/core/__init__.py
+++ b/pyart/core/__init__.py
@@ -27,7 +27,7 @@ Core functions
 
 """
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
Looks good enough to me.  There are a few areas where the logic could be simplified or clarified a bit but I'm planning refactoring most of this functionality soon (but no timeline right now) so I think this is good enough for the time being.

nguy:
Yeah, it at least get's a working model in. The logic may or may not have been done after a couple glasses of wine....

PR #433, opened 2015-11-24
ENH: Improvements to the Radar Class

### LINKED ISSUE #246: Cartesian locations for radar gates
A helper method should be added to the Radar object for returning the x, y, and z Cartesian locations from the radar location for a give or all gates in a radar volume (or sweep).

This functionality was suggested by Timothy Lang.

### LINKED ISSUE #247: Latitude, longitude and altitude of radar gates
A method should be added to the Radar class for determining the latitude, longitude and altitude of a particular gate or all gates in the radar volume.  

This functionality was suggested by Timothy Lang.

jjhelmus:
Adds gate_{x, y, z}, gate_edge_{x, y, z} and gate_{longitude, latitude, altitude} attributes to the Radar class which specify the Cartesian and geographic locations of the radar gates.  These attributes as well as the rays_per_sweep attribute are LazyLoadDict object which load the 'data' key into memory from a function on first access.  Class method for resetting these attribute to their initial lazy state are included to facilitate updates to the data from which these attributes are derived.  

Also included is an native implementation of the forward and inverse azimuth equidistant map projection, which is used by default to calculate the gate_longitude and gate_latitude attribute.  This allows these attributes to be accessible and contain reasonable data in the absence of a Basemap install.

To facilitate this work all of the functions which perform coordinate transformation in Py-ART were moved to a transforms module in the core sub-module.  

closes #247 
closes #246


nguy:
@jjhelmus it appears that ```antenna_to_cartesian``` a direct port of the old ```radar_coords_to_cart``` and ```antenna_vector_to_cartesian``` to ```sweep_coords_to_cart```?

Just want to double check. I've now run into two separate packages that use one or the other as helper functions.

tjlang:
Yeah, this change momentarily broke DualPol, SingleDop, and PyTDA. I basically just copied the old code into a helper function in each for those modules to insulate themselves from further Py-ART changes.

nguy:
Right on. I just found a similar dependency in brawl4d.

jjhelmus:
Yup, `radar_coords_to_cart` and `sweep_coords_to_cart` got moved to pyart.core and renamed.  I can add aliases to their older names and locations if needed.

nguy:
I have no strong opinion, BUT I like the new names better. I would hold off on aliases until you hear more grumbling. Though I'd say we may be approaching a time when aliases are going to be needed if packages keep sprouting up (good problem to have).

tjlang:
I would recommend aliases.

jjhelmus:
I could counter than neither of the functions were in the public API (aka included in the User Reference Manual) so other packages should not be using them... but I know that no-one (me included) follows that requirement.

I'll check into how best to add aliases for these.  I'd still like to depreciate the old names in the future but going through a release cycle with the Depreciation warnings and both names is probably best.

I do not think it will be possible to do thing for every function rename going forward, but it should be possible for the big one that folks bring up.

jjhelmus:
PR #439 adds aliases to the new transform functions so that the following function should be available as they previously were:

* pyart.io.common.radar_coords_to_cart
* pyart.graph.common.radar_coords_to_cart
* pyart.graph.sweep_coords_to_cart

Using these aliases will raise a warning informing users of their new location.  These can be suppressed using the simplefilter function in the warnings module if desired, but are turned on by default.  

My though is that these aliased will be retained for two released and removed afterwords.  For example anything aliased in 1.6.0+dev will be available in 1.6.0 and 1.7.0 and will be removed in 1.8.0+dev. 

Give me a ping if other function aliases are needed or if you have ideas on a better way to deal with function name and location changes.

jjhelmus:
Aliases for the `radar_coords_to_cart` and `sweep_coords_to_cart` functions in their former locations were added in PR #439.

PR #434, opened 2015-12-02
ENH: New Layout from Grid

### LINKED ISSUE #285: Replace the Grid class layout

jjhelmus:
Introduce a new improved layout for the **Grid** class.  The new class has the following attributes:

* fields
* metadata
* time
* origin_latitude, origin_longitude, origin_altitude : specifying the geographic location of the grid origin
* x, y, z : specifying the regular Cartesian/cartographic distances from the grid origin.
* nx, ny, nz : the sizes of the three above dimensions.
* projection : specified the map projection used to transform between Cartesian and geographic coordinates.
* radar_latitude, radar_longitude, radar_altitude : Geographic location of the radar which were used to create the grid.
* radar_name, radar_time : Additional information about the radars which were used to create the grid.
* nradar : number of radar used to create the grid.

Additional the following attribute are lazily loaded upon first access of the 'data' key:

* point_x, point_y, point_z : Cartesian/cartographic location of all grid points.
* point_latitude, point_longitude, point_altitude : Geographic location of all grid point using the map projection specified by the *projection* attribute

The legacy *axes* attribute is still available but will be depreciated in future Py-ART versions.  

**Grid** objects are creating using a new set of parameters. Support for creating **Grid** objects using the legacy parameters is available using the `Grid.from_legacy_parameters` class method, which will be depreciated in future Py-ART versions.

The variables inside of the NetCDF files which Py-ART creates to store Grid objects have changed to match these new parameter names.  The `write_grid` and `read_grid` function use this new format.  The `read_legacy_grid` function is provide to read in NetCDF file using the older grid format, this function will be depreciated in future Py-ART versions.  

To further aid in the transition to the new Grid file layout, a **convert_legacy_grid** script is included with Py-ART which will create new style NetCDF files from the legacy NetCDF grid files.  This script does not make use of any functions inside Py-ART and should be usable regardless of if or what version of Py-ART is installed.

All code in Py-ART and examples have been update to use the new Grid layout and file format.

Finally the as the functionally of the `add_2d_latlon_axis` is now available in the *point_latitude* and *point_longitude* this function is depreciated and will be removed in future version of Py-ART.

Implements and closes #285

### DIFF
--- a/pyart/core/grid.py
+++ b/pyart/core/grid.py
@@ -23,6 +23,12 @@ import warnings
 
 import numpy as np
 
+try:
+    from mpl_toolkits.basemap import pyproj
+    _PYPROJ_AVAILABLE = True
+except ImportError:
+    _PYPROJ_AVAILABLE = False
+
 from ..config import get_metadata
 from ..lazydict import LazyLoadDict
 from .transforms import cartesian_to_geographic
@@ -64,6 +70,8 @@ class Grid(object):
         interpreting this dictionary. If this key is present and set to True,
         which is required when proj='pyart_aeqd', then the radar longitude and
         latitude will be added to the dictionary as 'lon_0' and 'lat_0'.
+        Use the :py:func:`get_projparams` method to retrieve a copy of this
+        attribute dictionary with this special key evaluated.
     radar_longitude, radar_latitude, radar_altitude : dict or None, optional
         Geographic location of the radars which make up the grid.
     radar_time : dict or None, optional
@@ -72,6 +80,10 @@ class Grid(object):
         Names of the radars which make up the grid.
     nradar : int
         Number of radars whose data was used to make the grid.
+    projection_proj : Proj
+        pyproj.Proj instance for the projection specified by the projection
+        attribute.  If the 'pyart_aeqd' projection is specified accessing this
+        attribute will raise a ValueError.
     point_x, point_y, point_z : LazyLoadDict
         The Cartesian locations of all grid points from the origin in the
         three Cartesian coordinates.  The three dimensional data arrays
@@ -180,6 +192,32 @@ class Grid(object):
                    origin_latitude, origin_longitude, origin_altitude,
                    x, y, z)
         return grid
+
+    @property
+    def projection_proj(self):
+        """
+        Proj instance as specified by the projection attribute.
+
+        Raises a ValueError if the pyart_aeqd projection is specified.
+        """
+        projparams = self.get_projparams()
+        if projparams['proj'] == 'pyart_aeqd':
+            raise ValueError(
+                'Proj instance can not be made for the pyart_aeqd projection')
+        if not _PYPROJ_AVAILABLE:
+            raise MissingOptionalDependency(
+                "Basemap is required to create a Proj instance but it " +
+                "is not installed")
+        proj = pyproj.Proj(projparams)
+        return proj
+
+    def get_projparams(self):
+        """ Return a projparam dict from the projection attribute. """
+        projparams = self.projection.copy()
+        if projparams.pop('_include_lon_0_lat_0', False):
+            projparams['lon_0'] = self.origin_longitude['data'][0]
+            projparams['lat_0'] = self.origin_latitude['data'][0]
+        return projparams
 
     def _find_and_check_nradar(self):
         """
@@ -324,10 +362,7 @@ def _point_lon_lat_data_factory(grid, coordinate):
         """ The function which returns the geographic point locations. """
         x = grid.point_x['data']
         y = grid.point_y['data']
-        projparams = grid.projection.copy()
-        if projparams.pop('_include_lon_0_lat_0', False):
-            projparams['lon_0'] = grid.origin_longitude['data'][0]
-            projparams['lat_0'] = grid.origin_latitude['data'][0]
+        projparams = grid.get_projparams()
         geographic_coords = cartesian_to_geographic(x, y, projparams)
         # Set point_latitude['data'] when point_longitude['data'] is evaluated
         # and vice-versa.  This ensures that both attributes contain data from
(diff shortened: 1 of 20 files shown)


pyart/core/grid.py
```diff
@@ -81,18 +284,57 @@ def add_field(self, field_name, field_dict, replace_existing=False):
...
+        x = grid.point_x['data']
+        y = grid.point_y['data']
+        projparams = grid.projection.copy()
+        if projparams.pop('_include_lon_0_lat_0', False):
+            projparams['lon_0'] = grid.origin_longitude['data'][0]
+            projparams['lat_0'] = grid.origin_latitude['data'][0]
+        geographic_coords = cartesian_to_geographic(x, y, projparams)
+        # set the other geographic coordinate
+        if coordinate == 0:
+            grid.point_latitude['data'] = geographic_coords[1]
```
gamaanderson: `[1]`, is this correct? I don't know, just look strange. Down bellow it is  `[coordinate]`

pyart/graph/common.py
```diff
@@ -73,9 +73,9 @@ def parse_vmin_vmax(container, field, vmin, vmax):
 def parse_lon_lat(grid, lon, lat):
     """ Parse lat and lon parameters """
     if lat is None:
-        lat = grid.axes['lat']['data'][0]
+        lat = grid.origin_latitude['data'][0]
     if lon is None:
-        lon = grid.axes['lon']['data'][0]
+        lon = grid.origin_latitude['data'][0]
```
gamaanderson: shouldn't this be `origin_longitude`?

pyart/graph/gridmapdisplay.py
```diff
@@ -74,14 +74,14 @@ def __init__(self, grid, debug=False):
         self.debug = debug
 
         # set up the projection
-        lat0 = grid.axes['lat']['data'][0]
-        lon0 = grid.axes['lon']['data'][0]
+        lat0 = grid.origin_latitude['data'][0]
+        lon0 = grid.origin_longitude['data'][0]
         self.proj = pyproj.Proj(proj='aeqd', datum='NAD83',
```
gamaanderson: this should be the grid projection, now that it has one

pyart/graph/gridmapdisplay.py
```diff
@@ -74,14 +74,14 @@ def __init__(self, grid, debug=False):
...
                                 lat_0=lat0, lon_0=lon0)
 
         # determine grid latitudes and longitudes.
-        x_1d = grid.axes['x_disp']['data']
-        y_1d = grid.axes['y_disp']['data']
+        x_1d = grid.regular_x['data']
+        y_1d = grid.regular_y['data']
         x_2d, y_2d = np.meshgrid(x_1d, y_1d)
```
gamaanderson: should we let this as it is, or should we slice points_x and points_y, or even better slice points_latitude and points_longitude

pyart/core/grid.py
jjhelmus: Yes, [1] is correct in this context, but it is a bit confusing.  What is being done here is when the point_latitude or point_longitude 'data' key is evaluated 'data' for the other attribute is also calculated and stored.  This prevents the inverse map projection from having the be calculated twice.  Also, this prevents the point_longitude and point_latitude attributes from containing data from two differing projections.  Either both point_latitude['data'] and point_longitude['data'] are un-evaluated or both have been evaluated using the same map projection.

I'll expand the comment before this line to make this more clear

pyart/graph/common.py
jjhelmus: Yup, good catch.

pyart/graph/gridmapdisplay.py
jjhelmus: **GridMapDisplay** needs to be updated to use the new **Grid** attributes.  This is on my to-do list but I was going to implement these changes in a separate PR.  My plan is that the **GridMapDisplay** will make use of the **Grid** projection, but I think it can wait until this later PR.

pyart/graph/gridmapdisplay.py
jjhelmus: Again my plan was that this would be in another PR to fix up the **GridMapDisplay** class.  I was planning on using points_latitude and points_longitude here.

pyart/core/grid.py
```diff
@@ -10,39 +10,283 @@
...
+        # Raises a ValueError if the pyart_aeqd projection is specified.
+        projparams = self.get_projparams()
+        if projparams['proj'] == 'pyart_aeqd':
+            raise ValueError(
+                'Proj instance can not be made for the pyart_aeqd projection')
+        if not _PYPROJ_AVAILABLE:
+            raise MissingOptionalDependency(
+                "Basemap is required to create a Proj instance but it " +
+                "is not installed")
+        proj = pyproj.Proj(projparams)
```
gamaanderson: what happens if self.projection has invalid (user-introduced) attributes?

jjhelmus: Then pyproj.Proj will issue an exception which will not be caught.  Py-ART could catch this exception and offer its own error message, but I think seeing the error from pyproj would be more informative than a "the projection attribute is invalid" message.

I'm not seeing a good way to verify the projection attribute besides trying to use it to create a Proj instance.

gamaanderson: ok

PR #437, opened 2015-12-11
Added basemap keyword to RadarMapDisplay.plot_ppi_map

tjlang:
Sometimes you create a Basemap elsewhere and want to use it to plot radar data on. This enhancement allows you to plot a ppi on top of a separate Basemap, without creating a new Basemap every time you use plot_ppi_map. The example below leverages this new feature to simplify plotting a DEM, NPOL, and AMPR data all on the same Basemap instance, which is used in each of the 4 panels.

multi_panel_awesomeness


jjhelmus:
Great idea @tjlang, should also save some time when creating plots of the same basemap.  Looks good to me, merging.

Issue #364, opened 2015-08-28
Change Airborne RadarDisplay class


nguy:
The following change is suggested: ```pyart.graph.radardisplay_airborne.RadarDisplay_Airborne``` to ```pyart.graph.radardisplay_airborne.AirborneRadarDisplay```. In a side discussion in #335, @jjhelmus noted:
"Changing RadarDisplay_Airborne to AirborneRadarDisplay should probably go through a depreciation cycle (so the old name still works for at least one point release) but that can be accomplished by creating an alias in the __init__.py file. Other than that some imports and documentation would need to be change but I'm in favor of the name change."

As this class was originally created for the NOAA P-3 Tail Doppler Radar (TDR) it is somewhat specific to this type of radar system. Another possibility to is to rename the existing class **AirborneRadarDisplayTDR** and also include a **AirborneRadarDisplayLF** that would call a regular **RadarDisplay** for the lower fuselage radar, which behaves more like a surface-based system in terms of scanning and display properties.

jjhelmus:
I vote for renaming the Airborne class to **AirborneRadarDisplay**.  I would like to release version 1.5.0 prior to this change, but I can make this change after the release if you would like.

nguy:
That sounds good. I have to take care of some other things before thinking about further airborne radar additions. Question to chew on: What do you think about a plotting class for time series radar data such as upward-pointing lidar or cloud radar, etc?

josephhardinee:
A few quick things, those plots would be time height displays. Time series implies something else. Before we talk about the plotting, it is worth noting pyart doesn't actually have the capability to support reading these types of files yet. I believe Scott initiated a discussion about this in another thread, as it requires adding a new dimension(Doppler velocity) to the data model plus possibly addressing a few other issues.

scollis:
What we are dealing with here is a clash of nomenclatures.. In the rest 
of the world a time series is anything with datetime x-axis.. so 
technically a time height plot is a class of time series plots.. But 
yes.. "Time series data" has a special meaning in the radar world.


On 9/1/15 9:41 AM, Joseph Hardin wrote:
> A few quick things, those plots would be time height displays. Time 
> series implies something else. Before we talk about the plotting, it 
> is worth noting pyart doesn't actually have the capability to support 
> reading these types of files yet. I believe Scott initiated a 
> discussion about this in another thread, as it requires adding a new 
> dimension(Doppler velocity) to the data model plus possibly addressing 
> a few other issues.
>
[...]

PR #449, opened 2016-01-07
Airborne Display class name change

nguy:
Changed name from RadarDisplay_Airborne to AirborneRadarDisplay.
Added an alias in pyart.graph.__init__.py

As discussed in Issue #364 to change name, very minor and easy fix. There is further discussion in there regarding time series, but I think that maybe a separate issue should be opened on how to handle that.

### DIFF
--- a/pyart/graph/__init__.py
+++ b/pyart/graph/__init__.py
@@ -29,7 +29,8 @@ Plotting grid data
 
 from .radardisplay import RadarDisplay
 from . import cm
-from .radardisplay_airborne import RadarDisplay_Airborne
+from .radardisplay_airborne import AirborneRadarDisplay
+from .radardisplay_airborne import  AirborneRadarDisplay as RadarDisplay_Airborne
 from .gridmapdisplay import GridMapDisplay
 from .radarmapdisplay import RadarMapDisplay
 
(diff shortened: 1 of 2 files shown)


jjhelmus:
Thanks for putting this change in.

The module level docstring in `pyart/graph/__init__.py` and `pyart/graph/radardisplay_airborne.py` will need to be updated to reflect the new class name.  I think the old name can be removed from the docs. 

Should a  DepreciationWarning be issues when the old **RadarDisplay_Airborne** name is used?  That way this name can be removed completely in future released. 

I think sub-classing the new class in the `__init__.py` file is the best option.  The following should work:

``` Python
import warnings as _warnings


class RadarDisplay_Airborne(AirborneRadarDisplay):
    """ Depreciated name for the AirborneRadarDisplay class. """

    def __init__(self, *args, **kwargs):
        _warnings.warn(
            ("'RadarDisplay_Airborne' is depreciated and will be removed in"
             "future versions of Py-ART, please use 'AirborneRadarDisplay'"),
            DeprecationWarning)
        AirborneRadarDisplay.__init__(self, *args, **kwargs)
```

nguy:
I think I got the docstrings updated. Will you just double check?

jjhelmus:
Looks good to me, merging.

Issue #451, opened 2016-01-08
Time-height plotting


nguy:
Currently Py-ART does not support time-height plots it seems. The VPT (vertically-pointing) plot function plots the abscissa as distance ( I assume for a scanning type?). I would suggest modifying the ```RadarDisplay.plot_vpt``` to accept (or access) a time variable and format accordingly. Maybe use of a keyword?

This topic had some discussion in Issue #364 .

One potential issue would be to make sure that time was represented consistently.

nguy:
I found some time conversion utilities in ```pyart.util.datetime_utils.py```, but cannot find an instance of these being used in the package. 

I think adding a conversion to Epoch time here would be useful in the future when potentially combining radar instances.

jjhelmus:
@nguy Sounds like a great suggestion.  The current VPT plots do not properly represent gaps in time or even indicate the timing of the measurement.  I'll have a look at #453 later today and make some comments/suggestions.

jjhelmus:
Closed by PR #453

Issue #435, opened 2015-12-10
RadarCollection - Merged radar instances


nguy:
Following on a previous post, we've been talking about merging multiple radar instances. There was no clear way merge the metadata specifically as I recall. Just to bring this up again and to point out one type of solution that has been started. [Take a look](https://github.com/deeplycloudy/brawl4d/blob/master/radar/collection.py) at the way @deeplycloudy cloudy begins to handle this. 

Issues that remain are the metadata and the fact that this uses Pandas. I have spoken with @jjhelmus previously and we though that adding another dependency would not be good. But maybe exploring the [Pandas](https://github.com/pydata/pandas/blob/master/pandas/core/frame.py) or [xray](https://github.com/xray/xray/blob/master/xray/core/dataarray.py) could be a way to implement something in Py-ART?

This could also be very powerful for ARTView. I've been considering ways to cache radar instances for improved performance and potentially make animations.

I'd love to hear some thoughts.
@gamaanderson @scollis @kirknorth @tjlang

scollis:
Time of pandas, I know there is a lot of people talking of moving it into the scipy core 

-sent from a mobile device-

nguy:
By the way for the packages I've been working on, I convert the time to Epoch for a package standard. Whether this is best or not, *shrug*, but it's what I did in the ```pyart.core.radar.join_radar``` function.

jjhelmus:
From a  quick look at use of pandas in the **RadarCollection** class from brawl4d I believe the same results could be achieved using a NumPy [recarray](http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.recarray.html). I'm not convinced that pandas or xray are really needed in this case.

A good discussion that I would like to have before designing a **RadarCollect** (and likely a **GridCollection** class) are what the the use cases and needs of such an object.  With those requirement fleshed out it would be easier to discuss how best these can be implemented.

scollis:
Well one would be a quick way to go from a radar collection to a grid 
collection

Hell, some time we can work out how to use Jupyter cluster or 
multiprocessing to make it easier for users with clusters to map reduce 
lots of data..

also imagine being able to go from a radar collection to a QVP


Jonathan J. Helmus wrote:

gamaanderson:
Just stopping by to remember of issue #252.

@nguy creating a plugin to automatic rotate a list of cached radars (simple radar, no Vradar) is not difficult, it would be good to test it.

nguy:
[...]

PR #453, opened 2016-01-11
Time-height displays.

nguy:
Updated datetime_utils to convert to Epoch time if desired.
Updated RadarDisplay.plot_vpt to accept a flag to use time along x-axis.
Added time formatting for x-axis for time-height plots.

This is a suggestion at implementing the time-height plots discussed in Issue #451 and also allows to conversion of Epoch time for future development (perhaps Issue #435 ?).

### DIFF
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -1019,7 +1019,7 @@ def join_radar(radar1, radar2):
 
     # to combine times we need to reference them to a standard
     # for this we'll use epoch time
-    ### NG THIS should use new updated datetime_utils if accepted
+    ### TODO Use new updated datetime_utils if accepted
     estring = "seconds since 1970-01-01T00:00:00Z"
     r1dt = num2date(radar1.time['data'], radar1.time['units'])
     r2dt = num2date(radar2.time['data'], radar2.time['units'])
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -35,6 +35,7 @@ from netCDF4 import num2date, date2num
 from ..config import get_metadata
 from ..lazydict import LazyLoadDict
 from .transforms import antenna_vectors_to_cartesian, cartesian_to_geographic
+from ..util import datetime_utils
 
 
 class Radar(object):
@@ -1019,7 +1020,6 @@ def join_radar(radar1, radar2):
 
     # to combine times we need to reference them to a standard
     # for this we'll use epoch time
-    ### TODO Use new updated datetime_utils if accepted
     estring = "seconds since 1970-01-01T00:00:00Z"
     r1dt = num2date(radar1.time['data'], radar1.time['units'])
     r2dt = num2date(radar2.time['data'], radar2.time['units'])
@@ -1027,6 +1027,10 @@ def join_radar(radar1, radar2):
     r2num = date2num(r2dt, estring)
     new_radar.time['data'] = np.append(r1num, r2num)
     new_radar.time['units'] = estring
+    ### TODO Use new updated datetime_utils if accepted
+    # r1num = datetime_utils.datetimes_from_radar(radar1, epoch=True)
+    # r2num = datetime_utils.datetimes_from_radar(radar2, epoch=True)
+    # new_radar.time['units'] = datetime_utils.get_epoch_units()
 
     for var in new_radar.fields.keys():
         sh1 = radar1.fields[var]['data'].shape
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -1019,6 +1019,7 @@ def join_radar(radar1, radar2):
 
     # to combine times we need to reference them to a standard
     # for this we'll use epoch time
+    ### NG THIS should use new updated datetime_utils if accepted
     estring = "seconds since 1970-01-01T00:00:00Z"
     r1dt = num2date(radar1.time['data'], radar1.time['units'])
     r2dt = num2date(radar2.time['data'], radar2.time['units'])
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -35,7 +35,6 @@ from netCDF4 import num2date, date2num
 from ..config import get_metadata
 from ..lazydict import LazyLoadDict
 from .transforms import antenna_vectors_to_cartesian, cartesian_to_geographic
-from ..util import datetime_utils
 
 
 class Radar(object):
@@ -1028,9 +1027,6 @@ def join_radar(radar1, radar2):
     new_radar.time['data'] = np.append(r1num, r2num)
     new_radar.time['units'] = estring
     ### TODO Use new updated datetime_utils if accepted
-    # r1num = datetime_utils.datetimes_from_radar(radar1, epoch=True)
-    # r2num = datetime_utils.datetimes_from_radar(radar2, epoch=True)
-    # new_radar.time['units'] = datetime_utils.get_epoch_units()
 
     for var in new_radar.fields.keys():
         sh1 = radar1.fields[var]['data'].shape
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -493,7 +493,7 @@ class RadarDisplay(object):
                  colorbar_orient='vertical', edges=True,
                  filter_transitions=True,
                  time_axis_flag=False,
-                 dForm=None, tz=None, date_MinTicker=None,
+                 dForm=None, tz=None,
                  ax=None, fig=None):
         """
         Plot a VPT scan.
@@ -551,14 +551,14 @@ class RadarDisplay(object):
             No rays are filtered when the antenna_transition attribute of the
             underlying radar is not present.
         time_axis_flag : bool
-            True to plot the x-axis as time. False uses the default of index number.
-        dForm : str
-            Format of the time string for x-axis labels.
-        tz : str
+            True to plot the x-axis as time. False uses the index number.
+            Default is False - index-based.
+        dForm : str, optional
+            Format of the time string for x-axis labels. Parameter is
+            ignored if time_axis_flag is set to False.
+        tz : str, optional
             Time zone info to use when creating axis labels (see datetime).
-        date_MinTicker : str
-            Sting to set minor ticks of date axis,
-            'second','minute','hour','day' supported.
+            Parameter is ignored if time_axis_flag is set to False.
         ax : Axis
             Axis to plot on. None will use the current axis.
         fig : Figure
@@ -586,8 +586,7 @@ class RadarDisplay(object):
 
         # set up the time axis
         if time_axis_flag:
-            self._set_vpt_time_axis(ax, dForm=dForm, tz=tz,
-                               date_MinTicker=date_MinTicker)
+            self._set_vpt_time_axis(ax, dForm=dForm, tz=tz)
             x = self.times
 
         # mask the data where outside the limits
@@ -1050,8 +1049,7 @@ class RadarDisplay(object):
         else:
             ax.set_ylabel(y_label)
 
-    def _set_vpt_time_axis(self, ax, dForm=None, tz=None,
-                           date_MinTicker=None):
+    def _set_vpt_time_axis(self, ax, dForm=None, tz=None):
         """ Set the x axis as a time formatted axis.
 
         Parameters
@@ -1062,9 +1060,6 @@ class RadarDisplay(object):
             Format of the time string for x-axis labels.
         tz : str
             Time zone info to use when creating axis labels (see datetime).
-        date_MinTicker : str
-            Sting to set minor ticks of date axis,
-            'second','minute','hour','day' supported.
         """
         if dForm is None:
             dForm = '%H:%M'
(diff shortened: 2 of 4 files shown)


nguy:
Just FYI, I used a NEXRAD L2 file for test. Read in, converted using ```to_vpt``` and then set the ```time_axis_flag```. Worked for me.

pyart/core/radar.py
```diff
@@ -1019,6 +1019,7 @@ def join_radar(radar1, radar2):
 
     # to combine times we need to reference them to a standard
     # for this we'll use epoch time
+    ### NG THIS should use new updated datetime_utils if accepted
```
jjhelmus: Typically XXX to TODO are used to mark areas of code where future improvements might be needed.  This makes these areas easy to grep for.

pyart/graph/radardisplay.py
```diff
@@ -542,6 +550,15 @@ def plot_vpt(self, field, mask_tuple=None, vmin=None, vmax=None,
             sweeps from the plot.  False will include these rays in the plot.
             No rays are filtered when the antenna_transition attribute of the
             underlying radar is not present.
+        time_axis_flag : bool
+            True to plot the x-axis as time. False uses the default of index number.
+        dForm : str
```
jjhelmus: Add ",optional" to these and a message about how these parameters are ignored if time_axis_flag is set to False.

pyart/graph/radardisplay.py
```diff
@@ -1007,11 +1035,14 @@ def _label_axes_ray(self, axis_labels, field, ax):
         else:
             ax.set_ylabel(y_label)
 
-    def _label_axes_vpt(self, axis_labels, ax):
+    def _label_axes_vpt(self, axis_labels, time_axis_flag, ax):
```
jjhelmus: The 'test_radardisplay_user_specified_labels' test in test_radar_display.py assumes this method takes 3 parameters.  It should be updates with the new call signature (add False before the ax parameter).

pyart/graph/radardisplay.py
```diff
@@ -1019,6 +1050,44 @@ def _label_axes_vpt(self, axis_labels, ax):
...
 
+    def _set_vpt_time_axis(self, ax, dForm=None, tz=None,
+                           date_MinTicker=None):
+        """ Set the x axis as a time formatted axis.
+
+        Parameters
+        ----------
+        ax : Matplotlib axis instance
+            Axis to plot. None will use the current axis.
+        dForm : str
```
jjhelmus: Rename these parameters to match those mentioned above.

pyart/util/datetime_utils.py
```diff
@@ -1,30 +1,62 @@
...
     """ Return a datetime for the volume start in a Grid. """
-    return num2date(grid.time['data'][0], grid.time['units'])
+    if epoch:
+        dtrad = num2date(grid.time['data'][0], grid.time['units'])
+        epnum = date2num(dtrad, epoch_units())
+        return num2date(epnum, epoch_units())
+    else:
+        return num2date(grid.time['data'][0], grid.time['units'])
+
+def epoch_units():
```
jjhelmus: Why not make this a constant, EPOCH_UNITS instead of a function.

pyart/graph/radardisplay.py
```diff
@@ -1019,6 +1050,44 @@ def _label_axes_vpt(self, axis_labels, ax):
...
+        ax.xaxis.set_major_formatter(date_Fmt)
+        if date_MinTicker == 'second':
+            ax.xaxis.set_minor_locator(SecondLocator())
+        elif date_MinTicker == 'minute':
+            ax.xaxis.set_minor_locator(MinuteLocator())
+        elif date_MinTicker == 'hour':
+            ax.xaxis.set_minor_locator(HourLocator())
+        elif date_MinTicker == 'day':
+            ax.xaxis.set_minor_locator(DayLocator())
+
```
jjhelmus: I'm not seeing any minor ticks in my plot, might have something incorrect but it would be good to tests this.  I'm also not apposed to not adding minor ticks and leaving this customization to the users.

jjhelmus:
@nguy Thanks for this addition, having the ability to plot times in VPTs will be great.  I left a few comments on some places where I think the code could be improved.

pyart/core/radar.py
nguy: Changed.

pyart/graph/radardisplay.py
nguy: Added this text.

pyart/graph/radardisplay.py
nguy: Done

pyart/graph/radardisplay.py
nguy: Sorry, I don't follow this comment. Which names are you referring to here?

pyart/util/datetime_utils.py
nguy: Originally I had separate functions and therefore provided the user a way to get the units if desired. Since people have different ideas what the best way to handle time is, I didn't know if Epoch would be accepted as a standard. And maybe the keyword list could grow. Happy to change this to a constant at top of script if we feel it to be a standard.

pyart/graph/radardisplay.py
nguy: This might be too much customization to include. I'll remove it.

pyart/graph/radardisplay.py
jjhelmus: Oops, I must have forgotten to post the comment this was referring to after writing it.  

Py-ART uses lowercase_with_underscores for function arguments, so I think `time_form` and `date_minor_ticker` would be better names for `dForm` and `date_MinTicker`

pyart/util/datetime_utils.py
jjhelmus: I think the Unix epoch (Jan 1, 1970 00:00:00) is more or less an agreed upon standard so I'm in favor or including it as a constant here.

pyart/graph/radardisplay.py
nguy: Okay, renamed ```dForm```, kept ```tz``` the same (unless you want that more explicit). ```date_MinTicker``` was removed via comment below.

pyart/util/datetime_utils.py
nguy: This has been changed. I like using this time. I have heard arguments for calendar updates (e.g. Gregorian I think in 1570) when longer running datasets are used. Highly unlikely that this will matter for Py-ART.

pyart/core/radar.py
```diff
@@ -35,6 +35,7 @@
 from ..config import get_metadata
 from ..lazydict import LazyLoadDict
 from .transforms import antenna_vectors_to_cartesian, cartesian_to_geographic
+from ..util import datetime_utils
```
jjhelmus: This is going to cause a circular import problem if anything in the util sub-package imports from core, which the xsect module does.  For the time being I would remove this import and the commented out line but keep the TODO line

pyart/graph/radardisplay.py
```diff
@@ -112,6 +116,7 @@ def __init__(self, radar, shift=(0.0, 0.0)):
         units = radar.time['units']
         calendar = radar.time['calendar']
         self.time_begin = netCDF4.num2date(times, units, calendar)
+        self.times = datetimes_from_radar(radar)
```
jjhelmus: I think the calculation of times should be delayed until needed, not added as a RadarDisplay attribute.

pyart/graph/radardisplay.py
```diff
@@ -567,6 +584,11 @@ def plot_vpt(self, field, mask_tuple=None, vmin=None, vmax=None,
             x = np.arange(data.shape[1])
             y = self.ranges / 1000.
 
+        # set up the time axis
+        if time_axis_flag:
+            self._set_vpt_time_axis(ax, date_time_form=date_time_form, tz=tz)
+            x = self.times
```
jjhelmus: times can be calculated here instead of accessed from the attribute.

pyart/graph/radardisplay.py
```diff
@@ -1019,6 +1049,29 @@ def _label_axes_vpt(self, axis_labels, ax):
...
+        ax : Matplotlib axis instance
+            Axis to plot. None will use the current axis.
+        date_time_form : str
+            Format of the time string for x-axis labels.
+        tz : str
+            Time zone info to use when creating axis labels (see datetime).
+        """
+        if date_time_form is None:
+            date_time_form = '%H:%M'
+        if date_MinTicker is None:
```
jjhelmus: date_MinTicker is not defined in this scope, I believe this if statement can be removed.

pyart/graph/radardisplay.py
```diff
@@ -1019,6 +1049,29 @@ def _label_axes_vpt(self, axis_labels, ax):
...
+            date_MinTicker = 'minute'
+
+        # Set the date format
+        date_Fmt = DateFormatter(date_time_form, tz=tz)
```
jjhelmus: Is this used anywhere?

pyart/util/datetime_utils.py
```diff
@@ -1,30 +1,62 @@
...
+        return num2date(radar.time['data'][:], radar.time['units'])
 
 
-def datetime_from_dataset(dataset):
+def datetime_from_dataset(dataset, epoch=False):
     """ Return a datetime for the first time in a netCDF Dataset. """
-    return num2date(dataset.variables['time'][0],
+    if epoch:
+        dtdata = num2date(dataset.variables['time'][0],
+                         dataset.variables['time'].units)
```
jjhelmus: This line should have an extra space to align with the dataset in the line above to meet PEP8 style guidelines.

pyart/util/datetime_utils.py
```diff
@@ -1,30 +1,62 @@
 """ Functions for converting date and time between various forms """
 
-from netCDF4 import num2date
+from netCDF4 import num2date, date2num
 
+EPOCH_UNITS = "seconds since 1970-01-01T00:00:00Z"
 
-def datetime_from_radar(radar):
+def datetime_from_radar(radar, epoch=False):
```
jjhelmus: There needs to be two blank lines before this function to meet PEP8 conventions.

pyart/util/datetime_utils.py
```diff
@@ -1,30 +1,62 @@
...
     """ Return a datetime for the volume start in a Grid. """
-    return num2date(grid.time['data'][0], grid.time['units'])
+    if epoch:
+        dtrad = num2date(grid.time['data'][0], grid.time['units'])
+        epnum = date2num(dtrad, EPOCH_UNITS)
+        return num2date(epnum, EPOCH_UNITS)
+    else:
+        return num2date(grid.time['data'][0], grid.time['units'])
+
+def get_epoch_units():
```
jjhelmus: Not sure why this function is needed.

jjhelmus:
@nguy, A few additional comments.  I would keep any changes to core/radar.py independent of this PR as there are some special concerns that need to be considered when making changes to anything in the core sub-module.  Sorry to be a bugger on some of the style issues, I really like this new feature, but want to keep Py-ART's style consistent.

pyart/core/radar.py
nguy: Ooh, that's not good. Removed.

pyart/graph/radardisplay.py
nguy: Removed self.times and sent calculation into variable x here.

pyart/graph/radardisplay.py
nguy: Oops, missed that one when removing.

pyart/util/datetime_utils.py
nguy: Done. Also fixed the line below spacing.

pyart/util/datetime_utils.py
nguy: Didn't know that.

pyart/util/datetime_utils.py
nguy: This would potentially be needed in the future for ```join_radars``` to set the correct time units since only the data key is returned and not a dictionary.

pyart/util/datetime_utils.py
jjhelmus: The (pep8 tool)[https://pypi.python.org/pypi/pep8] is really helpful to find these.  I tend to run in on all Python files I work on before checking the files into git.

pyart/util/datetime_utils.py
jjhelmus: Could the time dictionaries 'units' key be set to datetime_utils.EPOCH_UNITS?

nguy:
@jjhelmus No worries. It's best practice to follow the same style. Sloppy on my part.

Removing those lines from radar.py should remove any changes to the core, excepting the not for potential future changes. Is there a reason that the ```join_radars``` is in ```pyart.core``` vs. ```pyart.util```?

jjhelmus:
I do not believe there is any good reason that `join_radars` is in `pyart.core` rather than `pyart.util` other than historical.  Moving `join_radars` to `pyart.util` makes sense to me and would resolve the circular import issue.

nguy:
Do you want me to do that as part of this PR or leave it for another?
I don't think it's widely used yet, so I doubt it would cause issues in moving. Would you alias or just move it straight out?

pyart/util/datetime_utils.py
nguy: I didn't know that would work. All you need is the string, so yeah I guess.

jjhelmus:
Do the move in another PR, makes it easier to find and revert if necessary.  Since join_radar is not in the user docs I don't see an issue with renaming it without an alias.  The alias can always be added in later if folks request it.

I think this PR is good to merge.  I'll wait util the CI checks finish, just to be sure.

nguy:
Sounds good. Do you want me to just do that as a mini-PR after we merge this?
Would another radar.py under ```pyart.util``` be confusing?

nguy:
Shoot, I forgot to remove the ```get_epoch_units``` function. Should I do that before the merge?

jjhelmus:
Go for it, I can hold off.

jjhelmus:
A mini-PR to move `join_radar` to `util` sounds good, maybe call the the module `radar_utils`?

nguy:
I just started the push. Should be done momentarily.
After this is merged, I'll do the mini-PR. That's a good name, analogous to datetime_utils and I could see ```radar_utils``` growing.

jjhelmus:
Sounds like a plan.  Just waiting to CI to pass, then I'll merge this.  Thanks for all you work on this!

nguy:
No problem. Luckily this was an easy one since it only took porting the code.

PR #455, opened 2016-01-18
ENH: Support for reading message 1 NEXRAD Level 2 files

### LINKED ISSUE #152: NEXRAD files with non-uniform gate locations
NEXRAD files can contain volumes where the gate spacing and gate locations are not constant between sweeps.  When these file are read, radar.range is set to the gate spacing for the sweep with the largest number of gates.  These results in some sweeps having an incorrect range.  An example file which displays this property is KMLB20101118_014806_V03 which is available from the NCDC.

File exhibiting this behavior should at the least detected and a warning/error raised. If possible find a method for returning all or some of the sweeps with the correct gate spacing (multiple radar objects?).

Thanks to Timoth Lang for reporting this bug.

### LINKED ISSUE #352: Read NEXRAD Level 2 files without Message 31
I am trying to work set of NEXRAD level 2 files where some contain Message 31, but a significant portion do not. In the current version of Py-ART, I see that only the message 31 files are readable.  Have there been any recent updates or developments that will make the message 1 files readable? If not, how far off is this feature?

jjhelmus:
Adds support for reading NEXRAD Level 2 files in message 1 format with the `read_nexrad_archive` function.  Low resolutions scans are interpolated to match the resolution of the highest resolution scan when the selected fields and scans do not contains the same gate resolutions.

closes #352 
closes #152


jjhelmus:
The Python 3.4, 64 bit AppVeyor build seems to get stuck resolving dependencies which may cause the CI to fail.  Merging this anyway since this is not related to this PR.

PR #456, opened 2016-01-19
Add radar_utils

nguy:
Created radar_utils.py in pyart.util.
Moved join_radar function to radar_utils.

PR #457, opened 2016-01-20
BUG: plot_vpt with time_axis_flag=True works.

jjhelmus:
Previously using RadarDisplay.plot_vpt with the time_axis_flag set to True
would raise an error as the radar variable was not in scope.  This should be
referring to the self._radar attribute.


jjhelmus:
This seems to have been missed in #453

jjhelmus:
Closing as this bug is being address in #456

PR #459, opened 2016-01-22
ENH: variational KDP retrievals

kirknorth:
Adding the ability to variationally retrieve specific differential phase (KDP) and propagation differential phase (PHIDP) from total differential phase (PSIDP) measurements. This enhancement is based largely off of the method outlined in [Maesaka et al. (2012)](http://www.meteo.fr/cic/meetings/2012/ERAD/extended_abs/QPE_233_ext_abs.pdf). Some things to note are:

1. This method is guaranteed to retrieve positive KDP and is therefore definitely valid for liquid clouds at S-, C-, and X-band radar frequencies. At cloud radar frequencies (e.g., 35-GHz) the assumption that propagation differential phase is monotonically increasing becomes suspect.

2. The total differential phase measurements of the input radar must be unfolded before using this new method.

3. This method is ready to support estimates of backscatter differential phase (DELTAHV) if they are available. While more research is required here, one way to estimate DELTAHV is to parameterize it using attenuation corrected differential reflectivity. However, at weather radar frequencies this term is usually negligible and is typically ignored.

4. I noticed that the linear programming submodule was located in the `pyart.correct` module. I believe the LP routine(s) also corrected (e.g., unfolded) differential phase measurements, which makes sense being in `pyart.correct`, however KDP is a retrieved quantity and therefore would make more sense in `pyart.retrieve`. This is why I have placed the new variational method in `pyart.retrieve`.

5. I still have to address one issue with the PHIDP boundary conditions. Outliers need to be handled better or else poor results are retrieved. This will be evident in the images I will post for this PR. This issue will be addressed in additional commits I will make.

6. Unfortunately this method requires some heavy duty computations (e.g., second-order derivatives) and since I don't know of any off-the-shelf solvers in Python that are fast, I had to wrap some Fortran routines. I am definitely open to Python alternatives here to make it easier for additions and/or fixes.

7. On my personal work machine running this routine in serial (it has not been parallelized in any way) on a full NEXRAD volume (~ 15 million radar gates) it took about 2 minutes to do 50 minimization iterations (`maxiter = 50`). When running it on the same radar but selecting only the lowest 5 elevations it took about 45 seconds.


kirknorth:
Results from a KVNX volume on 2011-05-20 near 1020 UTC which captured a squall line. Note the forward retrieved PHIDP suffered from poor boundary conditions close to the radar likely caused by ground clutter contamination. I will be addressing this in a future commit to this same PR.

nexradv06kvnx 20110520 101906 retrieval

kirknorth:
@scollis if I make this KVNX volume available can you perform KDP retrievals using the LP method to independently verify these results. I would do it but I don't have the necessary packages installed on my virtual machine for the LP routines and I remember them being a pain to install last time.

scollis:
Thanks for the contribution! This will be very valuable as it does not rely on any of the LP codes which are hard to install. 

I have been torn on correct versus retrieve I believe you are right and it is better put there. One of us will do a comparison to LP keeping in mind the technique we currently have implemented does not include Scott's recent improvements.

As a separate matter to Py-ART, since we now have LP, simple sobel, your method and thanks to @tjlang and @nguy the separate CSUtools repo we should use the may 20th case and distrometer data to do a bake off comparison.. I think such a study would be publishable.

tjlang:
Did someone say, "bake off"? This is from [CSU_RadarTools](https://github.com/CSU-Radarmet/CSU_RadarTools). FIR-based filtering of phase, plus adaptive window for calculating KDP based on reflectivity. It completes the entire volume in less than 2 seconds. The downside is the noisiness, and the possibility of negative KDP (hidden in this plot since I matched Kirk's color scale).

bakeoff

josephhardinee:
Does this work on windows(I had issues with FORTRAN on that platform)? As far as point 1 in Kirk's original post, it should be safe to include Ku band in the assumption of positive KDP for liquid drops as well. This assumption for sure breaks down at W, and on occasion at Ka band. It might be too much work, but it could be good if this has an option for incorporating the HID stuff Scott proposed a while back(Might be a bit too early for that too though) to ensure it stays constrained to liquid precipitation. 

I would like to see it tested with higher frequencies(say, Ku and below) as well. Also I have a large library of radar rays at multiple frequencies generated from scattering simulations of real DSDs from a paper I'm preparing on variational retrievals that I can pull a few out from so you can use them as test cases. It's not a substitute for testing on real data, but that can be the only way to truly characterize these algorithms quantitatively.

kirknorth:
@josephhardinee I have not tested it on a Windows machine but I suspect you'll have difficulties. This is why I mentioned point 6, I'm open to Python alternatives for the Fortran routines if they don't affect processing time.

@tjlang thanks for the quick work and providing the plot. Overall the two independent methods produce similar results. The "smoothness" of the variational results are controlled primarily by two free parameters: the low-pass filter constraint weight (`Clpf`) and the maximum number of iterations performed.

kirknorth:
Also to your HID point @josephhardinee, this routine is ready to use a `GateFilter` as input, and if I remember correctly the idea was to eventually include methods in this class to mask specific hydrometeor types if desired. @scollis correct me if I'm wrong.

scollis:
@kirknorth if the code has any kind of conditional application (ie rhohv > X, etc..) it would be far preferable if these were handled or had the option of being handled using a gatefilter instance.

I will be going back to the LP code and specific attenuation code and putting in the gatefilter option

tjlang:
@kirknorth Yeah, the FIR code has a variable window for filtering, and you can also adjust the number of filter passes. My results are for 1 pass with a 21-gate (5-km) filter window. So I could try to damp down the noisiness by being more aggressive with the filter.

kirknorth:
As I said, a `GateFilter` is already included as an option for this new method, so any kind of conditional sampling can be done with that and then used directly.

scollis:
Sorry I missed that.  Excellent work

-sent from a mobile device-

kirknorth:
The near range gate boundary condition (BC) problem is likely caused by ground clutter close to the radar. Here is a PDF of PHIDP_near BCs for the same KVNX radar volume. I basically just need to include a check on all PHIDP_near values to make sure they fall within a certain threshold of the main "real" distribution, say 2 standard deviations of that distribution. If a PHIDP_near value does not satisfy this criteria, then set it to the mean or median of the real distribution maybe.

nexrad_kvnx_near_bc

jjhelmus:
@kirknorth Awesome work, really excited about having another method of retrieving KDP in Py-ART!  I'll have to sit down and have an in-depth look at the  Maesaka et al. paper and this code before I can make more technical comments.  From a quick read through of the Fortran file it should be possible to rewrite the function in that module in Cython which should give similar performance but will make building  much easier, especially on Windows.

Next week my plan was to focus on getting the 1.6.0 release of Py-ART out and so I may not have the time to really look at this until the following week.

kirknorth:
Yes Cython would definitely improve the building process without affecting performance in any significant way. Since this was just a side project of mine I went with what I knew could get the job done quickly. With that said I really need to up my Cython game :stuck_out_tongue_closed_eyes:

pyart/retrieve/kdp_proc.py
```diff
@@ -0,0 +1,699 @@
...
+                    if slope > 0.0:
+                        phi_far[ray] = y[-1]
+                    else:
+                        phi_far[ray] = np.median(y)
+                    break
+
+                else:
+                    continue
+
+    # check for outliers in the near range gate boundary conditions
```
kirknorth: This is the section in `boundary_conditions_maesaka` that will address outliers in PHIDP_near. Additional pushes to this PR will address this (hopefully).

jjhelmus:
@kirknorth wanted to check with you on the status and your plan for this PR.  Were you planning on adding more to this, or would you like me to start taking a look and adding the the necessary code to have this build during install?

kirknorth:
Hi Jonathan,

Yes I still need to push the necessary boundary condition fixes upstream. I
am 100% in agreement that we should port the Fortran routines to Cython
making the building process easier. I'm not sure I can do this as
effectively as you right now so let me know if this is something you have
time for, if not I'll work on it in the coming weeks.

Kirk

jjhelmus:
Kirk,

I should be able to find some time in the coming weeks to port the Fortran to Cython.  I'll submit a PR against your branch when I have something working.

kirknorth:
Here is the same main figure as before, except the colour bar minimum now goes down to -1 deg/km. With the latest commit we are now able to get a retrieval for rays which have a significant ground clutter signal affecting the near range BC. Significant ground clutter returns along multiple gates which are further away from the radar can still affect the retrieval of propagation differential phase (not so much specific differential phase), but really it's outside the scope of this PR to check for this. 

test_kdp

kirknorth:
Testing the method on a SGP CSAPR volume from the same squall line event near 1120 UTC. Using the first 5 elevation scans (0.8-3.5 deg) and performing 100 iterations took ~ 40 sec on my machine. Note that this method is agnostic to the unambiguous range of differential phase (e.g., STAR versus alternate transmission modes; 360 deg versus 180 deg), just so long as the differential phase tendency is to increase with range. 

test_kdp_csapr

Should we remove (mask) gates in the retrieved fields which were not used, e.g., the excluded gates defined in a `GateFilter`?

Here are individual rays from the base elevation scan. The black dashed horizontal lines correspond to the near and far range BC values in which the retrieved PHI_DP solutions must vary between. The black dashed vertical line corresponds to the range gate of the far BC, i.e., not necessarily the maximum range of CSAPR. PSI_DP values where the significant detection was less than 1 have been masked. Note the spurious far range BC in the 270 deg azimuth panel and associated spike in KDP. This may be due to non-uniform beam filling near the edge of the cloud. Regardless of the cause, as of now there is no sanity check on the far range BC, just thought I'd point that out. 

test_kdp_csapr_ray

scollis:
Very nice. Phidp should be constant but Kdp I am torn on.. Either fill value or zero 

-sent from a mobile device-

jjhelmus:
Looking good @kirknorth.  I was able to re-write the Fortran subroutines in Cython. I'll submit a PR against your branch once I clean this up and verify that the performance is not significantly degraded.

scollis:
Great work @kirknorth and @jjhelmus !



Jonathan J. Helmus wrote:

jjhelmus:
@kirknorth I rewrote the brute force subroutines in Cython and submitted a PR against your branch (kirknorth/pyart#6).  The speed seems similar to the Fortran implementation.  Thanks again for your work on this, it is looking great.

jjhelmus:
Also don't worry about the failing Travis CI tests, they are due to #493

kirknorth:
The Fortran extensions have now been deprecated and replaced with Cython extensions thanks to @jjhelmus. This should help with making the build process much smoother for users wanting to use this variational method. With the Cython extensions, this method took ~ 50 sec on a CSAPR volume with functional size of 2,477,160 variables and `maxiter=100`. Results for a few sweeps are provided below.

csapr_kpd_example

jjhelmus:
@kirknorth, this is looking really good, do you think this is ready to be merged yet or is there additional work you want to do on the addition?  A few things I noticed that should be addressed at some time:

* the use of `xrange` on line 180 of kdp_proc.py needs to be replaced with range for Python 3 compatibility.  
* There are a few places where a space is needed after a comma to meet the PEP8 standard.
* If you wanted to add a unit test that would be awesome!  You can look at the test we do in the `test_phase_proc.py` file for inspiration.  `pyart.testing.make_single_ray_radar()` will make a radar with a single ray with a KDP field which can be processed.  In the phase processing tests we compare this with a known good result but you may want something different in this case.

kirknorth:
@jjhelmus please let me address these points over the next 2-4 days. I am especially interested in writing a unit test, as I need more experience with this. Just curious, do you know why Python 3 deprecated `xrange`?

kirknorth:
@jjhelmus let me know what you think of these unit tests and if I should change anything in them. To use them I was running `nosetests` from the retrieve directory.

kirknorth:
Okay, I see that the failed checks are due to the `print` statements not being called as functions for Python 3 compatibility. I actually was previously using the function call syntax, but when I ran `pylint` it told me that I was using unnecessary parentheses with my print statements so I removed them. I guess that was because I'm running Python 2.7.x

pyart/retrieve/tests/test_kdp_proc.py
```diff
@@ -0,0 +1,60 @@
...
+
+    """
+    radar = sample_objects.make_empty_ppi_radar(101, 1, 1)
+    psidp_dict = {
+        'data': np.atleast_2d(np.linspace(0.0, slope * 1000.0, radar.ngates))
+        }
+    radar.add_field(get_field_name('differential_phase'), psidp_dict)
+
+    return radar
+
```
jjhelmus: These blank lines at the end of the file can be removed.

jjhelmus:
This looks in great shape, nice job on the tests @kirknorth!  The only remaining change would be to add these new functions to the documentation.  Adding these to the documentation will make it easier to discover and then use this new feature!

Specifically:
* `kdp_maesaka` should be added to the function list in `pyart/retrieve/__init__.py`
* a module level function list should be added to `pyart/retrieve/kdp_proc.py`, see _kdp_proc.pyx for the syntax.  
* The `pyart.retrieve.kdp_proc` and `pyart.retrieve._kdp_proc` modules should be added to the developer reference by adding two lines to the `doc/source/dev_reference/retrieve.rst` file.

I'm happy to add these minor changes if you'd like, just give me a yell.

kirknorth:
@jjhelmus I'll push the PEP8 changes shortly. Just curious, so these `.. autosummary::` statements in the docstrings are added manually? I thought this was related to Sphinx somehow. Sorry if silly question, I don't have much experience with the documentation, but I'd like to learn more.

jjhelmus:
> Just curious, do you know why Python 3 deprecated xrange?

In Python 3 the `range` function is Python 2's `xrange` function and so the old name was dropped.  They are both generators which can be iterated over.  In Python 2, the `range` function returned a list, to get this behavior in Python 3 you need to explicitly call `list` on the range object, `list(range(10))`.  Why exactly this change was made is a bit beyond me.  I think it was intended to simplify and clarify the language.  Having two function that do nearly the same thing but in a slightly different manner is a bit confusing.

jjhelmus:
Yes the `autosummary:: ` blocks need to be added manually.  They are directives to [Sphinx's autosummary extension](http://www.sphinx-doc.org/en/stable/ext/autosummary.html) which produces tables of functions, methods and classes in the documentation which then link to the documentation stubs for the various functions, methods, and classes.  These stubs are produced automatically by [sphinx-autogen](http://www.sphinx-doc.org/en/stable/ext/autosummary.html#sphinx-autogen-generate-autodoc-stub-pages).  The other Sphinx extension Py-ART uses is [numpydoc](https://pypi.python.org/pypi/numpydoc) which converts the docstring in numpy format into a format compatible with Sphinx.

There are some methods to automate the creation of a complete set of API documentation, like [sphinx-apidoc](http://www.sphinx-doc.org/en/stable/invocation.html#invocation-of-sphinx-apidoc), but you need to be careful about how you design your API and I had problems getting it to work nicely with numpydoc.  In the end it seemed easier to just add the `autosummary::` blocks by hand.

You can really dig into [Sphinx](http://www.sphinx-doc.org/en/stable/index.html) if you want to work out all the details.  I often find looking at how other scientific Python projects do there documentation to be quicker.  I think for Py-ART setup, I borrowed heavily from the submodules in [SciPy](https://github.com/scipy/scipy).

jjhelmus:
This looks great, thanks for all the hard work @kirknorth.  Merging!

PR #460, opened 2016-01-26
BUG: Sigmet multi PRF flag outside of specified range

jjhelmus:
Sigmet files with a multi_prf_flag parameter outside of the known range are read assuming a Nyquist multiplier of 1.


jjhelmus:
This issue was reported on the Py-ART mailing list.  Waiting for the original reporter to provide a sample file which exhibits the issue to test the fix before merging.

jjhelmus:
Original poster did not respond with a file, merging this fix.

PR #463, opened 2016-01-27
MAINT: Refactor of Display classes

jjhelmus:
Refactor of **RadarDisplay**, **RadarMapDisplay**, **GridMapDisplay** and **AirborneRadarDisplay** to take into account the new layout and attribute of the **Grid** and **Radar** classes.  

Depreciate multiple unnecessary attributes of the **RadarDisplay** class:
    * x, y, z
    * starts, ends
    * time_begin
    * radar_name

Modification and additional unit tests for better coverage of the modules in the graph sub-package.

### DIFF
--- a/pyart/graph/gridmapdisplay.py
+++ b/pyart/graph/gridmapdisplay.py
@@ -47,9 +47,6 @@ class GridMapDisplay(object):
         Grid object.
     debug : bool
         True to print debugging messages, False to supressed them.
-    proj : Proj
-        Object for performing cartographic transformations specific to the
-        grid.
     basemap : Basemap
         Last plotted basemap, None when no basemap has been plotted.
     mappables : list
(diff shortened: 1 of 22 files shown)


jjhelmus:
@nguy @gamaanderson You may want to have a look at this PR to see if there is anything that will cause downstream problems with ARTView.  Many of the changes are purely cosmetic but I am hoping to depreciate some of the attributes of the **RadarDisplay** classes.   

Also @nguy I have no airborne radar data to test the changes to the **AirborneRadarDisplay** class with, could you have a look.  Getting some unit tests for this class in the near future would be nice.

nguy:
@jjhelmus thanks for the heads up! I'll pull this down and take a look at the airborne data side.

pyart/graph/gridmapdisplay.py
```diff
@@ -76,8 +76,8 @@ def __init__(self, grid, debug=False):
         # set up the projection
         lat0 = grid.origin_latitude['data'][0]
         lon0 = grid.origin_longitude['data'][0]
-        self.proj = pyproj.Proj(proj='aeqd', datum='NAD83',
-                                lat_0=lat0, lon_0=lon0)
+        self.proj = pyproj.Proj(
```
gamaanderson: You still need to update this to use the grid internal projection, right?

jjhelmus: Yes, thanks for getting another set of eyes on this.  I came up with a few other changes in addition that should go into this PR.

gamaanderson:
We will have to add version test for the new name of the airborne class.
Beside that I don't think we use this arguments. In grids and normal radars I had no problem. The airborne class however does not work since it try to set self.x (line 104 for instance).

nguy:
There is an alias that maps an old call to the new call now. I'm trying to think of a case when this would not suffice.

gamaanderson:
the name is deprecated, that means in the future the alias will be dropped. I prefer to migrate now to avoid surprises in the future.

jjhelmus:
> The airborne class however does not work since it try to set self.x (line 104 for instance).

This code only set the self.x, y and z attributes where were never access so I removed that block.

gamaanderson:
It worked.

jjhelmus:
I think this is PR is good to go, at least I've gotten to everything I can think of.  I'm going to look at it fresh tomorrow before merging.  If anyone spots anything before then let me know.

I am aware that I've misspelled Deprecate  as Depreciate (they are both real words so my spell checked didn't pick it up).  I'll fix those in another PR since it it covers more than just the Display classes.

nguy:
A quick test looks like it's working with Airborne data!

jjhelmus:
Not seeing any issue this morning, merging.

PR #464, opened 2016-01-27
ENH: Default field specific colormaps and luminosity when plotting

jjhelmus:
When plotting radar fields using the various **Display** classes if the colormap (cmap) and luminosity limits (vmin, vmax) are not specified then the defaults for these parameters will be determined by locating the field being plotted in a list of default colormaps and luminosity limits in the Py-ART configuration file.  

These parameters can still be explicitly specified.

### DIFF
--- a/pyart/default_config.py
+++ b/pyart/default_config.py
@@ -1313,32 +1313,45 @@ DEFAULT_FIELD_COLORMAP = {
     reflectivity: 'pyart_NWSRef',
     corrected_reflectivity: 'pyart_NWSRef',
     total_power: 'pyart_NWSRef',
-    velocity: 'pyart_NWSVel',
-    corrected_velocity: 'pyart_NWSVel',
-    spectrum_width: 'pyart_NWS_SPW',
-    differential_reflectivity: 'pyart_BrBu12',
-    corrected_differential_reflectivity: 'pyart_BrBu12',
-    cross_correlation_ratio: 'pyart_BrBu12',
-    normalized_coherent_power: 'pyart_Carbone17',
-    differential_phase: 'hsv',
-    unfolded_differential_phase: 'pyart_BlueBrown11',
-    corrected_differential_phase: 'pyart_BlueBrown11',
-    specific_differential_phase: 'pyart_BrBu12',
-    corrected_specific_differential_phase: 'pyart_BrBu12',
-    linear_depolarization_ratio: 'pyart_Carbone17',
-    linear_depolarization_ratio_h: 'pyart_Carbone17',
-    linear_depolarization_ratio_v: 'pyart_Carbone17',
     signal_to_noise_ratio: 'pyart_Carbone17',
+
+    velocity: 'pyart_BuDRd18',
+    corrected_velocity: 'pyart_BuDRd18',
+    eastward_wind_component: 'pyart_BuDRd18',
+    northward_wind_component: 'pyart_BuDRd18',
+    vertical_wind_component: 'pyart_BuDRd18',
+
+    spectrum_width: 'pyart_NWS_SPW',
+
+    normalized_coherent_power: 'pyart_Carbone17',
+
+    differential_reflectivity: 'pyart_RefDiff',
+    corrected_differential_reflectivity: 'pyart_RefDiff',
+
+    cross_correlation_ratio: 'pyart_RefDiff',
+
+    differential_phase: 'pyart_Wild25',
+    unfolded_differential_phase: 'pyart_Wild25',
+    corrected_differential_phase: 'pyart_Wild25',
+
+    specific_differential_phase: 'pyart_Theodore16',
+    corrected_specific_differential_phase: 'pyart_Theodore16',
+
+    linear_depolarization_ratio: 'pyart_SCook18',
+    linear_depolarization_ratio_h: 'pyart_SCook18',
+    linear_depolarization_ratio_v: 'pyart_SCook18',
+
     rain_rate: 'pyart_RRate11',
     radar_estimated_rain_rate: 'pyart_RRate11',
-    radar_echo_classification: 'pyart_EWilson17',
-    specific_attenuation: 'pyart_NWSVel',
+
+    radar_echo_classification: 'pyart_LangRainbow12',
+
+    specific_attenuation: 'pyart_Carbone17',
+
     differential_phase_texture: 'pyart_BlueBrown11',
-    eastward_wind_component: 'pyart_NWSVel',
-    northward_wind_component: 'pyart_NWSVel',
-    vertical_wind_component: 'pyart_NWSVel',
     height: 'pyart_SCook18',
     interpolated_profile: 'pyart_SCook18',
+
     # Additional reflectivity like fields
     'CZ': 'pyart_NWSRef',
     'DZ': 'pyart_NWSRef',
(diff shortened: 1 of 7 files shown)


jjhelmus:
This branch will be rebased one PR #463 is merged so only the commits after than PR are include.

jjhelmus:
This will make the [default field colormaps](https://github.com/ARM-DOE/pyart/blob/master/pyart/default_config.py#L1311-L1355) much more visible, does anyone have an opinion on if any of these should be changed?

jjhelmus:
Currently, when using Py-ART to plot radar data a single colormap (jet) and set of luminosity limits (-6 to 100) is used to generate the plot regardless of the field being visualized.  Specifying the cmap, vmin, and vmax arguments will override these defaults.  This is not an ideal situation.  The same colormap and luminosity limits should not be used for reflectivity data, Doppler velocities and other fields. 

We would like to change this behavior so that different colormaps and luminosity limits are provided for a number of known radar fields.  For example a "reflectivity" field will be plotted using the "pyart_NWSRef" colormap and limits of -30 to 75 where as a "velocity" field will use the "pyart_NWSVel" colormap with limits given by +/- the Nyquist velocity.  It will still be possible to override these defaults by specifying the cmap, vmin and vmax arguments and these defaults can be customized by [adjusting the Py-ART configuration file](http://arm-doe.github.io/pyart-docs-travis/user_reference/generated/pyart.load_config.html).

What we would like your help in is deciding on these default colormaps and luminosity limits for the various fields the Py-ART configuration file knows about.  Below is a table of proposed colormaps and limits, if you have an opinion on these please comment The [pyart_colormaps_and_limits repository](https://github.com/jjhelmus/pyart_colormaps_and_limits) has plots from a variety of radar for  many of these proposed colormaps and limits.

| Field                                 | Default colormap          | vmin        | vmax          |
| :------------------------------------ | :-----------------:       | :-------:   | :-------:     |
|                                       |                           |             |               |
| reflectivity                          |   pyart_NWSRef            |   -30       |   75          |
| corrected_reflectivity                |   pyart_NWSRef            |   -30       |   75          |
| total_power                           |   pyart_NWSRef            |   -30       |   75          |
| signal_to_noise_ratio                 |   pyart_Carbone17         |   -20       |   30          |
|                                       |                           |             |               |
| velocity                              |   pyart_NWSVel            |   -nyq      |   nyq         |
| corrected_velocity                    |   pyart_NWSVel            |   -nyq      |   nyq         |
| eastward_wind_component               |   pyart_NWSVel            |   -nyq      |   nyq         |
| northward_wind_component              |   pyart_NWSVel            |   -nyq      |   nyq         |
| vertical_wind_component               |   pyart_NWSVel            |   -nyq      |   nyq         |
|                                       |                           |             |               |
| spectrum_width                        |   pyart_NWS_SPW           |   0         |   nyq         |
|                                       |                           |             |               |
| normalized_coherent_power             |   pyart_Carbone17         |   0.0       |   1.0         |
|                                       |                           |             |               |
| differential_reflectivity             |   pyart_RefDiff           |   0         |   8           |
| corrected_differential_reflectivity   |   pyart_RefDiff           |   0         |   8           |
|                                       |                           |             |               |
| cross_correlation_ratio               |   pyart_RefDiff           |   0.5       |   1.05        |
|                                       |                           |             |               |
| differential_phase                    |   pyart_Wild25            |   -180      |   180         |
| unfolded_differential_phase           |   pyart_Wild25            |   -360      |   -360        |
| corrected_differential_phase          |   pyart_Wild25            |   -360      |   -360        |
|                                       |                           |             |               |
| specific_differential_phase           |   pyart_Theodore16        |   -2        |   5           |
| corrected_specific_differential_phase |   pyart_Theodore16        |   -2        |   5           |
|                                       |                           |             |               |
| linear_depolarization_ratio           |   pyart_SCook18           |   -40       |   0           |
| linear_depolarization_ratio_h         |   pyart_SCook18           |   -40       |   0           |
| linear_depolarization_ratio_v         |   pyart_SCook18           |   -40       |   0           |
|                                       |                           |             |               |
| rain_rate                             |   pyart_RRate11           |   0         |   150         |
| radar_estimated_rain_rate             |   pyart_RRate11           |   0         |   150         |
|                                       |                           |             |               |
| radar_echo_classification             |   pyart_LangRainbow12     |   0         |   11          |
|                                       |                           |             |               |
| specific_attenuation                  | pyart_Carbone17           | -10         | 65            |
|                                       |                           |             |               |
| differential_phase_texture            | pyart_BlueBrown11         | -180        | 180           |
|                                       |                           |             |               |
| height                                | pyart_SCook18             | 0           | 20000         |
| interpolated_profile                  | pyart_SCook18             | 0           | 10000         |

jjhelmus:
Colormaps available in Py-ART:

pyart_colormaps

kirknorth:
I believe the limits on specific attenuation are a bit much (maybe `vmax` around 10 dB/km), and I understand why we would include negative values to get a better feel for where zero is or spot bad values, but physically the radar signal would never gain energy within a unit distance. Also, a texture field, by definition will never have negative values, so `vmin` set to -180 deg for `differential_phase_texture` seems a bit much, once again though I understand including some negative values to get a better idea of where zero is.

josephhardinee:
My vote is to not use the NWS velocity colormap as a default. The green red is very colorblind unfriendly.  A diverging blue red(BuRd18 maybe) would be better or blue yellow.   
  
Differential reflectivity should go a little bit into the negative side. Maybe make vmin -1. It makes it easier to find differential attenuation, and negative Zdrs are valid for higher frequencies, and for ice in some conditions at lower frequencies.    

Although not in scope of this PR, ideally differential phase would be 0 to 360 or -180 to 180(or 0 to 180) depending on how differential phase is arranged in the file. A quick max/min check could achieve this. 

Rain rate could probably have a much lower maximum by default, maybe 50? 

Probably also outside the realm of this PR, but it would be nice to have some variables have a default norm applied(rhohv, NCP) that compress the low end of the range(0-.9 for rhohv for instance) and then expand out the upper side of the range.

jjhelmus:
@kirknorth Thanks for taking the time to look and this and for your input.  I agree with both your recommendations.  

I'll update the specific attenuation limits to 0 and 10, negative values with either be clipped or assigned to the set_over/set_under colors is the colormap.  If negative values are suspect then expecting the uses to manually adjust the limits is reasonable.

I'll also set the differential_phase_texture minimum to 0.

jjhelmus:
> My vote is to not use the NWS velocity colormap as a default. The green red is very colorblind unfriendly. A diverging blue red(BuRd18 maybe) would be better or blue yellow.

This one is hard, I agree that the colormap is bad it terms of colorblindness and the purple being part of the colormap instead of a `set_over`, `set_under` color is also annoying.  On the other hand it is the colormap used by the NWS for velocity.  I do like the look of the BuDRd18 colormap (BuOr12 is also nice).  Folds are easy to spot as is the 0 isodope, the downside is the 0 region is a bit large and the white color blends in with the default white background.  I'll go ahead and change the default to "pyart_BuDRd18".

> Differential reflectivity should go a little bit into the negative side. Maybe make vmin -1. It makes it easier to find differential attenuation, and negative Zdrs are valid for higher frequencies, and for ice in some conditions at lower frequencies.

Agreed, adjusted the limit to -1.

> Although not in scope of this PR, ideally differential phase would be 0 to 360 or -180 to 180(or 0 to 180) depending on how differential phase is arranged in the file. A quick max/min check could achieve this.

One method for accomplishing this would be to set the `valid_min` and `valid_max` dictionary entries in the field dictionary.  If these are present they override the default limits.

> Rain rate could probably have a much lower maximum by default, maybe 50?

This really depends on the time period, 150mm instantaneous or 150mm/hour is high but 150mm/day is not that outrageous.  We only have a single entry which covers all cases, which version is most often recorded in the file?

> Probably also outside the realm of this PR, but it would be nice to have some variables have a default norm applied(rhohv, NCP) that compress the low end of the range(0-.9 for rhohv for instance) and then expand out the upper side of the range.

PR #468 adds the `norm` parameter to the plotting methods which opens up the possibilities of non-uniform colormap spacing.  Making this scaling occur by default for some fields does make sense but will require a decent amount of work.

jjhelmus:
FYI, the table and plots in the [pyart_colormaps_and_limits repository](https://github.com/jjhelmus/pyart_colormaps_and_limits) have been updated.

jjhelmus:
FYI, the table and plots in the [pyart_colormaps_and_limits repository](https://github.com/jjhelmus/pyart_colormaps_and_limits) have been updated.

jjhelmus:
I want to include these defaults in the Py-ART 1.6.0 release which I want to get out tomorrow (Friday Feb 5th).  If anyone has any additional recommendations please submit before then.

jjhelmus:
Went ahead and changes the vmax for the rain rates to 50.

jjhelmus:
Rebased upon master and updated defaults to match values in the table in this thread.

swnesbitt:
Thanks for this.
Just noted that the NWS reflectivity colormap doesn't match up with the values used by the NWS.  e.g.,: http://radar.weather.gov/radar.php?rid=ilx

swnesbitt:
Here are hex values
"#646464" # ND
"#ccffff" # -30
"#cc99cc" # -25
"#996699" # -20
"#663366" # -15
"#cccc99" # -10
"#999966" # -5
"#646464" # 0
"#04e9e7" # 5
"#019ff4" # 10
"#0300f4" # 15
"#02fd02" # 20
"#01c501" # 25
"#008e00" # 30
"#fdf802" # 35
"#e5bc00" # 40
"#fd9500" # 45
"#fd0000" # 50
"#d40000" # 55
"#bc0000" # 60
"#f800fd" # 65
"#9854c6" # 70
"#fdfdfd" # 75

jjhelmus:
It should be possible to add this table to the `_cm.py` module.  Matching the NWS reflectivities to the colormap would likely require using passing a Matplotlib [BoundaryNorm](http://matplotlib.org/api/colors_api.html#matplotlib.colors.BoundaryNorm) instance as the **norm** argument.  A script showing how to do this would make a great addition to the Py-ART examples.

PR #472, opened 2016-02-03
BUG: Fix projection issue when creating large grids

### LINKED ISSUE #294: Projection issue for VERY large mosiac

### LINKED ISSUE #431: Question about routine radar_coords_to_cart in multi-radar gridding
Recently I reviewed the procedure of multi-radar gridding and find something might not be accurate in `grid_mapper`. If my understanding is correct, `map_to_grid` function in `grid_mapper` module iterate a list of radars, apply `radar_coords_to_cart` for each sweep to obtain a `(x, y, z)` for each gate. However, the `radar_coords_to_cart` function is a radar-station centric function. This means `azimuth==0` indicates the true north thus it is along the Y-axis. It is OK in any single radar mapping but in any multi-radar mapping, as long as two and more radar (especially they're far away)

Consider an unreal corner case like this. If two radar are located near to north pole: Radar A is (88.9N, 90E), Radar B is (88.9N, 90W). If mapping together, it would be a wrong to use `radar_coords_to_cart` individually and directly mosaick them using calculated (x, y): actually y_a and y_b should on same line but current routine in `grid_mapper` probably produces two parallel lines for y_a and y_b but put x_a and x_b on a same line. 

To correct this "rotating" issue, I got two ways now I could think out to solve this (if it is an issue). One way is plot multi-radar grids in a geographically projected map, not a lat-lon or unprojected planar map, then plus the geodesic convergence angle at the radar station to all azimuth recored; the second way is in `radar_coords_to_cart`, the finally result is converted back to lat-lon then all final results are plotted in lat-lon coordinates. The previous one needs the a calculation of geodesic convergence angle but I cannot figure out how to do this in proj.4 lib (unless we use a conic projection). The second way needs a lot of conversion from x-y to lat-lon along the great circle (might cause performance issue). Probably an another way would be completely change the way to calculate range gate using the equation given by Doviak and Zrnic in 1993. However, that could be more code to add.

I am not sure if it is considered in other procedure so the radar azimuth is actually corrected by geodesic convergence angle. If so, could you be kind to let me know the branch/file. I find this problem in my other research so I came back to find some references here but failed.

Thanks

jjhelmus:
Fixes the issue where radar locations were distorted when creating large grids with data from multiple radars which were far apart.  During the gridding/mapping process the geographic location of all radar gates is calculated using the map projection defined in the Radar instance.  These locations are then transformed into distances from the grid origin using a separate map projection.  This allows for the creation of distortion free grids which cover large areas.  The gridding/mapping process may be slightly slower due to the need for these additional transformation.

closes #431 
closes #294


jjhelmus:
Example of bug and results after fix.  The script below creates three fake radars located in Boston, Kansas City and Los Angeles and then creates a large grid over CONUS.  

Prior to this fix the radar locations were incorrect in the Grid.  Note that the cross hairs indicate the correct locations of the cities:

map_with_error_full

map_with_error_zoom

After this fix the locations are correct:

map_with_fix_full

map_with_fix_zoom


Script used:

``` Python

import pyart

la_latitude = 34.0500
la_longitude = -118.2500

boston_latitude = 42.3601
boston_longitude = -71.0589

kansas_city_latitude = 39.0997
kansas_city_longitude = -94.5783

# Fake radar at Los Angeles
radar1 = pyart.testing.make_target_radar()
radar1.range['data'] *= 100
radar1.latitude['data'][:] = la_latitude
radar1.longitude['data'][:] = la_longitude
gf1 = pyart.filters.GateFilter(radar1)
gf1.exclude_masked('reflectivity')


# Fake radar at Kansas City
radar2 = pyart.testing.make_target_radar()
radar2.range['data'] *= 100
radar2.latitude['data'][:] = kansas_city_latitude
radar2.longitude['data'][:] = kansas_city_longitude
gf2 = pyart.filters.GateFilter(radar2)
gf2.exclude_masked('reflectivity')

# Fake radar at Boston
radar3 = pyart.testing.make_target_radar()
radar3.range['data'] *= 100
radar3.latitude['data'][:] = boston_latitude
radar3.longitude['data'][:] = boston_longitude
gf3 = pyart.filters.GateFilter(radar2)
gf3.exclude_masked('reflectivity')

x_km = 2500
y_km = 1500
grid_limits = (
    (200, 200),
    (-y_km * 1000, y_km * 1000.),
    (-x_km * 1000, x_km * 1000.),
)

grid = pyart.map.grid_from_radars(
    (radar1, radar2, radar3),
    gatefilters = (gf1, gf2, gf3),
    gridding_algo = 'map_gates_to_grid',
    grid_shape=(1, 401, 401),
    grid_limits=grid_limits,
    grid_origin = (38.5, -95.0))

# backfill the grid with 1.0
fdata = grid.fields['reflectivity']['data']
fdata[fdata.mask] = 1.0
grid.fields['reflectivity']['data'] = fdata

display = pyart.graph.GridMapDisplay(grid)
display.plot_basemap(lat_lines=[], lon_lines=[])
display.plot_grid('reflectivity', 0, vmin=0, vmax=40, cmap='pyart_NWSRef')
display.basemap.drawcountries()

# plot crosshairs at the cities
display.plot_crosshairs(la_longitude, la_latitude,
                        line_style='r:')
display.plot_crosshairs(boston_longitude, boston_latitude,
                        line_style='b:')
display.plot_crosshairs(kansas_city_longitude, kansas_city_latitude,
                        line_style='k:')
plt.show()

plt.clf()
display = pyart.graph.GridMapDisplay(grid)
display.plot_basemap(
    auto_range=False,
    min_lat=30, max_lat=38, min_lon=-123, max_lon=-113,
    lat_lines=[], lon_lines=[])
display.plot_grid('reflectivity', 0, vmin=0, vmax=40, cmap='pyart_NWSRef')
display.basemap.drawcountries()
display.plot_crosshairs(la_longitude, la_latitude, line_style='r:')
plt.show()
```

striges:
Hi @jjhelmus, I'd just like to point out that unfortunately this bugfix didn't not correctly handle conformality. The aedp is not a conformal projection thus it is strange that the PPI image will be a perfect circle in a lat/lon map. I created a simple comparison using ArcMap and PyART.

pyart_plot


arcgis_reference

I simply recommend drawing multi-radar map in aedp rather than plain lat/lon, that would be the simplest fix.

striges:
I deleted my comments. Never mind, the plotted example map is web mercator auxiliary sphere. So everything is correct.

Issue #375, opened 2015-09-11
Documentation build using Travis CI broken


jjhelmus:
The documentation that is build when a Pull Request is merged into master is broken due to the use of Sphinx 1.3.1.

jjhelmus:
The cause of this issue is sphinx-doc/sphinx#1822.  This has been fixed in Sphinx but has not made it into a release.  Once Sphinx 1.3.2 or 1.4 is released the pinned version of Sphinx can be removed from the Travis setup.

jjhelmus:
Labelling this as Wontfix as this cannot be fixed for the time being.

jjhelmus:
Sphinx 1.3.5 is available as a conda package and does not suffer from the original bug that caused the issue with the docs.  Unfortunately the method of styling we are using for the pages is not supported so this will need to be fixed.

jjhelmus:
Removing `Wontfix` label, adding `Defect`.  Sphinx 1.3.5 can be used to build the docs but a new styling of the pages is needed.  Trying out some of the [builtin themes for Sphinx](http://www.sphinx-doc.org/en/stable/theming.html#builtin-themes) would be a good start.  I tried a few but the spacing on the examples page were a bit off and may need some tweaking.

PR #480, opened 2016-02-09
read Argentinean h5 files

swnesbitt:
This modification adds the ability to read ODIM-processed Argentinean h5 volume scan files from the SINARAME C-Band dual-pol network.  Raw BUFR files must be processed with bbufr with appropriate tables prior to use.  Contact Steve Nesbitt at snesbitt@illinois.edu if you need help.


pyart/aux_io/sinarame_h5.py
```diff
@@ -0,0 +1,319 @@
+"""
+pyart.aux_io.SINARAME_h5
+====================
```
jjhelmus: The docstring name should match the name of the file in case.  Also the underline needs to extend to the end of the line above.

pyart/aux_io/sinarame_h5.py
```diff
@@ -0,0 +1,319 @@
...
+    'SQI': 'normalized_coherent_power',
+    'SNR': 'signal_to_noise_ratio',
+    'VRAD': 'velocity',
+    'WRAD': 'spectrum_width',
+    'QIND': 'quality_index',
+}
+
+
+def read_sinarame_h5(filename, field_names=None, additional_metadata=None,
+                 file_field_names=False, exclude_fields=None, **kwargs):
```
jjhelmus: The indention on this line is off, it should line up with the line above.

pyart/aux_io/sinarame_h5.py
```diff
@@ -0,0 +1,319 @@
...
+        raise MissingOptionalDependency(
+            "h5py is required to use read_SINARAME_h5 but is not installed")
+
+    # test for non empty kwargs
+    _test_arguments(kwargs)
+
+    # create metadata retrieval object
+    if field_names is None:
+        field_names = SINARAME_H5_FIELD_NAMES
+    filemetadata = FileMetadata('SINARAME_h5', field_names, additional_metadata,
```
jjhelmus: This line a one character too long, it should be split into two lines

pyart/aux_io/sinarame_h5.py
```diff
@@ -0,0 +1,319 @@
...
+        start_epoch = t_data.min()
+        start_time = datetime.datetime.utcfromtimestamp(start_epoch)
+        _time['units'] = make_time_unit_str(start_time)
+        _time['data'] = (t_data - start_epoch).astype('float32')
+
+    # fields
+    fields = {}
+    h_field_keys = [k for k in hfile['dataset1'] if k.startswith('data')]
+    SINARAME_fields = [hfile['dataset1'][d]['what'].attrs['quantity'] for d in
+                   h_field_keys]
```
jjhelmus: indentation is incorrect on this line.

jjhelmus:
Thanks for putting this together @swnesbitt, it will make a nice addition to Py-ART.  

There are a few style issue in the file than need to be address that I pointed out with in code comments.  In addition the module should be added to the documentation by adding this function to to listing at the top of the `pyart/aux_io/__init__.py` file and by adding the module to the `doc/dev_reference/aux_io.rst` file. 

I'm not sure if I like the idea of including the `sinarame_to_cfradial.py` file in the scripts directory.  I think the script is important to make available but the files in the scripts directory get installed and added to the execution path of every Py-ART install.  This seem a bit much for a script with such a focused use case.  Perhaps including the script as an example or in the wiki would be a better solution?

Let me know if you want to revise this PR to address the above concerns. I'm happy to do this for you if you would prefer, just let me know.

swnesbitt:
OK, I should have run pep8, forgot to do that.  Thinking of trying sublime_text to help with that in the future.  Let me know if things look better.  Thanks for your help again!

jjhelmus:
Looks like you only pushed a commit to delete the script, did you miss pushing up another commit?

swnesbitt:
OK, think I'm back on the up and up here...

jjhelmus:
Sphinx is not liking the capitalized module name in the module level docstring and so the module doesn't get included in the documentation.  The relevant warnings are:

```
/home/jhelmus/dev/pyart/pyart/aux_io/sinarame_h5.py:docstring of pyart.aux_io.sinarame_h5:2: WARNING: Title underline too short.

pyart.aux_io.SINARAME_h5
====================
/home/jhelmus/dev/pyart/doc/source/dev_reference/aux_io.rst:9: WARNING: failed to import read_SINARAME_h5
/home/jhelmus/dev/pyart/doc/source/dev_reference/aux_io.rst:9: WARNING: toctree references unknown document u'dev_reference/generated/read_SINARAME_h5'
```

Changing line 2 and 3 of sinarame_h5.py to lowercase and to extending the underline should fix these.  Then this is good to go.

jjhelmus:
Sorry, you'll also need to change the function name on line 10 to be lowercase.  Crazy case specific programming languages, where is FoRTRAn when you need it :wink:.

swnesbitt:
OK, thanks.  Is there a way to check for sphinx compatibility on my machine?

jjhelmus:
@swnesbitt Looks goods, thanks for seeing this through.

You can install Sphinx on your machine and build to docs locally to check for errors.  We are still having issues with Sphinx 1.3.x (#375) so you need Sphinx 1.2.x.  With conda you can install this using `conda install sphinx==1.2.3`.  To build the docs go to the doc directory and run `make html`, on windows there is a make.bat file but I don't recall how, or if, this works.  The `rebuild_docs.sh` script in that directory can also be used to build the docs.

Building the docs also runs the examples, which requires that you have all the sample files in the correct locations.  You can skip running the examples by setting the CI environment variable to true, `export CI='true'` works in Linux.

PR #482, opened 2016-02-10
MAINT: Remove Python 2.6, 3.3 and add 3.5 to tags

jjhelmus:


jjhelmus:
Not sure why no CI was run for this...

PR #495, opened 2016-03-04
RHI contour example 2

Rumpkie:
This is the differential reflectivity example that Scott had discussed, I increased the opacity of reflectance map but not removed it because I wanted the example to cover how to add and adjust cmap alpha.

### DIFF
--- a/examples/plotting/plot_rhi_contours_differential_reflectivity.py
+++ b/examples/plotting/plot_rhi_contours_differential_reflectivity.py
@@ -14,7 +14,7 @@ print __doc__
 import matplotlib.pyplot as plt
 import pyart
 import numpy as np
-import scipy.ndimage as spyi
+import scipy.ndimage as ndimage
 
 filename = '220629.mdv'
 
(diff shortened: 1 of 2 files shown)


examples/plotting/plot_rhi_contours_differential_reflectivity.py
```diff
@@ -0,0 +1,88 @@
...
+
+"""
+print __doc__
+
+# Author: Cory Weber (cweber@anl.gov)
+# License: BSD 3 clause
+import matplotlib.pyplot as plt
+import pyart
+import numpy as np
+import scipy.ndimage as spyi
```
jjhelmus: The recommended name to import this is as ndimage

jjhelmus:
If you can rebase this, it should be good to go.

Rumpkie:
I rebased it and made the style change you recommended

jjhelmus:
@Rumpkie Looks good, the failing Travis-CI is unrelated.  Can you share a copy of the radar file (220629.mdv) or point out where it can be downloaded?

jjhelmus:
Found this file, merging.

PR #498, opened 2016-03-08
	modified:   pyart/graph/radarmapdisplay.py

swnesbitt:
Added `ax` keyword to call to `plot_colorbar()` in `radarmapdisplay.py` to allow proper locations of colorbars in multi-panel plots.
After fix, colorbars are on the correct panels, not on the last panel plotted:
image

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -301,7 +301,8 @@ class RadarDisplay(object):
             axislabels=(None, None), axislabels_flag=True,
             colorbar_flag=True, colorbar_label=None,
             colorbar_orient='vertical', edges=True, gatefilter=None,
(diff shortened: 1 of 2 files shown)


swnesbitt:
I also added optional keywords to the `graph` commands for plotting PPIs and RHIs to specify colorbar tick locations and labels.  These can be specified as arrays called `ticks` and `ticklabs`.  Note the non-uniform spacing on the correlation coefficient panel attached.
image

scollis:
Very nice! It would also be good to pull your examples into the gallery as well some time.. Once this is accepted @Rumpkie can work with you to make this so.

jjhelmus:
Looks great to me, thanks for the new feature @swnesbitt.

PR #508, opened 2016-03-18
Kalman Kdp filtering/smoothing

swnesbitt:
Working on implementing some unconstrained and constrained Kdp filters using Kalman Filters.

The first is an unconstrained empirical filter which creates the parameters of the filter bases on a priori information from the data itself.  Results are encouraging (NEXRAD supercell example shown).  Haven't included deltas in the filter yet, but this is a minor addition.  Avoids the negative Kdp that can happen with FIR filters.

I'll next take a look at the constrained filter by 
M. Schneebeli, J. Grazioli and A. Berne, "Improved Estimation of the Specific Differential Phase Shift Using a Compilation of Kalman Filter Ensembles," in IEEE Transactions on Geoscience and Remote Sensing, vol. 52, no. 8, pp. 5137-5149, Aug. 2014.

Thinking about ways to retrieve delta using these filters...

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -225,7 +225,7 @@ class RadarDisplay(object):
 
     def plot_ray(self, field, ray, format_str='k-', mask_tuple=None,
                  ray_min=None, ray_max=None, mask_outside=False, title=None,
-                 title_flag=True, axislabels=(None, None), gatefilter=None,
+                 title_flag=True, axislabels=(None, None),
                  axislabels_flag=True, ax=None, fig=None):
         """
         Plot a single ray.
@@ -260,9 +260,6 @@ class RadarDisplay(object):
             is False.
         title_flag : bool
             True to add a title to the plot, False does not add a title.
-        gatefilter : GateFilter
-            GateFilter instance. None will result in no gatefilter mask being
-            applied to data.
         axislabels : (str, str)
             2-tuple of x-axis, y-axis labels.  None for either label will use
             the default axis label.  Parameter is ignored if axislabels_flag is
@@ -279,7 +276,7 @@ class RadarDisplay(object):
         ax, fig = common.parse_ax_fig(ax, fig)
 
         # get the data and mask
-        data = self._get_ray_data(field, ray, mask_tuple, gatefilter)
+        data = self._get_ray_data(field, ray, mask_tuple)
 
         # mask the data where outside the limits
         _mask_outside(mask_outside, data, ray_min, ray_max)
@@ -1330,7 +1327,7 @@ class RadarDisplay(object):
 
         return data.T
 
-    def _get_ray_data(self, field, ray, mask_tuple, gatefilter):
+    def _get_ray_data(self, field, ray, mask_tuple):
         """ Retrieve and return ray data from a plot function. """
         data = self.fields[field]['data'][ray]
 
@@ -1338,11 +1335,6 @@ class RadarDisplay(object):
             mask_field, mask_value = mask_tuple
             mdata = self.fields[mask_field]['data'][ray]
             data = np.ma.masked_where(mdata < mask_value, data)
-
-        # mask data if gatefilter provided
-        if gatefilter is not None:
-            mask_filter = gatefilter.gate_excluded[ray]
-            data = np.ma.masked_array(data, mask_filter)
 
         return data
 


scollis:
Very nice @swnesbitt ! How fast does it run? Time to do a volume?

josephhardinee:
Your plot is showing negative KDP's though.

swnesbitt:
I should have said *helps* avoid. Bringi and Hubbert filter has the tendency to go large negative in these situations on isolated cells - will plot up later. This has no a priori, can condition the filter on Kdp-Z and Z-delta to constrain the filter guess.

jjhelmus:
@swnesbitt What are your plans for this.  Do you want to make this into a module for Py-ART or keep it as an example in a notebook?  We have a notebooks directory which this might be better suited for this although there is only a single example in there at the moment.

kirknorth:
@swnesbitt do you mind sharing which NEXRAD volume you were testing with? I would like to do a similar test using the variational KDP method I just submitted. Cheers.

swnesbitt:
Sure
[https://drive.google.com/file/d/0B6Z5EcBIjxY2UTdjWjdiUmtTT0E/view?usp=sharing](https://drive.google.com/file/d/0B6Z5EcBIjxY2UTdjWjdiUmtTT0E/view?usp=sharing
)

kirknorth:
Thanks @swnesbitt, here are a few plots which may help in future development and/or point out issues with the variational method. I can already see I need to add a minor fix to set rays with no differential phase data to the system phase offset instead of default 0 deg.

ktlx_kdp_ppi

Individual rays from 0.9 deg elevation surrounding the severe cell to the SW of the radar.

ktlx_kdp_ray

swnesbitt:
@kirknorth looks very promising - I'm still working on the Kalman filter but will be interested in comparing.  Have to tried any idealized tests, with added synthetic noise, to test your scheme?  I've been doing something like the following:
```python
def synthetic_kdp():
    from scipy.stats import norm
    x=np.arange(0,100000.,250.)
    xbar=50000.
    sigma=3000.
    maxval=4.
    sys_phase=60.
    sys_noise=3.
    kdp_true=(norm.pdf(x, xbar, sigma))
    kdp_true=(maxval/np.max(kdp_true))*kdp_true
    phidp_true=sys_phase+0.5*cumsum(kdp_true)
    phidp_noise=np.random.normal(0,sys_noise,len(x))+phidp_true
    return x,kdp_true,phidp_noise
```

kirknorth:
@swnesbitt no I have not, but this would be a valuable test to include in [test_kdp_proc](https://github.com/ARM-DOE/pyart/blob/master/pyart/retrieve/tests/test_kdp_proc.py)!

swnesbitt:
@kirknorth I just tried the Maesaka filter on the volume I shared with you with the default settings, no gate filter.  Can you please share any filter you applied to the data (to get the "significant echo"), because there seems to be some issue with noise near the radar? Any other "best practices" would be appreciated.  Thanks!
unknown-4

kirknorth:
@swnesbitt the variational routine certainly isn't immune to bad input data and without a `GateFilter` results could indeed be questionable. With that said, if the boundary conditions are not completely *off* then you should still retrieve a sensible KDP solution along the ray path with strong hydrometeor signal like you appear to show for some rays.

I wouldn't classify the PSIDP measurements near the radar and surrounding the supercell as noise, rather they are from insects and to lesser extent ground clutter, which can easily introduce gate-to-gate variations more than 50 deg. Here is a plot of results from `boundary_conditions_maesaka` without a `GateFilter` for the base sweep. Horizontal black dashed lines indicate the space in which the retrieved forward and reverse PHIDP solutions must vary between the near and far range boundaries (vertical dashed black lines<sup>**a**</sup>). You can immediately see some rays that are going to have issues, like 268 deg and 270 deg for starters. If a significant number of rays have poor boundary conditions this will manifest itself in large differences between retrieved forward and reverse PHIDP. If you set `verbose=True` in `kdp_maesaka` it will print out the mean bias and absolute errors between these two; if the MAE is greater than 1-2 deg you likely have BC issues. For example, running this volume with default settings and no `GateFilter` produced an MAE of ~ 8 deg. 

ktlx_bc

A quick `GateFilter` you can use which reduced the MAE ~ 2.5 deg:

```python
gf = GateFilter(radar, exclude_based=True)
gf.exclude_masked('reflectivity')
gf.exclude_below('reflectivity', 0.0, inclusive=False)
gf.exclude_below('cross_correlation_ratio', 0.8, inclusive=False)
```

<sup>**a**</sup> Now, with all that said, there are still some underlying issues that need to be addressed. The most pressing issue is making the range boundaries *hard* instead of *soft*; right now the hard range boundaries are `radar.range['data'][0]` and `radar.range['data'][-1]` as opposed to the ranges corresponding to the vertical lines in the above plot. This can lead to linear PHIDP profiles towards these boundaries and thus a constant non-zero KDP towards these hard boundaries, which in a PPI plot can make the retrieved KDP look *noisy* close to the radar.

swnesbitt:
@kirknorth thanks for your thoughts.  Was playing around with gatefilters yesterday similar to what you were proposing.  It did improve things in that cell close to the radar, but the Bringi Kdp filter as implemented by @tjlang was handling the cell better, primarily because it has a standard deviation of phase filter that removes those phase outliers.

unknown-5

Yes, you are right that there is clutter, and insects, etc. contributing to noisey phase values -- I should have been more specific.  Also, it appears that we are battling the clutter filters as well in meteorological echo.  I put together an illustration that shows the issues - you can see where the phase values don't follow the propagation norm occur in small targets like clutter, but also near the 0 isodop in precipitation, and sometimes in precipitation (could be other ground targets, although they occur at wide ranges of rho_hv, Z, and spectral width).  

movie

Seems like a standard deviation of phase filter (or other noise filter) could be useful in tossing out outliers that we shouldn't be concerned about.  The Bringi filter removes all gates in the filter where the standard deviation may be affected by an outlier, but it seems we could be smarter about that to preserve more information.  Kalman filters can handle this to some extent as long as you teach them to ignore those outliers.

kirknorth:
@swnesbitt yes I actually use PSIDP texture and RHOHV texture in my `GateFilter`, but I'm not sure Py-ART has routines that compute texture fields for arbitrary fields; I think it might for differential phase only. I'm using a toolkit that I'm in the process of migrating to GitHub called [Py-RECT](https://github.com/kirknorth/pyrect) for these utilities which works nicely with Py-ART. I'll post plots later today.

kirknorth:
Computing texture fields using a 3 x 3 window (ray x gate) and define `GateFilter` as

```python
gf = GateFilter(radar, exclude_based=True)
gf.exclude_masked('reflectivity')
gf.exclude_above('differential_phase_texture', 50.0, inclusive=False)
gf.exclude_above('cross_correlation_ratio_texture', 0.3, inclusive=False)
```

produces

output_grzkk4

ktlx_kdp

As I mentioned before, the *soft* boundary conditions must be addressed which can produce non-zero KDP back towards the radar if `R0 =/= radar.range['data'][0]`.

scollis:
Careful here.. PhiDP 2nd trips have smooth texture. You may want to also 
include texture of Radial Velocity

Kirk North wrote:

jjhelmus:
Is this PR still being actively worked on?  If not can it be closed?

scollis:
Since there is no actual method added to Py-ART closing this

PR #511, opened 2016-03-23
GeoTIFF output

tjlang:
Added GeoTIFF output capabilities to `pyart.io` module. New function is `pyart.io.write_grid_geotiff`, and it accepts Py-ART `Grid` objects. New functionality adds `gdal` as an optional dependency. If not installed, the function will complain and not work.

Example usage:
```
pyart.io.write_grid_geotiff(
    'test_radar_fp.tiff', grid, field='ZC', cmap='pyart_NWSRef',
    rgb=True, vmin=0, vmax=75, warp=True, sld=False)
```

This creates a simple 3-channel RGB GeoTIFF of composite reflectivity (ZC field), warped to a WGS84 lat/lon grid, using the `pyart_NWSRef` color table with min/max set at (0, 75). Here is what the output looks like in GIS software:
screen shot 2016-03-23 at 4 04 35 pm

One can also output single-channel floating point data to the GeoTIFF, and create an ancillary Style Layer Descriptor (SLD) file containing color table info.

Example usage:
```
import dualpol
hidcolor = dualpol.HidColors()
pyart.io.write_grid_geotiff(
    'test_radar.tiff', grid, field='FH', level=4, cmap=hidcolor.cmaphid,
    rgb=False, vmin=0, vmax=10, color_levels=10, warp=True, sld=True)
```

This creates a GeoTIFF with HID index values following the `dualpol` paradigm (https://github.com/nasa/DualPol). The SLD file that is created contains the default `dualpol` HID colormap, which is categorical and has 10 entries. The output is valid for vertical level 4 in the Py-ART `Grid` object (2.5 km MSL in this case).

What it looks like in GIS software:
screen shot 2016-03-23 at 4 36 50 pm


jjhelmus:
Very cool.  I'm about to leave for the day but will look at this in detail tomorrow.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+        print('GDAL not detected, GeoTIFF output failure!')
+        return
+
+    # Allow the user to swap filename/grid position in call
+    if type(filename) is not str:
+        fname = grid
+        gridvol = filename
+    else:
+        fname = filename
+        gridvol = grid
```
jjhelmus: I understand the idea behind this but I do not like it.  Either a user gives the parameter in the correct order or the function should fail.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+    specified, a 2D composite is created. User also can specify the field
+    to output.
+
+    This function requires GDAL Python libraries to be installed. These are
+    available via conda; e.g., 'conda install gdal'
+
+    Parameters
+    ----------
+    filename : str
+        Filename to save grid to
```
jjhelmus: This is not really the filename written to since the field name, etc is added to the name.  Perhaps specifying a template would be better.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+            if sld:
+                _create_sld(cmap, vmin, vmax, ofile, color_levels)
+
+            if warp:
+                # Warps TIFF to lat/lon WGS84 projection that is more useful
+                # for web mapping applications. Likely changes array shape.
+                os.system('gdalwarp -t_srs \'+proj=longlat +ellps=WGS84 ' +
+                          '+datum=WGS84 +no_defs\' ' + ofile + ' ' +
+                          ofile + '_tmp.tif')
+                os.system('mv ' + ofile + '_tmp.tif' + ' ' + ofile)
```
jjhelmus: This is not cross platform, `shutil.move` should work on more (all?) platforms.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+    if type(filename) is not str:
+        fname = grid
+        gridvol = filename
+    else:
+        fname = filename
+        gridvol = grid
+    name, end = fname.split('.')
+
+    f_flag = False
+    for f, field_dic in gridvol.fields.iteritems():
```
jjhelmus: iteritems does not work in Python 3, use items.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+import matplotlib.colors as colors
+import os
+try:
+    from osgeo import gdal
+    IMPORT_FLAG = True
+except ImportError:
+    IMPORT_FLAG = False
+
+
+def write_grid_geotiff(filename, grid, field=None, rgb=False, level=None,
```
jjhelmus: There is no documentation of the field parameter in the docstring.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+            rcells = ny * 0.5
+            dist = max(gridvol.axes['x_disp']['data'])
+            rangestep = gridvol.axes['x_disp']['data'][1] - \
+                gridvol.axes['x_disp']['data'][2]
+            # print("dist :" + str(dist))
+            # print("rcells :" + str(rcells))
+            # print("rangestep :" + str(rangestep))
+            lat = gridvol.axes['lat']['data'][0]
+            lon = gridvol.axes['lon']['data'][0]
+            filled = gridvol.fields[field]['data'].filled(fill_value=-32768)
```
jjhelmus: This does not work if the field data is not a masked array.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+    name, end = fname.split('.')
+
+    f_flag = False
+    for f, field_dic in gridvol.fields.iteritems():
+        if f == field:
+            f_flag = True
+            ofile = name + "_" + f + "." + end
+            nz, ny, nx = gridvol.fields[field]['data'].shape
+            rcells = ny * 0.5
+            dist = max(gridvol.axes['x_disp']['data'])
```
jjhelmus: axes is deprecated, use grid.x['data']

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+               to warp to a lat/lon WGS84 grid
+        False - No warping will be performed. Output will be Az. Equidistant.
+    sld : Boolean
+        True - Create a Style Layer Descriptor file (SLD) mapped to vmin/vmax
+               and cmap. File is named same as output TIFF, except for .sld
+               extension.
+        False - Don't do this
+    """
+    if not IMPORT_FLAG:
+        print('GDAL not detected, GeoTIFF output failure!')
```
jjhelmus: This should raise a MissingOptionalDependency exception like other functions in Py-ART which depend on an optional dependency.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+
+            if warp:
+                # Warps TIFF to lat/lon WGS84 projection that is more useful
+                # for web mapping applications. Likely changes array shape.
+                os.system('gdalwarp -t_srs \'+proj=longlat +ellps=WGS84 ' +
+                          '+datum=WGS84 +no_defs\' ' + ofile + ' ' +
+                          ofile + '_tmp.tif')
+                os.system('mv ' + ofile + '_tmp.tif' + ' ' + ofile)
+    if not f_flag:
+        print('Failed - field not found.')
```
jjhelmus: This should raise an exception.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,246 @@
...
+    Develop a Style Layer Descriptor file given a color table and
+    user-specified min/max files.
+    Arguments are the same as they are in write_grid_geotiff()
+    """
+    cmap = plt.cm.get_cmap(cmap)
+    if color_levels is None:
+        color_levels = 255
+    name, end = filename.split('.')
+    ofile = name + '.sld'
+    fileobj = open(ofile, 'w')
```
jjhelmus: This needs to be closed or the writing should be done in a context manager.

jjhelmus:
@tjlang This is a great start, and I'd like to get this included in Py-ART but it needs a bit or work.  I added a number of in-line comments on the code to begin with.  The docstrings will also need to be updated to match the [NumPy documentation standards](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) and the functions will need to be added to the module level docstrings.

pyart/io/output_to_geotiff.py
tjlang: Made `field` a required argument. Also reordered the requirement arguments to make them more logical in order.

pyart/io/output_to_geotiff.py
tjlang: Now refer to `filename` argument as a filename template. The function also can now handle it if the user did not add a suffix.

pyart/io/output_to_geotiff.py
tjlang: Added this capability.

pyart/io/output_to_geotiff.py
tjlang: Function now will fail if arguments are incorrect. The order of required arguments now is more logical, first `grid`, then `filename`, then `field`.

pyart/io/output_to_geotiff.py
tjlang: Removed the loop and now simply examine whether `field` appears in the dictionary's keys.

pyart/io/output_to_geotiff.py
tjlang: Fixed.

pyart/io/output_to_geotiff.py
tjlang: Added logic that handles the case of a standard unmasked `ndarray`.

pyart/io/output_to_geotiff.py
tjlang: Now use `shutil.move`.

pyart/io/output_to_geotiff.py
tjlang: A `KeyError` exception is now raised if the `field` doesn't exist in the `Grid` object.

tjlang:
I addressed the requested revisions and also updated docstrings. Please let me know if you see anything else that needs to be addressed before merging.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
+"""
+pyart.io.write_grid_geotiff
+====================
```
jjhelmus: This should extend to the end of the text above

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+        True - Use gdalwarp (called from command line using os.system)
+               to warp to a lat/lon WGS84 grid.
+        False - No warping will be performed. Output will be Az. Equidistant.
+    sld : bool, optional
+        True - Create a Style Layer Descriptor file (SLD) mapped to vmin/vmax
+               and cmap. File is named same as output TIFF, except for .sld
+               extension.
+        False - Don't do this.
+
+    Returns
```
jjhelmus: This section can be removed.  The convention is that the lack of a Return section indicates that the function does not return anything or more precisely returns None.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+    -------
+    None
+
+    """
+    if not IMPORT_FLAG:
+        raise MissingOptionalDependency(
+            'GDAL not detected, GeoTIFF output failure!')
+
+    if field in grid.fields.keys():
```
jjhelmus: This would be more readable if the failure is placed first.  Also less indentation is needed.

``` Python
    if field not in grid.fields.keys():
        raise KeyError('Failed -', field, 'field not found in Grid object.')

    # Determine whether filename template already contains a suffix
    ...
```

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+        # Determine whether filename template already contains a suffix
+        # If not, append an appropriate one.
+        if '.' in filename:
+            name, end = filename.split('.')
+        else:
+            name = filename
+            end = 'tif'
+        ofile = name + "_" + field + "." + end
```
jjhelmus: What is the reasoning behind building the filename from the user provided string and the field name?  Could we just use the filename that the user provides to the function instead of building it from a template?  
I'm not against using a template but none of the other functions that write files in Py-ART use this setup.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+        nz, ny, nx = grid.fields[field]['data'].shape
+        dist = max(grid.x['data'])
+        rangestep = grid.x['data'][1] - grid.x['data'][2]
+        lat = grid.origin_latitude['data'][0]
+        lon = grid.origin_longitude['data'][0]
+        # Check if masked array; if so, fill missing data
```
jjhelmus: I think `np.ma.filled(grid.fields[field]['data'], fill_value=-32768)` will accomplish the same thing.

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+        else:
+            filled = grid.fields[field]['data']
+        if level is None:
+            data = np.amax(filled, 0)
+        else:
+            data = filled[level]
+        data = data.astype(float)
+        try:
+            data[data == -32768] = np.nan
+        except:
```
jjhelmus: What exception is being caught here?  Is it safe to ignore it?

pyart/io/__init__.py
```diff
@@ -78,6 +78,7 @@
 from .uf import read_uf
 from .uf_write import write_uf
 from .grid_io import read_grid, write_grid, read_legacy_grid
+from .output_to_geotiff import write_grid_geotiff
```
jjhelmus: Add the `write_grid_geotiff` function in one of the documentation sections at the top of this file.  That way the function is included in the User Guide

jjhelmus:
Thanks for the updates, looking really good.  I made a few  few minor in code comments.  

Do you have any tests for this new functions that could be included?  The following will at least test that the function outputs to a file that contains something.  Let me know if you want to address this, I'm happy to write a few tests for the module if you would prefer.

``` Python
""" Unit Tests for Py-ART's output_to_geotiff.py module. """

import pyart


def test_write_grid_geotiff():
    grid = pyart.testing.make_target_grid()
    with pyart.testing.InTemporaryDirectory():
        pyart.io.write_grid_geotiff(grid, 'test.foo', 'reflectivity')
        # check that something was written to the file
        with open('test_reflectivity.foo', 'rb') as f:
            assert len(f.read()) > 0
```

pyart/io/output_to_geotiff.py
```diff
@@ -0,0 +1,318 @@
...
+        'radar.tif', and the field name is 'reflectivity', then
+        the final file will be called 'radar_reflectivity.tif'.
+    field : str
+        Field name to output to file. This also will be used to create the
+        final GeoTIFF filename.
+
+    Other Parameters
+    ----------------
+    rbg : bool, optional
+        True - Output 3-band RGB GeoTIFF
```
jjhelmus: Sphinx doesn't like something about these True/False blocks and is giving a warning.  I think a blank line is needed after each option, one after the True block, one after the False, but I'm not sure.  Easiest method I know to check if these work is to try building the docs.  I can do this before merging if you would prefer.

pyart/io/__init__.py
tjlang: Done

pyart/io/output_to_geotiff.py
tjlang: Done

pyart/io/output_to_geotiff.py
tjlang: Added blank spaces below each True and False text block

pyart/io/output_to_geotiff.py
tjlang: Removed

pyart/io/output_to_geotiff.py
tjlang: Done

pyart/io/output_to_geotiff.py
tjlang: Changed to not use the field in the filename. If there is not a suffix, however, '.tif' is appended so that `_create_sld` gets a suffix it can change to '.sld'

pyart/io/output_to_geotiff.py
tjlang: OK, trying that instead of the if block

pyart/io/output_to_geotiff.py
tjlang: I seem to recall some weird error during development, but I can't recall what it was. I took this exception handling block out and it seems to be working OK. If we start getting errors in the future we can add this block back in but be more specific about the exception being raised.

tjlang:
OK, further edits have been made. I did delete the appending of the field to the filename, so you'll need to fix your test function to account for that. I'll try to think about additional test functions, but if you have some ideas feel free to put them wherever they are supposed to go. Thanks!

Everything seems to be working. Here's a local multi-radar composite reflectivity mosaic from tonight, with all hell breaking loose around HSV.

jjhelmus:
Great.  Everything seems in order.  I'll create another PR adding some basic tests for the new functionality.  Thanks for seeing this through @tjlang.

PR #514, opened 2016-04-01
Add unit tests for the output_to_geotiff module

jjhelmus:


tjlang:
Thanks for doing this. If you can use GDAL to open the file, it will be very straightforward to inspect contents. But I am not sure how optional dependencies work in the Py-ART testing environment.

jjhelmus:
No problem.  I think for the time being the check that something is written to the file is enough.  Inspecting the content can be added if we find bugs in the function.  Inspecting and verifying the output from `gdalinfo --json file`  would probably be a good step if/when more complete tests are needed.

PR #522, opened 2016-04-20
BUG: ImportError on warn --> warnings.warn

kirknorth:
Simple ImportError where `warn` was used instead of `warnings.warn`. Since this is found in `read_legacy_grid` and this will be deprecated in future Py-ART releases it's not a huge deal but a bug nonetheless.


jjhelmus:
Good catch @kirknorth.  Thanks for the fix.

PR #527, opened 2016-05-04
TST: Test only non-locale dependent time string

### LINKED ISSUE #526: test_write_ppi_arm_time_vars unit tests depends on locale setting
The test_write_ppi_arm_time_vars unit test in `test_cfradial.py` depends on the currently locale setting producing the string `May` for the 5th month in 2011.  This condition is not meet for many locales (for example de_DE) and results in the unit test failing.  Only the non-locale depending portions of the base_time.string attribute should be tested.

jjhelmus:
closes #526


jjhelmus:
Superseded by #528, closing

Issue #534, opened 2016-05-18
BUG: pyart.aux_io.read_gamic()


kmuehlbauer:
As reported in [pyart-mailinglist](https://groups.google.com/d/msg/pyart-users/BtsRwN3CB-4/deu0mfLHDAAJ) there is some problem with the `read_gamic`-function. It seems, that the reader is trying to get some keys which are not within the files or have slightly other names (az_speed vs. azimuth_speed).

I tried with the provided file:

```python
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-32-365e531063c5> in <module>()
      3 filename = '2014-10-19--03_20_23_00.mvol'
      4 data, metadata = wrl.io.read_GAMIC_hdf5(filename)
----> 5 radar = pyart.aux_io.read_gamic(filename)

/home/kai/anaconda/envs/wradlib_27/lib/python2.7/site-packages/pyart/aux_io/gamic_hdf5.py in read_gamic(filename, field_names, additional_metadata, file_field_names, exclude_fields, valid_range_from_file, units_from_file, **kwargs)
    241 
    242     # instrument_parameters
--> 243     instrument_parameters = _get_instrument_params(gfile, filemetadata)
    244 
    245     gfile.close()

/home/kai/anaconda/envs/wradlib_27/lib/python2.7/site-packages/pyart/aux_io/gamic_hdf5.py in _get_instrument_params(gfile, filemetadata)
    279     dic = filemetadata('pulse_width')
    280     dic['data'] = gfile.sweep_expand(
--> 281         gfile.how_attrs('pulse_width_us', 'float32') * 1e-6)
    282     instrument_params['pulse_width'] = dic
    283 

/home/kai/anaconda/envs/wradlib_27/lib/python2.7/site-packages/pyart/aux_io/gamicfile.py in how_attrs(self, attr, dtype)
    107         """ Return an array of an attribute for each scan's how group. """
    108         return np.array([self._hfile[s]['how'].attrs[attr]
--> 109                          for s in self._scans], dtype=dtype)
    110 
    111     def how_ext_attrs(self, attr):

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/feedstock_root/build_artefacts/work/h5py-2.6.0/h5py/_objects.c:2687)()

h5py/_objects.pyx in h5py._objects.with_phil.wrapper (/feedstock_root/build_artefacts/work/h5py-2.6.0/h5py/_objects.c:2645)()

[...]

PR #535, opened 2016-05-19
FIX: several fixes for pyart.aux_io.read_gamic()

kmuehlbauer:
This fixes the `pulse_width_us` vs `pulse_width_mks` described in #534

### DIFF
--- a/pyart/aux_io/gamic_hdf5.py
+++ b/pyart/aux_io/gamic_hdf5.py
@@ -277,8 +277,13 @@ def _get_instrument_params(gfile, filemetadata):
     instrument_params['radar_beam_width_v'] = dic
 
     dic = filemetadata('pulse_width')
+    pw_names = ['pulse_width_us', 'pulse_width_mks', 'pulse_width']
+    pw_name = 'pulse_width_us'
+    for pw_name in pw_names:
+        if gfile.is_attr_in_group('how', pw_name):
+            break
     dic['data'] = gfile.sweep_expand(
-        gfile.how_attrs('pulse_width_us', 'float32') * 1e-6)
+        gfile.how_attrs(pw_name, 'float32') * 1e-6)
     instrument_params['pulse_width'] = dic
 
     dic = filemetadata('prt')
--- a/pyart/aux_io/gamic_hdf5.py
+++ b/pyart/aux_io/gamic_hdf5.py
@@ -233,9 +233,19 @@ def read_gamic(filename, field_names=None, additional_metadata=None,
     # scan_rate
     scan_rate = filemetadata('scan_rate')
     if scan_type == 'ppi':
-        scan_rate['data'] = gfile.ray_header('az_speed', 'float32')
+        azs_names = ['az_speed', 'azimuth_speed']
+        azs_name = azs_names[0]
+        for azs_name in azs_names:
+            if gfile.is_field_in_ray_header(azs_name):
+                break
+        scan_rate['data'] = gfile.ray_header(azs_name, 'float32')
     elif scan_type == 'rhi':
-        scan_rate['data'] = gfile.ray_header('el_speed', 'float32')
+        els_names = ['el_speed', 'elevation_speed']
+        els_name = els_names[0]
+        for els_name in els_names:
+            if gfile.is_field_in_ray_header(els_name):
+                break
+        scan_rate['data'] = gfile.ray_header(els_name, 'float32')
     else:
         scan_rate = None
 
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
keying into wrong group, fixed, sorry

kmuehlbauer:
- added the missing moments to the config file
- added fix for `az_speed` vs. `azimuth_speed` naming issue
- check for nyquist_velocity attribute in `extended`
- added hack for getting `pulse_width` enum correctly for BoXPol gamic files

The latter should be made versatile via config-file, but I've no clue how to to this. 

For testing purposes you can find here [PPI](https://github.com/wradlib/wradlib-data/blob/master/hdf5/2014-08-10--182000.ppi.mvol) and [RHI](https://github.com/wradlib/wradlib-data/blob/master/hdf5/2014-06-09--185000.rhi.mvol) scans.

pyart/aux_io/gamic_hdf5.py
```diff
@@ -277,8 +287,17 @@ def _get_instrument_params(gfile, filemetadata):
...
-    dic['data'] = gfile.sweep_expand(
-        gfile.how_attrs('pulse_width_us', 'float32') * 1e-6)
+    pw_names = ['pulse_width_us', 'pulse_width_mks', 'pulse_width']
+    pw_name = 'pulse_width_us'
+    for pw_name in pw_names:
+        if gfile.is_attr_in_group('/scan0/how', pw_name):
+            break
+    if pw_name == 'pulse_width':
+        dic['data'] = gfile.sweep_expand(
+        [0.2, 0.5, 1.0, 2.0][gfile.how_attrs(pw_name, 'int')] * 1e-6)
```
kmuehlbauer: These are the pulsewidth's (in us) for the BoxPol Radar.

pyart/aux_io/gamic_hdf5.py
```diff
@@ -300,8 +319,9 @@ def _get_instrument_params(gfile, filemetadata):
     instrument_params['unambiguous_range'] = dic
 
     dic = filemetadata('nyquist_velocity')
```
jjhelmus: I would move this inside the if block, there is no reason to create the dictionary if it is not used.

pyart/aux_io/gamic_hdf5.py
```diff
@@ -300,8 +319,9 @@ def _get_instrument_params(gfile, filemetadata):
...
-    dic['data'] = gfile.sweep_expand(gfile.how_ext_attrs('nyquist_velocity'))
-    instrument_params['nyquist_velocity'] = dic
+    if gfile.is_attr_in_group('/scan0/how/extended', 'nyquist_velocity'):
+        dic['data'] = gfile.sweep_expand(gfile.how_ext_attrs('nyquist_velocity'))
```
jjhelmus: This line is a bit toi long, splitting in into two lines would be good and could improve readability.

pyart/aux_io/gamic_hdf5.py
jjhelmus: With the latest version of NumPy this line is raising DeprecationWarning:

```
/home/jhelmus/dev/pyart/pyart/aux_io/gamic_hdf5.py:297: DeprecationWarning: converting an array with ndim > 0 to an index will result in an error in the future
  [0.2, 0.5, 1.0, 2.0][gfile.how_attrs(pw_name, 'int')] * 1e-6)
```
Selecting the first element of the extracted slice solves this:
```
[0.2, 0.5, 1.0, 2.0][gfile.how_attrs(pw_name, 'int')[0]] * 1e-6)
```

jjhelmus:
This looks good @kmuehlbauer, thanks for putting this together.  I made a few recommendation for improvement, then this should be good to merge.

I did notice that the gamic modules are broken in Python 3, although not a result of any changes you made.  I will work on getting these to work in Python 3 once this gets merged.

kmuehlbauer:
@jjhelmus I'll incorporate everything by tomorrow. Do you have any idea, how we could make the pulsewidth list a parameter or a config-file member. Normally I'm against hardcoding stuff. This might work then only for the two radars here, but there might be others too, which have different pulsewidth'. 

If you have some neat idea, I'll gladly fit this in.

jjhelmus:
> Do you have any idea, how we could make the pulsewidth list a parameter or a config-file member.

I'm not sure if this information should be encoding in the configuration file since it is radar specific.  Would adding it as an optional keyword argument to the `read_gamic` function work?  I have not work enough with GAMIC files to know how different these values tend to be be.

kmuehlbauer:
> Would adding it as an optional keyword argument to the read_gamic function work? I have not work enough with GAMIC files to know how different these values tend to be be.

I've really no idea, what radar pulsewidth gamic radar processors are dealing with, but we should assume that pw are not like ours. So I like the idea of having this as keyword argument of the read_gamic function with an additional error message if `pulse_width` is given while keyword is missing. You know what I mean. I'll try to get something working.

kmuehlbauer:
@jjhelmus I added an parameter to `read_gamic()` which defaults to None. Added also to the docstring. Works here fine for my use-case and should also not distrub any other gamic-flavours.

jjhelmus:
Thanks for the updates. Looks good to me and is able to open all the gamic files I have laying around.  I'll fix up Python 3 support in a separate PR shortly.

PR #536, opened 2016-05-20
MAINT: Python 3 fixes in gamicfile and gamic_hdf5

jjhelmus:


jjhelmus:
Follow up to #535 to fix up issues with using Python 3 with `read_gamic`

PR #538, opened 2016-05-25
Finding objects in scans and despeckling

tjlang:
This PR brings despeckling functionality to the `pyart.correct` module, via two functions - `find_objects` and `despeckle_field`. The latter leverages the former to identify small regions of contiguous pixels in any and every sweep in a `Radar` object. User can specify field thresholds, size of speckles, provide an initial `GateFilter`, etc. Then, an updated `GateFilter` is returned to help mask out the speckles. For volumes with 360-deg PPIs, the sweep edges are checked and objects are relabeled if they match, accounting for the periodic boundary.

Here is an example of the application of `pyart.correct.despeckle_field` to a NEXRAD sweep. In (a) no filter has been applied. I've already calculated standard deviation of differential phase (SD_PHIDP), so in (b) we apply a `GateFilter` that eliminates high SD_PHIDP as well as low Z. Finally, in (c) we add despeckling to this `GateFilter`, leading to a very clean scan. Despeckling an entire busy NEXRAD volume takes a few seconds on my 3-year-old Mac laptop.

screen shot 2016-05-25 at 10 33 53 am

As I mentioned, we find objects (via `scipy.ndimage.label`) to power the despeckling. But we also can just use `find_objects` to locate cells, for example. Smoothing is also available prior to object locating, to keep from over-identifying cells. Here's an example of finding 40-dBZ cells in the filtered Z field from above. Note how the output from `find_objects` is a dictionary that is ready to be imported as a new field to the `Radar` object. You can also use this function on a subset of sweeps, although that would be harder to import as a new field. Finding objects in a volume takes < 1 second on my setup. Note that objects are identified on a scan-by-scan basis, not per volume.

screen shot 2016-05-25 at 1 49 39 pm

Finally, the despeckling/object-finding also works for RHIs. Below is not the most dramatic example, but it shows how this works in practice. Same as before: (a) is no filter, (b) is gross filters on certain fields, (c) adds despeckling.

screen shot 2016-05-25 at 10 53 05 am

### DIFF
--- a/doc/source/dev_reference/correct.rst
+++ b/doc/source/dev_reference/correct.rst
@@ -7,7 +7,6 @@ Radar Moment correction routines.
 .. automodule:: pyart.correct.attenuation
 .. automodule:: pyart.correct.dealias
 .. automodule:: pyart.correct.despeckle
-.. automodule:: pyart.correct.find_objects
 .. automodule:: pyart.correct.phase_proc
 .. automodule:: pyart.correct.region_dealias
 .. automodule:: pyart.correct.unwrap
--- a/doc/source/dev_reference/correct.rst
+++ b/doc/source/dev_reference/correct.rst
@@ -6,6 +6,8 @@ Radar Moment correction routines.
 
 .. automodule:: pyart.correct.attenuation
 .. automodule:: pyart.correct.dealias
+.. automodule:: pyart.correct.despeckle
+.. automodule:: pyart.correct.find_objects
 .. automodule:: pyart.correct.phase_proc
 .. automodule:: pyart.correct.region_dealias
 .. automodule:: pyart.correct.unwrap
--- a/pyart/correct/__init__.py
+++ b/pyart/correct/__init__.py
@@ -25,6 +25,7 @@ Other corrections
 
     calculate_attenuation
     phase_proc_lp
+    despeckle
 
 Helper functions
 ================
@@ -33,6 +34,7 @@ Helper functions
     :toctree: generated/
 
     find_time_in_interp_sonde
+    find_objects
 
 """
 
@@ -43,5 +45,6 @@ from .phase_proc import phase_proc_lp
 from ..filters.gatefilter import GateFilter, moment_based_gate_filter
 from .unwrap import dealias_unwrap_phase
 from .region_dealias import dealias_region_based
+from .despeckle import find_objects, despeckle
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/correct/__init__.py
+++ b/pyart/correct/__init__.py
@@ -45,6 +45,6 @@ from .phase_proc import phase_proc_lp
 from ..filters.gatefilter import GateFilter, moment_based_gate_filter
 from .unwrap import dealias_unwrap_phase
 from .region_dealias import dealias_region_based
-from .despeckle import find_objects, despeckle
+from .despeckle import find_objects, despeckle_field
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/correct/despeckle.py
+++ b/pyart/correct/despeckle.py
@@ -22,7 +22,6 @@ Find contiguous objects in scans and despeckle away ones that are too small.
 """
 
 import numpy as np
-from ..exceptions import MissingOptionalDependency
 from ..filters.gatefilter import GateFilter
 from scipy.ndimage import label
 from scipy.signal import convolve2d
@@ -98,8 +97,8 @@ def find_objects(radar, field, threshold, sweeps=None, smooth=None,
     return _generate_dict(label_storage)
 
 
-def despeckle(radar, field, label_dict=None, threshold=-100,
-              size=10, gatefilter=None, delta=DELTA):
+def despeckle_field(radar, field, label_dict=None, threshold=-100,
+                    size=10, gatefilter=None, delta=DELTA):
     """
     Despeckle a radar volume by identifying small objects in each scan and
     masking them out. User can define which field to investigate, as well as
--- a/pyart/correct/despeckle.py
+++ b/pyart/correct/despeckle.py
@@ -1,0 +1,482 @@
+"""
+pyart.correct.despeckle
+=======================
+
+Find contiguous objects in scans and despeckle away ones that are too small.
+
+.. autosummary::
+    :toctree: generated/
+
+    despeckle
+    find_objects
+    _adjust_for_periodic_boundary
+    _append_labels
+    _check_for_360
+    _check_sweeps
+    _check_threshold
+    _generate_dict
+    _get_data
+    _get_labels
+    _smooth_data
+
+"""
+
+import numpy as np
+from ..exceptions import MissingOptionalDependency
+from ..filters.gatefilter import GateFilter
+try:
+    from scipy.ndimage import label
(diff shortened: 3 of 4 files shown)


hhuangmeso:
Finally. I always removes non-meteorological echos through the combination of rhohv, phidp, and sqi/spectrum width, and the using of the radial continuity. Thanks for the consideration of 2D properties.

kirknorth:
@tjlang awesome work, I've also used `ndimage` to find objects in PPI radar displays and despeckle using `ndimage.binary_erosion`, however this routine is (for some reason) missing the `mode` keyword found in other `ndimage` routines and therefore periodic boundaries are still not handled properly.

I know we're currently using [wradlib](https://github.com/wradlib/wradlib) to compute texture fields in Py-ART, and if I'm not mistaken only for differential phase, but I would also point out that `ndimage` can be used to elegantly compute texture fields of arbitrary scans (e.g., PPI, RHI, etc.) using `ndimage.uniform_filter`. See this [Stack Overflow thread](http://stackoverflow.com/questions/18419871/improving-code-efficiency-standard-deviation-on-sliding-windows) for more information. I have implemented something similar in [Py-RECT](https://github.com/kirknorth/pyrect/blob/master/pyrect/util/texture.py), but maybe this is something we want directly in Py-ART too.

scollis:
hey @kirknorth .. Have not used the texture in Wradlib for some time.. We are using circular statistics at the moment as seen in this notebook: http://nbviewer.jupyter.org/github/ARM-DOE/openvap/blob/master/development/cmac2/notebooks/full%20suite.ipynb

kirknorth:
@scollis I noticed that you used `ndimage.generic_filter`, which took a considerable amount of time to run (~ 220 sec). During my testing I ran into this issue as well, but `ndimage.uniform_filter` drastically reduces processing time, e.g., computing the texture field of 17 CSAPR sweeps takes ~ 0.3 sec.

scollis:
yes.. yes it does! very nice @kirknorth so to answer your question YES! It would be great to have in Py-ART.. Nice work on Py-RECT will be taking a look soon..

tjlang:
I agree that computing texture should not take much time. I've written out the 1D approach in full Cython here, and it takes under a fraction of a second for a full radar volume:
https://github.com/CSU-Radarmet/CSU_RadarTools/blob/master/csu_radartools/calc_kdp_ray_fir.pyx#L95

I probably could have used scipy to do the same thing in about the same time, but old fortran skills (and repurposed code) die hard ...

doc/source/dev_reference/correct.rst
```diff
@@ -6,6 +6,8 @@ Radar Moment correction routines.
 
 .. automodule:: pyart.correct.attenuation
 .. automodule:: pyart.correct.dealias
+.. automodule:: pyart.correct.despeckle
+.. automodule:: pyart.correct.find_objects
```
jjhelmus: There is no find_objects module only a function, this line is not needed.

pyart/correct/despeckle.py
```diff
@@ -0,0 +1,474 @@
...
+    _check_threshold
+    _generate_dict
+    _get_data
+    _get_labels
+    _smooth_data
+
+"""
+
+import numpy as np
+from ..exceptions import MissingOptionalDependency
```
jjhelmus: `MissingOptionDependency` is not used, this can be removed

pyart/correct/__init__.py
```diff
@@ -43,5 +45,6 @@
 from ..filters.gatefilter import GateFilter, moment_based_gate_filter
 from .unwrap import dealias_unwrap_phase
 from .region_dealias import dealias_region_based
+from .despeckle import find_objects, despeckle
```
jjhelmus: There are some issues if a module and function have the same name. Can the module be renamed (maybe despecking) or the function (despeckle_field)?

pyart/correct/despeckle.py
```diff
@@ -0,0 +1,474 @@
...
+
+    """
+    label_dict = {}
+    label_dict['data'] = label_storage
+    label_dict['units'] = 'None'
+    label_dict['long_name'] = 'Objects in Scan'
+    label_dict['standard_name'] = 'objects_in_scan'
+    label_dict['coordinates'] = 'elevation azimuth range'
+    label_dict['valid_max'] = np.max(label_storage)
+    label_dict['valid_min'] = 1
```
jjhelmus: A better location for this metadata would be the configuration file but I'm not against keeping it here for simplicity

jjhelmus:
@tjlang This is a really great PR, only a few minor quips. The biggest is that naming a module and function identically causes a big headache.  

My only other question before merging is if you wanted to add tests (it was in the to do list...).  I'm fine getting this into Py-ART without tests so that users can start exploring the new feature as it is very much in demand. Unit tests can be always be added later.

scollis:
Once it is accepted I will volunteer to help with adding some tests..

I need the practice. I am very rusty at the moment.


Dr Scott Collis
Argonne National Laboratory
Mb: +1 630 235 8025
Of: +1 630 252 0550

tjlang:
Updated with most requested fixes. Biggest is function now called `despeckle_field`. I am skipping tests for now since Scott graciously volunteered. I am also leaving the metadata definition in the local module, for simplicity.

jjhelmus:
This looks great @tjlang.  The function name needs to be changed in the docstring but I'll fix that in a seperate PR.

tamisalex:
I don't know if this is the correct place to ask, but I can't figure out what to do with the gatefilter that pyart.correct.despeckle_field(radar,"reflectivity",size=20) returns.

I have tried passing it as an argument to display.plot (..., gatefilter=gf), but the plots that are being generated appear to not be affected at all gatefilter.

[Despeckle+Test.pdf](https://github.com/ARM-DOE/pyart/files/414562/Despeckle.Test.pdf)

tjlang:
Couple things - 1) It doesn't look like there is much to despeckle in the first place. 2) It doesn't look like you are passing the `GateFilter` object to your custom plot function.

tamisalex:
Maybe I am misunderstanding what despeckling is for. Is that circular cloud around the radar not bioscatter? Is this method for cleaning up bioscatter?

In the pdf I uploaded, I mistakenly added the gate filter to both.

```
display.plot(plot[0], plot[2], ax=ax, title=plot[1],
                 colorbar_label='', gatefilter=gf,
                 axislabels=('East-West distance from radar (km)' if plotno == 6 else '', 
                             'North-South distance from radar (km)' if plotno == 1 else ''))
```

plot[0] is the field ( reflectivity ), plot[2] is the sweep, gatefilter=gf to add the despeckle mask to the radar
[Despeckle+Test+2.pdf](https://github.com/ARM-DOE/pyart/files/415408/Despeckle.Test.2.pdf)

tjlang:
Despeckling removes small contiguous "speckles" of echo, not specifically biological echoes. You would run despeckling after any and all filtering, to clean up any remaining speckles of echo that made it thru the filters.

For polarimetric data the module `csu_radartools` has a simple, adjustable insect/bird filter that is based on low-Z/high-ZDR, and is compatible with `pyart`.

https://github.com/CSU-Radarmet/CSU_RadarTools/blob/master/csu_radartools/csu_misc.py#L43

tamisalex:
Cool, thanks for the help!

PR #541, opened 2016-06-07
ENH: NEXRAD VCP included in radar metadata

jjhelmus:
The NEXRAD volume coverage pattern (VCP) is included in the radar.metadata
dictionary when NEXRAD Level 2 files which contain this information are read.


jjhelmus:
Thanks to Youngsun Jung for suggesting this improvement.

PR #554, opened 2016-07-05
CI: Pin conda version when installing from recipe

jjhelmus:


jjhelmus:
Workaround for conda/conda#2974

PR #555, opened 2016-07-21
Added rainbow_wrl.py. Function to read rainbow5 files

jfigui:
Dear py-art users,

I would like to add a function to read rainbow files for Py-ART using the wradlib library.

It is contained in file rainbow_wrl.py in module aux_io.

Let me know if it is up to standard or I should change something.

Greetings from Switzerland,

Jordi

### DIFF
--- a/pyart/aux_io/__init__.py
+++ b/pyart/aux_io/__init__.py
@@ -42,5 +42,6 @@ from .edge_netcdf import read_edge_netcdf
 from .odim_h5 import read_odim_h5
 from .gamic_hdf5 import read_gamic
 from .sinarame_h5 import read_sinarame_h5
+from .rainbow_wrl import read_rainbow_wrl
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/aux_io/rainbow_wrl.py
+++ b/pyart/aux_io/rainbow_wrl.py
@@ -1,0 +1,493 @@
+"""
+pyart.aux_io.rainbow
+====================
+
+Routines for reading RAINBOW files (Used by SELEX) using the wradlib library
+
+This routine has been tested to read rainbow5 files version 5.22.3, 5.34.16 and 5.35.1.
+Since the rainbow file format is evolving constanly there is no guaranty that it can work with other versions.
+If necessary, the user should adapt to code according to its own version.
+
+Data types read by this routine:
+Reflectivity: dBZ, dBuZ, dBZv, dBuZv
+Velocity: V, Vu, Vv, Vvu
+Spectrum width: W, Wu, Wv, Wvu
+Differential reflectivity: ZDR, ZDRu
+Co-polar correlation coefficient: RhoHV, RhoHVu
+Co-polar differential phase: PhiDP, uPhiDP, uPhiDPu
+Specific differential phase: KDP, uKDP, uKDPu
+Signal quality parameters: SQI, SQIu, SQIv, SQIvu
+Temperature: TEMP
+Position of the range bin respect to the ISO0: ISO0
+
+History
+
+    V0.1 20160629 Jordi Figueras i Ventura, MeteoSwiss, first prototype
+        
+    read_rainbow_wrl
+
+"""
+
+# specific modules for this function
+import os
+import wradlib as wrl
+
+import datetime
+
+import numpy as np
+
+from ..config import FileMetadata, get_fillvalue
+from ..io.common import make_time_unit_str, _test_arguments
+from ..core.radar import Radar
+from ..exceptions import MissingOptionalDependency
+
+RAINBOW_FIELD_NAMES = {
+    'W': 'spectrum_width',
+    'Wv': 'spectrum_width_vv', # non standard name
+    'Wu': 'unfiltered_spectrum_width', # non standard name
+    'Wvu': 'unfiltered_spectrum_width_vv', # non standard name
+    'V': 'velocity',
+    'Vv': 'velocity_vv', # non standard name
+    'Vu': 'unfiltered_velocity', # non standard name
+    'Vvu': 'unfiltered_velocity_vv', # non standard name
+    'dBZ': 'reflectivity',
+    'dBZv': 'reflectivity_vv',      # non standard name
+    'dBuZ': 'unfiltered_reflectivity', # non standard name
+    'dBuZv': 'unfiltered_reflectivity_vv', # non standard name
+    'ZDR': 'differential_reflectivity',
+    'ZDRu': 'unfiltered_differential_reflectivity', # non standard name
+    'RhoHV': 'cross_correlation_ratio',
+    'RhoHVu': 'unfiltered_cross_correlation_ratio', # non standard name
+    'PhiDP': 'differential_phase',
+    'uPhiDP': 'uncorrected_differential_phase', # non standard name
+    'uPhiDPu': 'uncorrected_unfiltered_differential_phase', # non standard name 
+    'KDP': 'specific_differential_phase',
+    'uKDP': 'uncorrected_specific_differential_phase', # non standard name
+    'uKDPu': 'uncorrected_unfiltered_specific_differential_phase', # non standard name
+    'SQI': 'signal_quality_index', # non standard name
+    'SQIv': 'signal_quality_index_vv', # non standard name
+    'SQIu': 'unfiltered_signal_quality_index', # non standard name
+    'SQIvu': 'unfiltered_signal_quality_index_vv', # non standard name
+    'TEMP': 'temperature', # non standard name
+    'ISO0': 'iso0', # non standard name
+}
+
+
+def read_rainbow_wrl(filename, field_names=None, additional_metadata=None, file_field_names=False, exclude_fields=None, **kwargs):
+    """
+    Read a RAINBOW file.
+    
+    Parameters
+    ----------
+    filename : str
+        Name of the RAINBOW file to read.
+    field_names : dict, optional
+        Dictionary mapping RAINBOW field names to radar field names. If a
+        data type found in the file does not appear in this dictionary or has
+        a value of None it will not be placed in the radar.fields dictionary.
+        A value of None, the default, will use the mapping defined in the
+        Py-ART configuration file.
+    additional_metadata : dict of dicts, optional
+        Dictionary of dictionaries to retrieve metadata during this read.
+        This metadata is not used during any successive file reads unless
+        explicitly included.  A value of None, the default, will not
+        introduct any addition metadata and the file specific or default
+        metadata as specified by the Py-ART configuration file will be used.
+    file_field_names : bool, optional
+        True to use the MDV data type names for the field names. If this
+        case the field_names parameter is ignored. The field dictionary will
+        likely only have a 'data' key, unless the fields are defined in
+        `additional_metadata`.
+    exclude_fields : list or None, optional
+        List of fields to exclude from the radar object. This is applied
+        after the `file_field_names` and `field_names` parameters.
+    
+    
+    Returns
+    -------
+    radar : Radar
+        Radar object containing data from RAINBOW file.
+    
+    """
+    
+    # check if it is the right file. Open it and read it
+    bfile = os.path.basename(filename)
+    if bfile.endswith('.vol') or bfile.endswith('.azi') or bfile.endswith('.ele'):
+    
+        # create metadata retrieval object
+        if field_names is None:
+            field_names = RAINBOW_FIELD_NAMES
+        filemetadata = FileMetadata('RAINBOW', field_names, additional_metadata,
+                                file_field_names, exclude_fields)
+                                
+        rbf=wrl.io.read_Rainbow(filename, loaddata=True)
+            
+        # check if file contains a known field
+        # all slices should have the same data type
+        nslices=int(rbf['volume']['scan']['pargroup']['numele'])
+        if nslices > 1:
+            datatype=rbf['volume']['scan']['slice'][0]['slicedata']['rawdata']['@type']
+        else:
+            datatype=rbf['volume']['scan']['slice']['slicedata']['rawdata']['@type']
+            
+        field_name = filemetadata.get_field_name(datatype)
+        if field_name is not None:
+        
+            # get definitions from filemetadata class
+            latitude = filemetadata('latitude')
+            longitude = filemetadata('longitude')
+            altitude = filemetadata('altitude')     
+            metadata = filemetadata('metadata')
+            sweep_start_ray_index = filemetadata('sweep_start_ray_index')
+            sweep_end_ray_index = filemetadata('sweep_end_ray_index')
+            sweep_number = filemetadata('sweep_number')
+            sweep_mode = filemetadata('sweep_mode')
+            fixed_angle = filemetadata('fixed_angle')
+            elevation = filemetadata('elevation')
+            _range = filemetadata('range')
+            azimuth = filemetadata('azimuth')
+            _time = filemetadata('time')
+            field_dic = filemetadata(field_name)
+            
+            # other metadata
+            scan_rate=filemetadata('scan_rate')
+            frequency=filemetadata('frequency')
+            
+            # get general file information          
+                
+            # position
+            if 'sensorinfo' in rbf['volume'].keys() :
+                latitude['data']=np.array([rbf['volume']['sensorinfo']['lat']], dtype='float64')
+                longitude['data']=np.array([rbf['volume']['sensorinfo']['lon']], dtype='float64')
+                altitude['data']=np.array([rbf['volume']['sensorinfo']['alt']], dtype='float64')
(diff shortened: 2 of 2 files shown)


pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+
+    V0.1 20160629 Jordi Figueras i Ventura, MeteoSwiss, first prototype
+        
+    read_rainbow_wrl
+
+"""
+
+# specific modules for this function
+import os
+import wradlib as wrl
```
jjhelmus: wradlib is an optional dependency in Py-ART and therefore it cannot be imported unconditionally.  You can add a try/except block similar to the one used in the `sinarame_h5.py` module for importing h5py:

``` Python
try:
    import wradlib as wrl
    _WRADLIB_AVAILABLE = True
except:
    _WRADLIB_AVAILABLE = False
```

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+Signal quality parameters: SQI, SQIu, SQIv, SQIvu
+Temperature: TEMP
+Position of the range bin respect to the ISO0: ISO0
+
+History
+
+    V0.1 20160629 Jordi Figueras i Ventura, MeteoSwiss, first prototype
+        
+    read_rainbow_wrl
+
```
jjhelmus: This section should be moved to the function docstring.  Module level docstring need to follow the layout used in other modules so that they can be used to generate the documentation.

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+        after the `file_field_names` and `field_names` parameters.
+    
+    
+    Returns
+    -------
+    radar : Radar
+        Radar object containing data from RAINBOW file.
+    
+    """
+
```
jjhelmus: A check should be added here to see if wradlib is available, see `sinarame_h5.py` for an example

``` Python
if not _WRADLIB_AVAILABLE:
   raise MissingOptionalDepedency("wradlib is required to use read_rainbow_wrl but it is not installed")

File: pyart/aux_io/rainbow_wrl.py [hunk 4]
```diff
@@ -0,0 +1,493 @@
...
+    
+    
+    Returns
+    -------
+    radar : Radar
+        Radar object containing data from RAINBOW file.
+    
+    """
+    
+    # check if it is the right file. Open it and read it
```
jjhelmus: `_test_arguments(kwargs)` should likely be called here to check for non empty keyword arguments.

jjhelmus:
@jfigui Thanks for the submission! I know a number of people who would be interesting in being able to read rainbow5 files with Py-ART.  I added a few specific comments in the code which should be addressed.  Additionally, a few more general notes:

* Py-ART follows the [PEP8](https://www.python.org/dev/peps/pep-0008/) standard which can be checked using the [pycodestyle](https://pypi.python.org/pypi/pycodestyle) tool.  There are a number of placed in this code that do not follow this convention.
* There are some very long and deeply nested if/else blocks in the `read_rainbow_wrf` function which make following the flow of the code difficult.  Can any of these be simplified or refactored into functions? 

To begin with the first if block can be removed by initially checking for unsupported files:

``` Python
supported_file = bfile.endswith('.vol') or bfile.endswith('.azi') or bfile.endswith('.ele')
if not supported_file:
   raise ValueError('Only data files with extension .vol, .azi or .ele are supported')
```

The if block for unknown field names can also be removed in a similar manner:

``` Python
if field_name is None:
    raise alueError('Field Name Unknown')
```

kmuehlbauer:
@jfigui @jjhelmus If there are any questions concerning the used wradlib-functions, I'll be happy to help.  Lately we removed some bug in the rainbow reader found by community. 

I came along several rainbow flavours, so I can also contribute to tests and code hardening.

jfigui:
@kmuehlbauer  and @jjhelmus. I will apply the suggested changes and hopefully tomorrow I will resubmit the function.

Function testing is always welcome!

jjhelmus:
If someone can provide some additional example files for the various combinations of scan types (PPI, RHI, etc) and single/multiple sweeps I would really appreciate it.  Currently the only rainbow file I have to test with is from the [wradlib-data repository](https://github.com/wradlib/wradlib-data) which is  a single scan PPI file.

jfigui:
@jjhelmus I just commited a new version of the rainbow reader:

I implemented all your suggestions and made sure the code is formally complying with Python using pycodestyle.

I also got read of most of the nested if-else and decluttered the read_rainbow_wrl function by creating several auxiliary functions.

Let me know what else I should change.

As for the rainbow files I will send you examples of .azi, .vol and .ele files with two different versions of the file format.

kmuehlbauer:
> Currently the only rainbow file I have to test with is from the wradlib-data repository which is a single scan PPI file.

I have some rainbow files in the backyard. Move them into the wradlib-data repo soon (hopefully).

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
+"""
+pyart.aux_io.rainbow
+====================
+
+Routines for reading RAINBOW files (Used by SELEX) using the wradlib library
+
```
jjhelmus: Add the following lines to the section so that the module function are included in the documentation:
```
.. autosummary::
    :toctree: generated/
```

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+    instrument_parameters = dict()
+    instrument_parameters.update({'frequency': frequency})
+
+    return Radar(_time, _range, fields, metadata, scan_type, latitude,
+                 longitude, altitude, sweep_number, sweep_mode, fixed_angle,
+                 sweep_start_ray_index, sweep_end_ray_index, azimuth,
+                 elevation, instrument_parameters=instrument_parameters)
+
+
+def get_angle(ray_info, angle_step=None, scan_type='ppi'):
```
jjhelmus: Would it make sense to have this function be private, typically this is indicated by prefixing the function with an underscore, `_get_angle`.

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+        angle_stop = angle_start+angle_step
+
+    moving_angle = np.angle((np.exp(1.j * np.deg2rad(angle_start))
+                            + np.exp(1.j * np.deg2rad(angle_stop))) / 2.,
+                            deg=True)
+
+    return moving_angle, angle_start, angle_stop
+
+
+def get_data(rawdata, nrays, nbins):
```
jjhelmus: Same as above, `_get_data`?

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+
+    data = np.reshape(data, [nrays, nbins])
+    mask = np.reshape(mask, [nrays, nbins])
+
+    masked_data = np.ma.array(data, mask=mask, fill_value=get_fillvalue())
+
+    return masked_data
+
+
+def get_time(date_sweep, time_sweep, first_angle_start, last_angle_stop,
```
jjhelmus: `_get_time`?

pyart/aux_io/__init__.py
```diff
@@ -42,5 +42,6 @@
 from .odim_h5 import read_odim_h5
 from .gamic_hdf5 import read_gamic
 from .sinarame_h5 import read_sinarame_h5
+from .rainbow_wrl import read_rainbow_wrl
```
jjhelmus: The `read_rainbow_wrl` function should be added to the docstring at the top of this module so that the function is included in the User Reference Manual.

jjhelmus:
@jfigui Great job on the refactor!  This is nearly ready to go, just few minor comments that I made inline.  Also a line needs to be added to the `doc/source/dev_reference/aux_io.rst` file so that the module gets added to the developer reference manual.  

I'm happy to fix up these few issue before merging if you would prefer, if not I'll wait on you.

pyart/aux_io/rainbow_wrl.py
```diff
@@ -0,0 +1,493 @@
...
+        raise ValueError(
+            'Only data files with extension .vol, .azi or .ele are supported')
+
+    # create metadata retrieval object
+    if field_names is None:
+        field_names = RAINBOW_FIELD_NAMES
+    filemetadata = FileMetadata('RAINBOW', field_names, additional_metadata,
+                                file_field_names, exclude_fields)
+
+    rbf = wrl.io.read_Rainbow(filename, loaddata=True)
```
kmuehlbauer: @jfigui @jjhelmus We might get a bit of speed-up here, depending on the situation, if we do not load the data here. That being the case, if you just don't need all moments from the file. We can load the data, when the actual moment is processed. See snippet:

``` python
rbdict = wrl.io.read_Rainbow(filename, loaddata=False)
# checking header only rbblob
rbblob = rbdict['volume']['scan']['slice']['slicedata']['rawdata']

# get data dict of this blob
data = wrl.io.get_RB_blob_from_file(filename, rbblob)
```
This could be done as a follow-up PR, I'm volunteering.

jjhelmus: @kmuehlbauer Great idea.  Some of the read function support a `delay_field_loading` argument which delays loading of the field data until the data is requested from the field dictionary.  Adding this as a follow up for rainbow files sounds like a good addition.  A good example of this is the [_NetCDF4VariableDataExtractor class](https://github.com/ARM-DOE/pyart/blob/master/pyart/io/cfradial.py#L354-L384) in the cfradial module.

jfigui: @jjhelmus, since I am going on holiday today and what is left are just minor changes could you be so kind to make them yourself? Feel free to change whatever is needed. I agree that probably the auxiliary functions are better declared as private.
@kmuehlbauer, yours is a good idea but Rainbow raw data files tipically contain only one moment and sooner or later you have to read it so I do not think there would be a major improvement.
If we created a function that would read the files of several moments at once to load the Radar object then probably would be worth it.
Nevertheless feel free to modify the function as you please.

kmuehlbauer: > Rainbow raw data files tipically contain only one moment and sooner or later you have to read it so I do not think there would be a major improvement.

Yes, didn't thought about having the moments in different files. We can try anyway and have some performance checks.

> If we created a function that would read the files of several moments at once to load the Radar object then probably would be worth it.

This is even more intriguing idea. We should definitely try this out. @jfigui I have some multi-scan .vol files as well as single scan .azi and .ele files to test. How is your data filename build up? Here I have eg. `YYYYMMDDHHMMSS00mom.vol`. Then we could build in some magic to retrieve all the other corresponding files. Or we can try to give a filelist to the reading function. Many opportunities. I'll give it a try, but this might take some time.

jfigui: @kmuehlbauer the default data file name of Rainbow is standard and it is what you mentioned. However, there may always be a possibility of the user having changed the name. Anyway, if you keep the standard file name the code to do that should be something similar to that:

# read master data type
        # if radar object does not exist yet create it
        # otherwise add the new field
        daydir=voltime.strftime('%Y-%m-%d')
        datapath=cfg['datapath']+cfg['ScanList'][0]+daydir+'/'
        
        fdatetime=voltime.strftime('%Y%m%d%H%M%S')+'00'
        filename=glob.glob(datapath+fdatetime+datatype_rainbow[0]+'.*')
        if radar is None:
            radar = pyart.aux_io.read_rainbow_wrl(filename[0])
        else:
            radar_aux = pyart.aux_io.read_rainbow_wrl(filename[0])
            for field_name in radar_aux.fields.keys():
                break
            field_data = radar_aux.fields[field_name]['data']
            field_metadata = pyart.config.get_metadata(field_name)
            field_metadata['data'] = field_data
            radar.add_field(field_name, field_metadata)

        # add other fields in the same scan
        for i in range(1, ndatatypes_rainbow):
            filename=glob.glob(datapath+fdatetime+datatype_rainbow[i]+'.*')
            radar_aux = pyart.aux_io.read_rainbow_wrl(filename[0])
            for field_name in radar_aux.fields.keys():
                break
            field_data = radar_aux.fields[field_name]['data']
            field_metadata = pyart.config.get_metadata(field_name)
            field_metadata['data'] = field_data
            radar.add_field(field_name, field_metadata)

With the difference that the metadata would be read only once and then the other fields simply added to the existing radar object. I am using that lines in my signal processor.

I hope this helps

jjhelmus:
I'm setting aside tomorrow as a Py-ART day and top on the list with be getting this PR merged.  Thanks again for the contribution @jfigui!

jjhelmus:
Merged via command line.  Thanks @jfigui!

jfigui:
@jjhelmus I just realized the azimuth angles are set in the range [-180, 180] degrees from north. Tipically, they are set as [0, 360] degrees from north. Should I change that?

jjhelmus:
@jfigui The [Cf/Radial specifications](https://github.com/NCAR/CfRadial) does not appear to specify if the azimuth angles should be [-180, 180] or [0, 360] although in Py-ART they angles tend to be [0, 360].  What format is used in the Rainbow files?  If it is [-180, 180] then I am fine keeping that format although would have no disagreement with converting these to [0, 360].

jfigui:
@jjhelmus The original rainbow format is [0, 360]. I would put it as [0, 360]. It is just an extra line of code.
Actually at least internally I would force the radar object to have a single format (for my taste [0, 360]). It would avoid stupid mistakes. In any case I would recommend to specify the ranges of the azimuth, elevation in the documentation.

jjhelmus:
In that case [0, 360] sounds like the best option.  I'll wait for others to chime in on the topic on the mailing list, but will add some text to the Radar docstring about the recommended range for angles.

josephhardinee:
My vote is for [0,360] for azimuth. Should elevation be the same or would [-180,180] make more sense for elevation?

jjhelmus:
Elevation angles outside of [0, 180] should only happen if you scan below the horizon, correct?  I would expect that the be relatively rare (although I think ARM has done this).  I like using small negative angles when this occurs on the one side but jumping from 180 to -180 on the other side seems odd. I would almost prefer [-20, 200] or something similar but formulating this into something concrete would be difficult.

scollis:
More like -90 to 90

I think with horizon to horizon scans the azimuth should change as we go through 90 

0 to -90 for downward pointing traders

-sent from a mobile device-

scollis:
What about airborne radars?

-sent from a mobile device-

josephhardinee:
Well there's two ways of looking at it. Scott is right that you only need 180 degrees for elevation in a polar coordinate system.  Only allowing 180 degrees though means you need to handle modifying elevation as a function of azimuth which can make a few things confusing (Cross zenith and you need to rotate the azimuth to the new direction).  This would also break a lot of the fixed angle variable (You now need to break up horizon-horizon into two sweeps).

While less mathematically correct, [-180,180] avoids these issues (As does 0-360). 

We (and many others) do a lot of below the horizon scanning, though we often don't release all of those datasets (A lot are for trouble shooting and maintenance).

We can always just mandate(suggest) azimuth and leave elevation to the individual formats as well.

jfigui:
At MeteoSwiss we have a regular scan at -0.2 elevation (do not forget that we have radars as high as 3000 m). I would favour having the possibility to operate on the entire cercle (there are below the horizon looking airborne radars after all) so my vote goes for [-180, 180]

jjhelmus:
Probably should have mentioned in my previous comment that I was thinking of elevation angles for RHI scans, where at least the files I've come across tend to use elevation angles of [0, 180] to indicate a horizon to horizon scan.  My guess is that this is not a universal convention.

From the discussion here my plan is to add a recommendation to express azimuth angles as [0, 360] and a preference for elevation angles in [-180, 180] but not to forbid other representations.  Does this seem reasonable?

josephhardinee:
It does to me.

kirknorth:
:+1:

jfigui:
That seems pretty reasonable but in my opinion we should make sure that the native Py-ART reading functions comply with the recommendation. I have seen pieces of code where that is not true.

jjhelmus:
Sounds like the proposal is acceptable to everyone.  I'll create an issue so that I don't forgot to add this to the documentation.

jjhelmus:
> we should make sure that the native Py-ART reading functions comply with the recommendation. I have seen pieces of code where that is not true.

I agree, can you provide a list of where you have found this convention not followed in the Py-ART source code?  I'd be be happy to fix these cases.

jfigui:
@jjhelmus from memory I have spotted odim_h5.py (and Rainbow). If I see other functions I will let you know.

jfigui:
I also found out that the field frequency has to be declared as an array, otherwise the writing function complains. I will have to submit another pull request for the rainbow reader.

PR #563, opened 2016-08-05
MAINT: Change default gridding algorithm

jjhelmus:
Use the faster 'map_gates_to_grid' algorithm as the default gridding algorithm
when using the grid_from_radars function.


jjhelmus:
Though this had been done a long time ago but for some reason never made it in.

PR #565, opened 2016-08-09
Advection calculation and shifting of grids

scollis:
This pull request covers the calculation of the displacement between two grids using phase correlation and treating the grid as an image. The user must pick the field and the level. The code can return a number of pixels (tuple) or, using the time difference, a velocity. 
It also contains code to shift, using ndimage, a grid by a certain number of pixels and trim the edges. 

This is the first of two pull requests aimed at implementing advection correction into Py-ART. The next pull request will enable the addition of two grids by advecting them to a common time step. 

More detailed examples and cookbook recipes will also follow

### DIFF
--- a/pyart/retrieve/__init__.py
+++ b/pyart/retrieve/__init__.py
@@ -28,7 +28,7 @@ from .gate_id import map_profile_to_gates, fetch_radar_time_profile
 from .simple_moment_calculations import calculate_snr_from_reflectivity
 
 try:
-    from .advection import  grid_displacememt_pc, grid_shift
+    from .advection import  grid_displacement_pc, grid_shift
     _ADVECTION_AVAILABLE = True
 except:
     _ADVECTION_AVAILABLE = False
--- a/pyart/retrieve/__init__.py
+++ b/pyart/retrieve/__init__.py
@@ -27,4 +27,12 @@ from .echo_class import steiner_conv_strat
 from .gate_id import map_profile_to_gates, fetch_radar_time_profile
 from .simple_moment_calculations import calculate_snr_from_reflectivity
 
+try:
+    from .advection import  grid_displacememt_pc
+    _ADVECTION_AVAILABLE = True
+except:
+    _ADVECTION_AVAILABLE = False
+
+
+
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/retrieve/__init__.py
+++ b/pyart/retrieve/__init__.py
@@ -26,11 +26,6 @@ from .kdp_proc import kdp_maesaka
 from .echo_class import steiner_conv_strat
 from .gate_id import map_profile_to_gates, fetch_radar_time_profile
 from .simple_moment_calculations import calculate_snr_from_reflectivity
-
-try:
-    from .advection import  grid_displacement_pc, grid_shift
-    _ADVECTION_AVAILABLE = True
-except:
-    _ADVECTION_AVAILABLE = False
+from .advection import  grid_displacement_pc, grid_shift
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/retrieve/__init__.py
+++ b/pyart/retrieve/__init__.py
@@ -33,6 +33,4 @@ try:
 except:
     _ADVECTION_AVAILABLE = False
 
-
-
 __all__ = [s for s in dir() if not s.startswith('_')]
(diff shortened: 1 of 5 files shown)


scollis:
Can only see one warning here: https://ci.appveyor.com/project/JonathanHelmus/pyart-l711v/build/1.0.389/job/79gleqy7f41xolab#L1025

Looks like a windows + py3.5 future warning.. test still passes but need to be careful

pyart/retrieve/__init__.py
```diff
@@ -27,4 +27,10 @@
 from .gate_id import map_profile_to_gates, fetch_radar_time_profile
 from .simple_moment_calculations import calculate_snr_from_reflectivity
 
+try:
```
jjhelmus: Why does this need to be optionally imported?  Everything in the module is a required dependency

pyart/retrieve/advection.py
```diff
@@ -0,0 +1,165 @@
+"""
+pyart.retrieve.advection
+=========================
+
+
+.. autosummary::
+    :toctree: generated/
+
```
jjhelmus: You need to list out the functions in this module here

pyart/retrieve/__init__.py
scollis: I think the main concern was ndimage which I do not know if it automatically ships with scipy


Dr Scott Collis
Argonne National Laboratory
Mb: +1 630 235 8025
Of: +1 630 252 0550

pyart/testing/__init__.py
```diff
@@ -19,6 +19,7 @@
     make_velocity_aliased_radar
     make_empty_grid
     make_target_grid
+    make_normal_storm
```
jjhelmus: Add the other functions to this list

pyart/retrieve/__init__.py
jjhelmus: It does

pyart/testing/__init__.py
```diff
@@ -46,5 +47,6 @@
 from .sample_objects import make_empty_rhi_radar
 from .sample_objects import make_velocity_aliased_rhi_radar
 from .tmpdirs import InTemporaryDirectory
+from .sample_objects import make_target_grid, make_storm_grid, make_normal_storm
```
scollis: my bad.. will fix
Will also remove the conditional import

jjhelmus:
Can you run pep8 on these modification?  There are a lot of places where it breaks convention.

pyart/retrieve/advection.py
```diff
@@ -0,0 +1,165 @@
...
+    -------
+    displacement : two-tuple
+         integers if pixels, otherwise floats. Result of the calculation
+
+
+    """
+    #create copies of the data
+
+    ige1 = copy.deepcopy(grid2.fields[var]['data'][level,:,:])
+    ige2 = copy.deepcopy(grid1.fields[var]['data'][level,:,:])
```
jjhelmus: .copy() can be used here, deepcopy is not needed.

hhuangmeso:
Could this function be used to correct the upper level shift of echo due to the advection in the time interval of a volume scan? If can, a better structure can be deduced from the gridded data.

pyart/retrieve/advection.py
```diff
@@ -0,0 +1,165 @@
...
+
+    #set anything below valid min to valid min, this helps nuke _fill_values
+
+    ige1[np.where(ige1 < grid1.fields[var]['valid_min'])] = \
```
jjhelmus: A fallback will be needed when the dictionary does not have 'valid_min'

scollis:
@hhuangmeso Not quite yet.. This works on mass advection of grids treating the 3d image as a snapshot.. What you are looking for needs to be implemented in the gridding code itself. Allowing the gridding code to advect the gate positions within the grid with an advection determined by this code.. This is something I want to do one day, ie correct the "tilt" in storms seen by the radar

scollis:
@jjhelmus Thank you for all the fixes. The simple ones are done and the more tricky one, fixing the fall through, is fixed here: https://github.com/scollis/pyart/blob/advection_shift/pyart/retrieve/advection.py#L68

I decided using an exception was easier and it basically replaces all values where there is a fill_value with the minimum value from the set not including fill_values.. I tested on a simple case and it worked there and tests are passing on my machine

hhuangmeso:
@scollis Yes, you're right. I noticed that the speed of transform Radar Class to Grid Class is kinda slower than Reorder/Radx. Could the computational efficiency of this transformation be improved in the future?  Also, the students of our group apply an adjustment of tilt storm when using Reorder. Maybe someday I could merge it into pyart.

PR #572, opened 2016-08-16
New gate filter based on texture of polarimetric variables

jfigui:
Dear Py-ART users,

I have added the following functions:
In filters/gatefilter.py:  moment_based_gate_filter2 filters out gates according to radial textures
In util/sigmath.py: texture_along_ray computes the texture along a a ray with a user defined length.

Please let me know if this is up to standard.

Greetings from Switzerland,

Jordi

### DIFF
--- a/pyart/util/sigmath.py
+++ b/pyart/util/sigmath.py
@@ -34,11 +34,10 @@ def texture(myradar, var):
     return tex
 
 
-def texture_along_ray(myradar, var, wind_size=7, min_valid=3):
+def texture_along_ray(myradar, var, wind_size=7):
     """
     Compute field texture along ray using a user specified
-    window size. If the number of valid range bins within the
-    window is lower than min_valid the bin is set to invalid.
+    window size.
 
     Parameters
     ----------
@@ -48,9 +47,6 @@ def texture_along_ray(myradar, var, wind_size=7, min_valid=3):
         Name of the field which texture has to be computed
     wind_size : int
         Optional. Size of the rolling window used
-    min_valid : int
-        Optional. Minimum number of valid range bins to
-        compute the texture
 
     Returns
     -------
--- a/pyart/util/sigmath.py
+++ b/pyart/util/sigmath.py
@@ -32,3 +32,38 @@ def texture(myradar, var):
         tex[timestep, 0:4] = np.ones(4) * ray[0]
         tex[timestep, -5:] = np.ones(5) * ray[-1]
     return tex
+
+
+def texture_along_ray(myradar, var, wind_size=7, min_valid=3):
+    """
+    Compute field texture along ray using a user specified
+    window size. If the number of valid range bins within the
+    window is lower than min_valid the bin is set to invalid.
+
+    Parameters
+    ----------
+    myradar : radar object
+        The radar object where the field is
+    var : str
+        Name of the field which texture has to be computed
+    wind_size : int
+        Optional. Size of the rolling window used
+    min_valid : int
+        Optional. Minimum number of valid range bins to
+        compute the texture
+
+    Returns
+    -------
+    tex : radar field
+        the texture of the specified field
+
+    """
+    half_wind = int((wind_size-1)/2)
+    fld = myradar.fields[var]['data']
+    tex = np.ma.zeros(fld.shape)
+    for timestep in range(tex.shape[0]):
+        ray = np.ma.std(rolling_window(fld[timestep, :], wind_size), 1)
+        tex[timestep, half_wind:-half_wind] = ray
+        tex[timestep, 0:half_wind] = np.ones(half_wind) * ray[0]
+        tex[timestep, -half_wind:] = np.ones(half_wind) * ray[-1]
+    return tex
--- a/pyart/filters/__init__.py
+++ b/pyart/filters/__init__.py
@@ -15,10 +15,11 @@ Filtering radar data
 
     GateFilter
     moment_based_gate_filter
+    moment_and_texture_based_gate_filter
 
 """
 
 from .gatefilter import GateFilter, moment_based_gate_filter
-from .gatefilter import moment_based_gate_filter2
+from .gatefilter import moment_and_texture_based_gate_filter
 
 __all__ = [s for s in dir() if not s.startswith('_')]
--- a/pyart/filters/__init__.py
+++ b/pyart/filters/__init__.py
@@ -19,5 +19,6 @@ Filtering radar data
 """
 
 from .gatefilter import GateFilter, moment_based_gate_filter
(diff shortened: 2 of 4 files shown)


hhuangmeso:
Actually, the folding of the PHIDP is not considered in  the computation of PHIDP texture. Also, in my opinion, ZH texture should be calculated in the linear domain. I do use PHIDP texture to remove non meteorological echo like @tjlang does.

jfigui:
@hhuangmeso, the functions assume that data manipulations have been done a-priori. It is not meant to be stand-alone but integrated in a processing chain. As for whether Zh texture should be computed in the linear or in the log domain I do not think it really matters. We could even think of generalizing the function so that the user decides which fields and thresholds to use among a wider selection that what I have put here.
The function as it is is the one applied to the X-band radar at MeteoSwiss. I am open to suggestions that improve the performance.

pyart/util/sigmath.py
```diff
@@ -32,3 +32,38 @@ def texture(myradar, var):
         tex[timestep, 0:4] = np.ones(4) * ray[0]
         tex[timestep, -5:] = np.ones(5) * ray[-1]
     return tex
+
+
+def texture_along_ray(myradar, var, wind_size=7, min_valid=3):
+    """
+    Compute field texture along ray using a user specified
+    window size. If the number of valid range bins within the
+    window is lower than min_valid the bin is set to invalid.
```
jjhelmus: The descriptions of this function does not match what is being done in the code.  The most obvious difference is that `min_valid` is not used at all.  Either to documentation or the code should be updated.

jjhelmus:
@jfigui This is a great addition to Py-ART, thanks for putting it together.  One concern I have is that the new function names are not very expressive, perhaps we can think of better names for these.   I would suggested `moment_and_texture_based_gate_filter` and `std_along_ray` but am open to discussion.  

Also the **sigmath.py** module is not used elsewhere in Py-ART and was more-or-less a dead module.  There may be more efficient methods for calculating standard deviation, for example as [described in this stack overflow post](http://stackoverflow.com/questions/18419871/improving-code-efficiency-standard-deviation-on-sliding-windows). I'd be happy to get this functionality into Py-ART even without a more efficient method which can always be added at a later date.

jfigui:
@jjhelmus I agree on changing the name of the moment_based_gate_filter. I am not that sure about texture_along_ray because although it is true that this is the local standard deviation texture is a concept widely used in the weather radar community.
I have no problem in changing texture_along_ray place if sigmath.py is deprecated. Where do you think I should put it?
As for the texture computation, I agree that it could be implemented more efficiently although I would need help there because I do not know enough Python. Initially I though to force a minimum number of valid neighbouring gates to compute the texture (and set it as non valid otherwise) . This would act effectivily as a speckle filter. However, I could not think of an efficient way to implement it so in the end I did not. It probably makes sense to simply compute the radial texture and putting an explicit speckle filter elsewhere in the processing chain.
So if you agree I will remove the parameter min_valid an update the function documentation. At this moment I would leave the function as it is (with the proposed change) and think about ways to improving it at a latter stage. I have other pressing things to do in the meantime.
Let me know what you think on all that and I will implement the changes.

tjlang:
A Cython implementation of a basic standard deviation calculation in a moving window along a ray can be found here:
https://github.com/CSU-Radarmet/CSU_RadarTools/blob/master/csu_radartools/calc_kdp_ray_fir.pyx#L95

The analysis makes use of the trick discussed in that stackoverflow.com post.

This is part of a KDP algorithm, where phase texture is used as a identifier for bad gates. The entire KDP algorithm, which makes 3 passes per ray, runs in 2 seconds or less on my 4-yo laptop when analyzing a busy radar volume, so the texture part is maybe only a third of that cost.

Anyway, this seemed germane to the discussion, and might be useful as a template for implementing fast texture processing in Py-ART.

scollis:
There are some texture codes in pyart but using the circular statistics method. At least I though I submitted it.. But the more the better as long as they are well named and documented 

-sent from a mobile device-

scollis:
Ok I was wrong. I had made my own implementations but not submitted a PR yet. See cells 5-7 here:
https://github.com/scollis/openvap/blob/master/development/cmac2/notebooks/full%20suite.ipynb

@jfigui After you clean and have the PR accepted I will submit a new PR following the lines of yours that implement the circular technique

kirknorth:
Sorry for the plug, but don't forget the texture implementation in [Py-RECT](https://github.com/kirknorth/pyrect/blob/master/pyrect/util/texture.py) everyone :wink: This uses SciPy's `ndimage` image processing library, takes on the order of microseconds to compute the texture field over a full radar volume, and properly handles contiguous-ray sweeps (e.g., PPI scans). And last but not least it's Py-ART ready!

jjhelmus:
I appears as if there are a number of methods which can be used to speed up the calculation of the texture (standard deviation) of a field.  I especially like @kirknorth's use of `ndimage.uniform_filter`.  That said I think this PR can be merged with the current implementation with the intention of replacing it will a faster texture calculation later.  

With the following changed I would be happy to merge this PR:

* Rename the `moment_based_gate_filter2` function to `moment_and_texture_based_gate_filter`
* Add the `moment_and_texture_based_gate_filter` to the docstring list of functions in `pyart/filters/__init__.py` and `pyart/filters/gatefilter.py`
* Adjust the parameter and documentation of the `texture_along_ray` function to match what is being performed.

Later a fast implementation of textures, both linear and circular, can be implement.  Also the **sigmath** module should be fleshed out and moved at some point.  Currently the module is not included in the documented and therefore should be considered for users to be an implementation detail subject to change in the future.  I am fine keeping it that way until it is more fleshed out and a proper home in Py-ART can be found for the procedures contained within.

I should let everyone know that ARM has tasked me with creating algorithms to detect and classify clutter, noise, and other echo signatures in the program's radars.  For this task I expect that Py-ART will grow some echo classification and fuzzy logic procedures and will be looking at Py-RECT and CSU_RadarTools for inspiration.

jjhelmus:
@jfigui Also, if you wanted to add a unit test or two for these function I would not object but do not feel obligated.

jfigui:
@jjhelmus I just commited the changes you requested but apparently they conflicted with the previous commit. I guess you will have to solve it. I am sorry. I did not include a unit test. I do not have the time for it.
As for texture computation methods, the conversation has been very stimulating. Unfortunately I cannot spend time testing the best option. I also think @kirknorth  solution would be a good one. As far as I am concerned my only wish/desire is that the dimensions of the local window where to compute the standard deviation are user defined (including whether computing it in 1D or 2D).

Have a nice weekend!

jjhelmus:
Thanks for updating this @jfigui, looks good to go.  I'll merge this manually in a few minutes.

PR #576, opened 2016-08-18
ENH: Simulated radial velocities from a profile

jjhelmus:
Add the `simulated_vel_from_profile` function to the pyart.util namespace.
This function creates a simulated radial velocity field from a profile of
horizontal winds.


jjhelmus:
Sample simulated radial velocity fields can be seen on the left of this figure:

sample_sims

PR #583, opened 2016-08-25
Computation of secondary moments, noise correction of RhoHV and pseudo-PPI from RHI volume

jfigui:
Dear Py-ART users,

I have added the following functionalities to Py-ART:
- A correction of RhoHV for noise in correct/noise.py
- The computation of the new fields: noise in dBZ, signal to noise ratio, logarithmic RhoHv (L parameter), and circular depolarization ratio in retrieve/simple_moment_calculations.
- A cross-section of a volume of RHIs along particular azimuths (pseudo-PPI) in util/xsect.py

@jjhelmus, I think we should set standard names for the new fields.

Let me know if you think these new functionalities can be implemented in Py-ART and are up to standard.

Greetings from Switzerland,

Jordi

### DIFF
--- a/pyart/correct/bias_and_noise.py
+++ b/pyart/correct/bias_and_noise.py
@@ -91,12 +91,8 @@ def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
     zdr = np.ma.power(10., 0.1*zdrdB)
     alpha = np.ma.power(10., 0.1*(nh-nv))
 
-    mask = np.ma.getmaskarray(urhohv)
-
     rhohv_data = urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h)))
     rhohv_data[rhohv_data > 1.] = 1.
-    rhohv_data[mask] = np.ma.masked
-    rhohv_data.set_fill_value(get_fillvalue())
 
     rhohv = get_metadata(rhohv_field)
     rhohv['data'] = rhohv_data
@@ -136,11 +132,7 @@ def correct_bias(radar, bias=0., field_name=None):
     else:
         raise KeyError('Field not available: ' + field_name)
 
-    mask = np.ma.getmaskarray(field_data)
-
     corr_field_data = field_data - bias
-    corr_field_data[mask] = np.ma.masked
-    corr_field_data.set_fill_value(get_fillvalue())
 
     if field_name.startswith('corrected_'):
         corr_field_name = field_name
--- a/pyart/correct/bias_and_noise.py
+++ b/pyart/correct/bias_and_noise.py
@@ -1,5 +1,5 @@
 """
-pyart.correct.noise
+pyart.correct.bias_and_noise
 ===================
 
 Corrects polarimetric variables for noise
@@ -92,12 +92,11 @@ def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
     alpha = np.ma.power(10., 0.1*(nh-nv))
 
     mask = np.ma.getmaskarray(urhohv)
-    fill_value = urhohv.get_fill_value()
 
     rhohv_data = urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h)))
     rhohv_data[rhohv_data > 1.] = 1.
-    rhohv_data.set_fill_value(fill_value)
-    rhohv_data.data[mask.nonzero()] = fill_value
+    rhohv_data[mask] = np.ma.masked
+    rhohv_data.set_fill_value(get_fillvalue())
 
     rhohv = get_metadata(rhohv_field)
     rhohv['data'] = rhohv_data
@@ -138,11 +137,10 @@ def correct_bias(radar, bias=0., field_name=None):
         raise KeyError('Field not available: ' + field_name)
 
     mask = np.ma.getmaskarray(field_data)
-    fill_value = field_data.get_fill_value()
 
     corr_field_data = field_data - bias
-    corr_field_data.set_fill_value(fill_value)
-    corr_field_data.data[mask] = fill_value
+    corr_field_data[mask] = np.ma.masked
+    corr_field_data.set_fill_value(get_fillvalue())
 
     if field_name.startswith('corrected_'):
         corr_field_name = field_name
--- a/pyart/correct/bias_and_noise.py
+++ b/pyart/correct/bias_and_noise.py
@@ -8,6 +8,7 @@ Corrects polarimetric variables for noise
     :toctree: generated/
 
     correct_noise_rhohv
+    correct_bias
 
 """
 
@@ -20,7 +21,9 @@ def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
                         zdr_field=None, nh_field=None, nv_field=None,
                         rhohv_field=None):
     """
-    Corrects RhoHV for noise
+    Corrects RhoHV for noise according to eq. 6 in Gourley et al. 2006.
+    This correction should only be performed if noise has not been subtracted
+    from the signal during the moments computation.
 
     Parameters
     ----------
@@ -39,13 +42,18 @@ def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
 
     Returns
     -------
-    rhohv : dictionary of dictionaries
+    rhohv : dict
         noise corrected RhoHV field
+
+    References
+    ----------
+    Gourley et al. Data Quality of the Meteo-France C-Band Polarimetric
+    Radar, JAOT, 23, 1340-1356
 
     """
     # parse the field parameters
     if urhohv_field is None:
-        rhohv_field = get_field_name('uncorrected_cross_correlation_ratio')
+        urhohv_field = get_field_name('uncorrected_cross_correlation_ratio')
     if snr_field is None:
         snr_field = get_field_name('signal_to_noise_ratio')
     if zdr_field is None:
@@ -86,14 +94,62 @@ def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
     mask = np.ma.getmaskarray(urhohv)
     fill_value = urhohv.get_fill_value()
 
-    rhohv_data = np.ma.masked_where(
-        mask, urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h))))
+    rhohv_data = urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h)))
+    rhohv_data[rhohv_data > 1.] = 1.
     rhohv_data.set_fill_value(fill_value)
     rhohv_data.data[mask.nonzero()] = fill_value
-    is_above1 = rhohv_data > 1.
-    rhohv_data[is_above1.nonzero()] = 1.
 
     rhohv = get_metadata(rhohv_field)
     rhohv['data'] = rhohv_data
 
     return rhohv
+
+
+def correct_bias(radar, bias=0., field_name=None):
+    """
+    Corrects a radar data bias. If field name is none the correction is
+    applied to horizontal reflectivity by default
+
+    Parameters
+    ----------
+    radar : Radar
+        radar object
+
+    bias : float
+        the bias magnitude
+
+    field_name: str
+        names of the field to be corrected
+
+    Returns
+    -------
+    corrected_field : dict
+        The corrected field
+
+    """
+    # parse the field parameters
+    if field_name is None:
+        field_name = get_field_name('reflectivity')
+
+    # extract fields from radar
+    if field_name in radar.fields:
+        field_data = radar.fields[field_name]['data']
+    else:
+        raise KeyError('Field not available: ' + field_name)
+
+    mask = np.ma.getmaskarray(field_data)
+    fill_value = field_data.get_fill_value()
+
+    corr_field_data = field_data - bias
+    corr_field_data.set_fill_value(fill_value)
+    corr_field_data.data[mask] = fill_value
+
+    if field_name.startswith('corrected_'):
+        corr_field_name = field_name
+    else:
+        corr_field_name = 'corrected_'+field_name
+
+    corr_field = get_metadata(corr_field_name)
+    corr_field['data'] = corr_field_data
+
+    return corr_field
--- a/pyart/correct/noise.py
+++ b/pyart/correct/noise.py
@@ -1,0 +1,99 @@
+"""
+pyart.correct.noise
+===================
+
+Corrects polarimetric variables for noise
+
+.. autosummary::
+    :toctree: generated/
+
+    correct_noise_rhohv
+
+"""
+
+import numpy as np
+
+from ..config import get_metadata, get_field_name, get_fillvalue
+
+
+def correct_noise_rhohv(radar, urhohv_field=None, snr_field=None,
+                        zdr_field=None, nh_field=None, nv_field=None,
+                        rhohv_field=None):
+    """
+    Corrects RhoHV for noise
+
+    Parameters
+    ----------
+    radar : Radar
+        radar object
+
+    urhohv_field : str
+        name of the RhoHV uncorrected for noise field
+
+    snr_field, zdr_field, nh_field, nv_field: str
+        names of the SNR, ZDR, horizontal channel noise in dBZ and vertical
+        channel noise in dBZ used to correct RhoHV
+
+    rhohv_field: str
+        name of the rhohv field to output
+
+    Returns
+    -------
+    rhohv : dictionary of dictionaries
+        noise corrected RhoHV field
+
+    """
+    # parse the field parameters
+    if urhohv_field is None:
+        rhohv_field = get_field_name('uncorrected_cross_correlation_ratio')
+    if snr_field is None:
+        snr_field = get_field_name('signal_to_noise_ratio')
+    if zdr_field is None:
+        zdr_field = get_field_name('differential_reflectivity')
+    if nh_field is None:
+        nh_field = get_field_name('noisedBZ_hh')
+    if nv_field is None:
+        nv_field = get_field_name('noisedBZ_vv')
+    if rhohv_field is None:
+        rhohv_field = get_field_name('cross_correlation_ratio')
+
+    # extract fields from radar
+    if urhohv_field in radar.fields:
+        urhohv = radar.fields[urhohv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + urhohv_field)
+    if snr_field in radar.fields:
+        snrdB_h = radar.fields[snr_field]['data']
+    else:
+        raise KeyError('Field not available: ' + snr_field)
+    if zdr_field in radar.fields:
+        zdrdB = radar.fields[zdr_field]['data']
+    else:
+        raise KeyError('Field not available: ' + zdr_field)
+    if nh_field in radar.fields:
+        nh = radar.fields[nh_field]['data']
+    else:
+        raise KeyError('Field not available: ' + nh_field)
+    if nv_field in radar.fields:
+        nv = radar.fields[nv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + nv_field)
+
+    snr_h = np.ma.power(10., 0.1*snrdB_h)
+    zdr = np.ma.power(10., 0.1*zdrdB)
+    alpha = np.ma.power(10., 0.1*(nh-nv))
+
+    mask = np.ma.getmaskarray(urhohv)
+    fill_value = urhohv.get_fill_value()
+
+    rhohv_data = np.ma.masked_where(
+        mask, urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h))))
+    rhohv_data.set_fill_value(fill_value)
+    rhohv_data.data[mask.nonzero()] = fill_value
+    is_above1 = rhohv_data > 1.
+    rhohv_data[is_above1.nonzero()] = 1.
+
+    rhohv = get_metadata(rhohv_field)
+    rhohv['data'] = rhohv_data
+
+    return rhohv
--- a/pyart/retrieve/simple_moment_calculations.py
+++ b/pyart/retrieve/simple_moment_calculations.py
@@ -108,7 +108,7 @@ def compute_noisedBZ(nrays, noisedBZ_val, range, ref_dist,
 
     noisedBZ = get_metadata(noise_field)
     noisedBZ['data'] = np.tile(noisedBZ_vec, (nrays, 1))
-    noisedBZ['data'].set_fill_value(get_fillvalue())
+#    noisedBZ['data'].set_fill_value(get_fillvalue())
 
     return noisedBZ
 
@@ -153,7 +153,7 @@ def compute_snr(radar, refl_field=None, noise_field=None, snr_field=None):
         raise KeyError('Field not available: ' + noise_field)
 
     snr_data = refl-noisedBZ
-    snr_data.set_fill_value(get_fillvalue())
+#    snr_data.set_fill_value(get_fillvalue())
 
     snr = get_metadata(snr_field)
     snr['data'] = snr_data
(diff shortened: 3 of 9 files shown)


pyart/correct/noise.py
```diff
@@ -0,0 +1,99 @@
...
+
+    snr_h = np.ma.power(10., 0.1*snrdB_h)
+    zdr = np.ma.power(10., 0.1*zdrdB)
+    alpha = np.ma.power(10., 0.1*(nh-nv))
+
+    mask = np.ma.getmaskarray(urhohv)
+    fill_value = urhohv.get_fill_value()
+
+    rhohv_data = np.ma.masked_where(
+        mask, urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h))))
```
josephhardinee: Can you provide a reference in the literature to the source of this equation?

jfigui: Hi @josephhardinee!

It is all about playing with signal and noise to subtract. I am not sure we ever published this one formally but I will try to see in my archives if I can find something.

Thanks for the comment,


Jordi

pyart/correct/noise.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns
+    -------
+    rhohv : dictionary of dictionaries
+        noise corrected RhoHV field
+
+    """
+    # parse the field parameters
+    if urhohv_field is None:
+        rhohv_field = get_field_name('uncorrected_cross_correlation_ratio')
```
jjhelmus: Should this be urhohv_field = get_field_name('uncorrected_cross_correlation_ratio')

pyart/correct/noise.py
```diff
@@ -0,0 +1,99 @@
...
+    zdr = np.ma.power(10., 0.1*zdrdB)
+    alpha = np.ma.power(10., 0.1*(nh-nv))
+
+    mask = np.ma.getmaskarray(urhohv)
+    fill_value = urhohv.get_fill_value()
+
+    rhohv_data = np.ma.masked_where(
+        mask, urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h))))
+    rhohv_data.set_fill_value(fill_value)
+    rhohv_data.data[mask.nonzero()] = fill_value
```
jjhelmus: This destroys the masking of `rhohv_data` setting them to fill_value, is there a reason why the data should be un-masked?  Can the following be used instead?  

rhohv_data = urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h)))

pyart/correct/noise.py
```diff
@@ -0,0 +1,99 @@
...
+    is_above1 = rhohv_data > 1.
+    rhohv_data[is_above1.nonzero()] = 1.
```
jjhelmus: The nonzero() call is not needed.  In fact `rho_hv[rhohv_data > 1] = 1` will work fine

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+
+    ref_dist: float
+        reference distance in Km
+
+    noise_field: str
+        name of the noise field to use
+
+    Returns
+    -------
+    noisedBZ : dictionary of dictionaries
```
jjhelmus: This is just a dictionary ` noisedBZ : dict`.

pyart/correct/noise.py
```diff
@@ -0,0 +1,99 @@
...
+    snr_field, zdr_field, nh_field, nv_field: str
+        names of the SNR, ZDR, horizontal channel noise in dBZ and vertical
+        channel noise in dBZ used to correct RhoHV
+
+    rhohv_field: str
+        name of the rhohv field to output
+
+    Returns
+    -------
+    rhohv : dictionary of dictionaries
```
jjhelmus: This is a dictionary not a dictionary of dictionaries.  `rhohv : dict`

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+
+    refl_field, noise_field : str
+        name of the reflectivity and noise field used for the calculations
+
+    snr_field : str
+        name of the SNR field
+
+    Returns
+    -------
+    snr : dictionary of dictionaries
```
jjhelmus: dictionary

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        raise KeyError('Field not available: ' + refl_field)
+    if noise_field in radar.fields:
+        noisedBZ = radar.fields[noise_field]['data']
+    else:
+        raise KeyError('Field not available: ' + noise_field)
+
+    mask = np.ma.getmaskarray(refl)
+    fill_value = refl.get_fill_value()
+
+    snr_data = np.ma.masked_where(mask, refl-noisedBZ)
```
jjhelmus: refl-noiseDBZ will take care of the masking.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        rhohv = radar.fields[rhohv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + rhohv_field)
+
+    mask = np.ma.getmaskarray(rhohv)
+    fill_value = rhohv.get_fill_value()
+    is_one = rhohv >= 1.
+    rhohv[is_one.nonzero()] = 0.9999
+
+    l_data = np.ma.masked_where(mask, -np.ma.log10(1.-rhohv))
```
jjhelmus: l_data = -np.ma.log10(1.-rhohv) is sufficient.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        zdrdB = radar.fields[zdr_field]['data']
+    else:
+        raise KeyError('Field not available: ' + zdr_field)
+
+    zdr = np.ma.power(10., 0.1*zdrdB)
+
+    mask = np.ma.getmaskarray(rhohv)
+    fill_value = rhohv.get_fill_value()
+
+    cdr_data = np.ma.masked_where(
```
jjhelmus: The use of np.ma.masked_where is not needed.

pyart/util/xsect.py
```diff
@@ -101,6 +102,89 @@ def cross_section_ppi(radar, target_azimuths):
...
+    ppi_nsweeps = len(target_elevations)
+    rhi_nsweeps = radar.nsweeps
+
+    for target_elevation in target_elevations:
+        for sweep_slice in radar.iter_slice():
+            sweep_elevations = radar.elevation['data'][sweep_slice]
+            ray_number = np.argmin(np.abs(sweep_elevations - target_elevation))
+            pppi_rays.append(ray_number + sweep_slice.start)
+
+    _range = _copy_dic(radar.range)
```
jjhelmus: The lines from here on seems to be nearly identical to those in `cross_section_ppi`.  Would it make sense to make these into a private function which both call?

jjhelmus:
@jfigui Thanks for the new features!  I made a few comment on suggested changes.  Mostly they deal with what looks to be a bit of a misunderstanding on how [masked arrays](http://docs.scipy.org/doc/numpy/reference/maskedarray.generic.html) work.  Typically masked arrays can be treated just like normal ndarray objects except that `np.ma` functions should be used (np.ma.power, np.ma.sin, etc).  The masking of the results is automatically taken care of by NumPy.

I would really appreciate some unit tests for these new function.

I agree that some new fields need to be added to the default config, feel free to add these.

The CI is failing for a unrelated reason which has been fixed in #586.

jfigui:
@jjhelmus thanks for checking the code. Indeed looks like my understanding of masked arrays is still a bit fuzzy.
I will correct the functions according to your suggestions and try to create some tests functions in the coming days.
I do think that repeated code in cross_section_xxx should be placed in a private function. I will do that.

jjhelmus:
Great @jfigui, let me know if you need any help with this.

jfigui:
@jjhelmus it just crossed my mind that the xsect functions look for the nearest neighbour without bothering to check if it is close enough. I would recommend to allow for a user defined angle distance tolerance (if angle distance is None the function would behave as it is now). What do you think about the idea?

jjhelmus:
I like the idea.  I would probably make sense to issue a warning when a angle is dropped because it fails the distance tolerance.  Feel free to either add that to this PR or submit a second PR with the change.

jfigui:
@jjhelmus, I have implemented the changes in the code but I am busy doing other stuff and I did not have the time to prepare the testing functions. Probably next week I will have some time to spend on this issue but I think I will need your help for the ingestion of test data.
Do you prefer that I first submit the changes in the code or I wait until I have also the testing functions ready?

jjhelmus:
@jfigui Go ahead and submit the changes, tests can be added later.  I might even be able to find some time in the next week to write them myself.

jfigui:
@jjhelmus, I finally managed to submit the requested corrections. In addition to that:
I renamed correct/noise.py to correct/bias_and_noise.py and added a function that corrects the bias (meant for reflectivity and differential reflectivity bias correction)
I found out there was a bug in join_radar at utils/radar_utils.py. It was not picking up the right fill_value so I corrected that.
I corrected rainbow_wrl.py to get the azimuth between [0, 360] and make the frequency an array as discussed in another thread.
Let me know if now the things are up to standard.

Greetings from Switzerland! (which is getting busier in the Open Source :))

pyart/correct/bias_and_noise.py
```diff
@@ -0,0 +1,155 @@
+"""
+pyart.correct.noise
```
jjhelmus: Should be pyart.correct.bias_and_noise

pyart/correct/bias_and_noise.py
```diff
@@ -0,0 +1,155 @@
...
+        nv = radar.fields[nv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + nv_field)
+
+    snr_h = np.ma.power(10., 0.1*snrdB_h)
+    zdr = np.ma.power(10., 0.1*zdrdB)
+    alpha = np.ma.power(10., 0.1*(nh-nv))
+
+    mask = np.ma.getmaskarray(urhohv)
+    fill_value = urhohv.get_fill_value()
```
jjhelmus: This can fail if urhohv is an ndarray not a masked array.  Can the `get_fillvalue` function be used here?

pyart/correct/bias_and_noise.py
```diff
@@ -0,0 +1,155 @@
...
+
+    rhohv_data = urhohv*np.ma.sqrt((1.+1./snr_h)*(1.+zdr/(alpha*snr_h)))
+    rhohv_data[rhohv_data > 1.] = 1.
+    rhohv_data.set_fill_value(fill_value)
+    rhohv_data.data[mask.nonzero()] = fill_value
```
jjhelmus: This should be `rhohv_data.data[mask] = np.ma.masked`  Additional I do not think setting the fill_value is really needed.

pyart/correct/bias_and_noise.py
```diff
@@ -0,0 +1,155 @@
...
+        field_name = get_field_name('reflectivity')
+
+    # extract fields from radar
+    if field_name in radar.fields:
+        field_data = radar.fields[field_name]['data']
+    else:
+        raise KeyError('Field not available: ' + field_name)
+
+    mask = np.ma.getmaskarray(field_data)
+    fill_value = field_data.get_fill_value()
```
jjhelmus: Same as above, I do not think this is need.

pyart/correct/bias_and_noise.py
```diff
@@ -0,0 +1,155 @@
...
+
+    corr_field_data = field_data - bias
+    corr_field_data.set_fill_value(fill_value)
+    corr_field_data.data[mask] = fill_value
```
jjhelmus: `corr_field_data.data[mask] = np.ma.masked` and remove the line which sets the fill_value.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+    """
+    # parse the field parameters
+    if noise_field is None:
+        noise_field = get_field_name('noisedBZ_hh')
+
+    noisedBZ_vec = noisedBZ_val+20.*np.ma.log10(1e-3*range/ref_dist)
+
+    noisedBZ = get_metadata(noise_field)
+    noisedBZ['data'] = np.tile(noisedBZ_vec, (nrays, 1))
+    noisedBZ['data'].set_fill_value(get_fillvalue())
```
jjhelmus: This is fine but probably is not needed.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        refl = radar.fields[refl_field]['data']
+    else:
+        raise KeyError('Field not available: ' + refl_field)
+    if noise_field in radar.fields:
+        noisedBZ = radar.fields[noise_field]['data']
+    else:
+        raise KeyError('Field not available: ' + noise_field)
+
+    mask = np.ma.getmaskarray(refl)
+    fill_value = refl.get_fill_value()
```
jjhelmus: `refl` may not have a `get_fill_value()` method.  Use the `get_fillvalue()` method or remove.

pyart/correct/bias_and_noise.py
jjhelmus: Sorry missed that this is using `.data`.  Setting .data should not be done as it is not preserved.  Use `rhohv_data[mask] = np.ma.masked` to mask values.  Modifying `.data` and expecting the mask to be updates is not a good idea.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+
+    snr_data = refl-noisedBZ
+    snr_data.set_fill_value(fill_value)
+    snr_data.data[mask.nonzero()] = fill_value
```
jjhelmus: `snr_data[mask] = np.ma.masked` although this should have propagated from the refl array automatically.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        l_field = get_field_name('logarithmic_cross_correlation_ratio')
+
+    # extract rhohv field from radar
+    if rhohv_field in radar.fields:
+        rhohv = radar.fields[rhohv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + rhohv_field)
+
+    mask = np.ma.getmaskarray(rhohv)
+    fill_value = rhohv.get_fill_value()
```
jjhelmus: Remove or use `get_fillvalue()`

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+
+    rhohv[rhohv >= 1.] = 0.9999
+    l_data = -np.ma.log10(1.-rhohv)
+    l_data.set_fill_value(fill_value)
+    l_data.data[mask.nonzero()] = fill_value
```
jjhelmus: `l_data[mask] = np.ma.masked`

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+        raise KeyError('Field not available: ' + rhohv_field)
+    if zdr_field in radar.fields:
+        zdrdB = radar.fields[zdr_field]['data']
+    else:
+        raise KeyError('Field not available: ' + zdr_field)
+
+    zdr = np.ma.power(10., 0.1*zdrdB)
+
+    mask = np.ma.getmaskarray(rhohv)
+    fill_value = rhohv.get_fill_value()
```
jjhelmus: Same as above.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,200 @@ def calculate_snr_from_reflectivity(
...
+
+    cdr_data = (
+        10.*np.ma.log10(
+            (1.+1./zdr-2.*rhohv*np.ma.sqrt(1./zdr)) /
+            (1.+1./zdr+2.*rhohv*np.ma.sqrt(1./zdr))))
+    cdr_data.set_fill_value(fill_value)
+    cdr_data.data[mask.nonzero()] = fill_value
```
jjhelmus: `cdr_data[mask] = np.ma.masked`

pyart/util/xsect.py
```diff
@@ -41,64 +46,176 @@ def cross_section_ppi(radar, target_azimuths):
...
-            prhi_rays.append(ray_number + sweep_slice.start)
+            d_az = np.abs(sweep_azimuths - target_azimuth)
+            if az_tol is None:
+                ray_number = np.argmin(d_az)
+                prhi_rays.append(ray_number + sweep_slice.start)
+                valid_azimuths.append(target_azimuth)
+            else:
+                d_az_min = np.min(d_az)
+                if d_az_min > az_tol:
+                    print('WARNING: No azimuth found whithin tolerance ' +
```
jjhelmus: This should use Python's built in `warnings` module so that this can be logged or caught by the user.  `warnings.warn` is the function to use.

pyart/util/xsect.py
```diff
@@ -41,64 +46,176 @@ def cross_section_ppi(radar, target_azimuths):
...
+                          str(d_az_min)+' larger than tolerance '+
+                          str(az_tol))
+                else:
+                    ray_number = np.argmin(d_az)
+                    prhi_rays.append(ray_number + sweep_slice.start)
+                    valid_azimuths.append(target_azimuth)
+
+    rhi_nsweeps = len(valid_azimuths)
+    if rhi_nsweeps == 0:
+        print('WARNING: No azimuth found whithin tolerance.')
```
jjhelmus: Perhaps this should raise an Error instead of returning None.

pyart/util/xsect.py
```diff
@@ -41,64 +46,176 @@ def cross_section_ppi(radar, target_azimuths):
...
+            sweep_elevations = radar.elevation['data'][sweep_slice]
+            d_el = np.abs(sweep_elevations - target_elevation)
+            if el_tol is None:
+                ray_number = np.argmin(d_el)
+                pppi_rays.append(ray_number + sweep_slice.start)
+                valid_elevations.append(target_elevation)
+            else:
+                d_el_min = np.min(d_el)
+                if d_el_min > el_tol:
+                    print('WARNING: No elevation found whithin tolerance ' +
```
jjhelmus: same as above

pyart/util/xsect.py
```diff
@@ -41,64 +46,176 @@ def cross_section_ppi(radar, target_azimuths):
...
+                          str(d_el_min)+' larger than tolerance '+
+                          str(el_tol))
+                else:
+                    ray_number = np.argmin(d_el)
+                    pppi_rays.append(ray_number + sweep_slice.start)
+                    valid_elevations.append(target_elevation)
+
+    ppi_nsweeps = len(valid_elevations)
+    if ppi_nsweeps == 0:
+        print('WARNING: No elevation found whithin tolerance.')
```
jjhelmus: same as above

jjhelmus:
@jfigui Added a few new comments.  Most of them are about setting `fill_value` and the `data` attribute of masked arrays.  

In general one should not depend on the `fill_value` of a masked array being correct as this value is not guaranteed to propagate.  

Additionally setting the `data` value for elements of a masked array which are masked in pointless.  NumPy assumes that these values can take on arbitrary values since the values are masked and does not attempt to preserve these values when operations are performed on the masked array.

Also should mention that it best to separate the addition of new functionality into different PR.  Each PR should only add one new feature or fix one particular bug.  This makes reviewing and merging easier.  Happy to get this one through as-is just something to keep in mind for the future.

meteoswiss-mdr:
@jjhelmus,
I checked the issue and the fill_value is not propagated (it takes the default 1e+20 for float arrays). So I think we should explicitly set the fill_value. There may be contexts where it can get handy to know the fill_value directly from the data. What do you think?
The rest of your comments I have taken them on board. I agree with you that as a general rule each functionality should have a different PR but it was such a minor addition that I though there was no need to bother.
I will commit the changes in a few minutes.

Greetings from Switzerland,

Jordi

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,190 @@ def calculate_snr_from_reflectivity(
...
+    """
+    # parse the field parameters
+    if noise_field is None:
+        noise_field = get_field_name('noisedBZ_hh')
+
+    noisedBZ_vec = noisedBZ_val+20.*np.ma.log10(1e-3*range/ref_dist)
+
+    noisedBZ = get_metadata(noise_field)
+    noisedBZ['data'] = np.tile(noisedBZ_vec, (nrays, 1))
+#    noisedBZ['data'].set_fill_value(get_fillvalue())
```
jjhelmus: Can you remove this line, dead code is not good to include.

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,190 @@ def calculate_snr_from_reflectivity(
...
+        refl = radar.fields[refl_field]['data']
+    else:
+        raise KeyError('Field not available: ' + refl_field)
+    if noise_field in radar.fields:
+        noisedBZ = radar.fields[noise_field]['data']
+    else:
+        raise KeyError('Field not available: ' + noise_field)
+
+    snr_data = refl-noisedBZ
+#    snr_data.set_fill_value(get_fillvalue())
```
jjhelmus: Remove

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,190 @@ def calculate_snr_from_reflectivity(
...
+    if l_field is None:
+        l_field = get_field_name('logarithmic_cross_correlation_ratio')
+
+    # extract rhohv field from radar
+    if rhohv_field in radar.fields:
+        rhohv = radar.fields[rhohv_field]['data']
+    else:
+        raise KeyError('Field not available: ' + rhohv_field)
+
+#    mask = np.ma.getmaskarray(rhohv)
```
jjhelmus: Remove

pyart/retrieve/simple_moment_calculations.py
```diff
@@ -66,3 +70,190 @@ def calculate_snr_from_reflectivity(
...
+
+    rhohv[rhohv >= 1.] = 0.9999
+    l_data = -np.ma.log10(1.-rhohv)
+#    l_data.data[mask] = np.ma.masked
+#    l_data.set_fill_value(get_fillvalue())
```
jjhelmus: Remove

jjhelmus:
@jfigui I'm going to remove the commented out lines of code and merge this, thanks for putting this together and seeing it through.  The CI failures are unrelated to this change and I'll investigate shortly.

jjhelmus:
> I checked the issue and the fill_value is not propagated (it takes the default 1e+20 for float arrays). So I think we should explicitly set the fill_value. There may be contexts where it can get handy to know the fill_value directly from the data. What do you think?

The issue I was referring to is that the fill_value is not propogated when masked arrays are operated on.  For example, see how the fill_value of the `output` variable varies below.

```
import numpy as np
a = np.ma.array([1,2,3], mask=[False, False, True], fill_value=999)
b = np.ma.array([1,2,3], mask=[False, False, True], fill_value=888)
c = np.ma.array([1,2,3])

output = a + b  # output.fill_value is 999
output = b + a  # output.fill_value is 888
output = a + c  # output.fill_value is 999
output = c + a  # output.fill_value is 999999 (sometimes...)
```

I agree that setting the `fill_value` of a masked array is prudent, but I would not rely on it being correctly set when retrieving the value.  Using `np.ma.filled` with an explicit `fill_value` parameter is a more reliable method for creating a ndarray with a particular value where the original data is masked.

meteoswiss-mdr:
@jjhelmus, thank you for your patience.

Now that I have a better understanding of masked arrays I agree with you. There is no use on setting the fill_value and the data value of masked entries and whenever needed filled should be used.

Have a nice day,


Jordi

PR #585, opened 2016-08-29
BUG: bugfix for reading NEXRAD L3 product 135 (EET)

jjhelmus:
Fixes a bug when reading NEXRAD Level 3 produce 135 (Enhanced Echo  Tops) files
which prevented the data from being properly read.


jjhelmus:
Travis CI failure is un-related to this change, will address separately.

PR #588, opened 2016-09-07
Addition of two KDP retrieval methods

Wolfigo:
Dear all,

As my lab's first contribution to the pyart project I would like to suggest adding two methods for KDP estimation, to the _retrieve_ module. I have added these methods in the suggested _kdp_proc.py_ file. The two methods are 

_kdp_vulpiani_ : a simple method for estimating kdp with moving windows (Vulpiani and al., 2012)

_kdp_schneebeli_ : the ensemble Kalman filter method by Schneebeli and al. (2014).

One advantage of these methods over the Maesaka method is that they can retrieve negative KDP. These two methods have very similar calling procedures to the already implement method _kdp_maesaka_. Another addition is the function _filter_psidp_ which can be optionally used prior to estimating Kdp to remove short subsequences of psidp, unravel angles with large discontinuities, use a median filter and mask psidp when rhohv is smaller than a given a threshold.

I provide a comparison of the three methods in the _tests/compare_kdp_proc.py_ file for a real profile of measured psidp (on an X-band radar) as well as a comparison for one PPI (see below).

[comparison_kdp_single_profile.pdf](https://github.com/ARM-DOE/pyart/files/459307/comparison_kdp_single_profile.pdf)
ex_kdp_ppi

Best regards,

Daniel Wolfensberger
EPFL-LTE


hhuangmeso:
good work

kirknorth:
Awesome work, @Wolfigo. I'll have to read over that Schneebeli paper describing the Kalman filter approach since that sounds very cool. This also may have been the same idea @swnesbitt was describing or at least going for in PR #508. 

Being the developer of `kdp_maesaka` I thought I'd just point out that the *streaks* in some of the rays are due to the *soft* boundary condition problem that still needs to be addressed.  While it's not a huge issue right now as it only affects a small percentage of rays, it's something I'd still like to fix.

scollis:
Seconding @kirknorth ! Thanks for the PR.. Most welcome 
Seems you have added unit tests too. This is great. We will need to take a look at why they are failing but I am sure we will sort this out. We may be a little time accepting this. 

Folks are going to have no excuse for not using KDP 👍

jjhelmus:
@Wolfigo Thanks for the contribution.  I don't have the time right now to give this a full review but hope to get to it tomorrow (Friday) or early next week.  

One suggestion I have from a quick look at the code is to look at the [PEP 8 -- Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/).  The code in Py-ART follows this style guide quite stringently.  The [pycodestyle](https://pypi.python.org/pypi/pycodestyle) tool will check your code against this style convention and report location which are in conflict with the conventions.

jjhelmus:
@Wolfigo The last few commit you added to this PR do not seem related to the KDP calculation and are a bit odd (especially the `wrapper` files which looks to be a html file).  Are these commit intended to be part of this PR?  Using a [feature branch](https://guides.github.com/introduction/flow/) for the KDP related changes rather than your `master` branch might be a good idea and would allow us to review this work while you continue to work on additional changes to Py-ART.  Please feel free to reach out if you need any assistance with this.

jfigui:
@Wolfigo I am very happy you became an active member of Py-ART! Welcome!

kirknorth:
@Wolfigo, please be aware that once you initiate a PR from your development branch (in this case `master`), any additional commits you make locally on that branch and push to your remote (e.g., `origin`) will automatically get pushed `upstream` as well. I believe, as @jjhelmus has pointed out, that these additional commits you've pushed recently are unrelated to your original PR which had to do with two new KDP retrieval methods.

scollis:
Yep..

Everyone needs to learn how to use Git Branches.. They seem kind of useless until the don’t and then you realize they are completely needed! 

I think Jonathan covers it in his intermediate git presentation here:
https://github.com/scollis/SusSoPrac/tree/master/presentations <https://github.com/scollis/SusSoPrac/tree/master/presentations>

Again, I am annoyed I will not be at ERAD as this could be something we could cover in the short course.. 

Dr Scott Collis
Argonne National Laboratory
Mb: +1 630 235 8025
Of: +1 630 252 0550

jjhelmus:
closing in favor of #593

PR #593, opened 2016-09-14
Addition of two KDP method (Corrected version)

wolfidan:
_Please accept my apologies for messing up with the PR, I know I should have used a branch. Here is a corrected version, with only the necessary commits_

Dear all,

As my lab's first contribution to the pyart project I would like to suggest adding two methods for KDP estimation, to the retrieve module. I have added these methods in the suggested kdp_proc.py file. The two methods are

kdp_vulpiani : a simple method for estimating kdp with moving windows (Vulpiani and al., 2012)

kdp_schneebeli : the ensemble Kalman filter method by Schneebeli and al. (2014).

One advantage of these methods over the Maesaka method is that they can retrieve negative KDP. These two methods have very similar calling procedures to the already implement method kdp_maesaka. Another addition is the function filter_psidp which can be optionally used prior to estimating Kdp to remove short subsequences of psidp, unravel angles with large discontinuities, use a median filter and mask psidp when rhohv is smaller than a given a threshold.

I provide a comparison of the three methods in the tests/compare_kdp_proc.py file for a real profile of measured psidp (on an X-band radar) as well as a comparison for one PPI (see below).

[comparison_kdp_single_profile.pdf](https://github.com/ARM-DOE/pyart/files/471734/comparison_kdp_single_profile.pdf)

19737f5a-7514-11e6-9b53-f5a5a6fa891a

### DIFF
--- a/pyart/aux_io/radx.py
+++ b/pyart/aux_io/radx.py
@@ -19,7 +19,7 @@ from ..io.cfradial import read_cfradial
 from ..io.common import _test_arguments
 
 
-def read_radx(filename, **kwargs):
+def read_radx(filename, radx_dir=None, **kwargs):
     """
     Read a file by first converting it to Cf/Radial using RadxConvert.
 
@@ -27,6 +27,9 @@ def read_radx(filename, **kwargs):
     ----------
     filename : str
         Name of file to read using RadxConvert.
+
+    radx_dir : str, optional
+        path to the radx install
 
     Returns
     -------
@@ -36,12 +39,16 @@ def read_radx(filename, **kwargs):
     """
     # test for non empty kwargs
     _test_arguments(kwargs)
+    if radx_dir is not None:
+        executable = os.path.join(radx_dir, 'RadxConvert')
+    else:
+        executable = 'RadxConvert'
 
     tmpfile = tempfile.mkstemp(suffix='.nc', dir='.')[1]
     head, tail = os.path.split(tmpfile)
     try:
         subprocess.check_call(
-            ['RadxConvert', '-const_ngates',
+            [executable, '-const_ngates',
              '-outdir', head, '-outname', tail, '-f', filename])
         if not os.path.isfile(tmpfile):
             raise IOError(


Wolfigo:
I will be abroad for a week, but I will make sure to correct the code according to your remarks when I am back

pyart/retrieve/kdp_proc.py
```diff
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
```
jjhelmus: These lines are not used in other pyart modules and should be removed.

pyart/retrieve/kdp_proc.py
```diff
@@ -21,12 +33,978 @@
...
+        specified in the Py-ART configuration file.
+    kdp_field : str, optional
+        Specific differential phase field. If None, the default field name must
+        be specified in the Py-ART configuration file.
+    phidp_field : str, optional
+        Propagation differential phase field. If None, the default field name
+        must be specified in the Py-ART configuration file.
+    band : char, optional
+       Radar frequency band string. Accepted "X", "C", "S" (capital
+       or not). IT is used to compute Rcov and Pcov when not provided
```
jjhelmus: It

pyart/retrieve/kdp_proc.py
```diff
@@ -21,12 +33,978 @@
...
+    
+    all_psidp_prof = list(psidp_o)
+    
+    if parallel:
+        list_est = pool.map(func,all_psidp_prof)
+    else:
+        list_est = map(func,all_psidp_prof)    
+        
+    kdp = np.zeros(psidp_o.shape)*np.nan
+    kdp = np.ma.masked_array(kdp, fill_value = fill_value)
```
jjhelmus: Is the goal here to create a masked array which is initially completely masked?  If so that is NOT what is being done, and filling a masked array with NaNs is dangerous.  NaNs in masked array start as not masked but can become masked later on if certain checks are triggered.  

A better method to create a fully masked array is:
``` Python
kdp = np.ma.array(np.zeros(phidp_o.shape), mask=True, fill_value=fill_value)
```

pyart/retrieve/kdp_proc.py
```diff
@@ -21,12 +33,978 @@
...
+        estimated error on Kdp values
+
+    """
+   
+    #--------------------------------------------------------
+    #Define the input
+    psidp = psidp_in
+   
+    #Get indices of finite data
+    real_data_ind = np.where(np.isfinite(psidp))[0]
```
jjhelmus: remove?

pyart/retrieve/kdp_proc.py
```diff
@@ -21,12 +33,978 @@
...
+           10**(-0.2), 10**(-0.4), 10**(-0.6), 10**(-0.8),1, 10]
+           
+from scipy import optimize, stats, interpolate, linalg, signal
 
 from . import _kdp_proc
 from ..config import get_field_name, get_metadata, get_fillvalue
 
+def kdp_schneebeli(radar, gatefilter=None, fill_value = None, psidp_field=None,
+                kdp_field=None, phidp_field=None, band = 'C', rcov = 0, pcov = 0,
+                prefilter_psidp = False, filter_opt = {}, parallel = True):
```
jjhelmus: The default value of filter_opt should be `None`.  Using a mutable default type [is dangerous](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

jjhelmus:
@Wolfigo Before I review this can you change the style of the code so that it meets the [PEP8 conventions](https://www.python.org/dev/peps/pep-0008/).  The [pycodestyle](https://pypi.python.org/pypi/pycodestyle) tool will provide you a list of violations of these conventions.

pyart/retrieve/tests/test_kdp_proc.py
```diff
@@ -7,7 +7,24 @@
 from pyart.testing import sample_objects
```
jjhelmus: This file should be mode 644 not 755.  The new mode results in the tests in this file being overlooked in some cases.

pyart/retrieve/tests/test_kdp_proc.py
```diff
@@ -7,7 +7,24 @@
...
 from pyart.config import get_field_name
 
+def test_kdp_vulpiani_linear_psidp(slope=0.002):
+    radar = _make_linear_psidp_radar(slope=slope)
+    kdp_dict, phidp_dict = kdp_proc.kdp_vulpiani(radar,windsize=10)
+    valid_kdp = kdp_dict['data'][0][np.isfinite(kdp_dict['data'][0])]
+    assert np.allclose(np.diff(valid_kdp), 0.0, atol=0.1)
```
jjhelmus: The first five values in the first ray of kdp_dict['data'] are -- (masked) 0.001, 0.001, 0.001 and 0.001 which seem incorrect.  Please investigate.

pyart/retrieve/tests/test_kdp_proc.py
```diff
@@ -7,7 +7,24 @@
...
+    assert np.allclose(valid_kdp, 1000.0 * slope / 2.0, atol=0.1)
+
+    return
 
+def test_kdp_kalman_linear_psidp(slope=0.002):
+    radar = _make_linear_psidp_radar(slope=slope)
+    kdp_dict, kdp_std_dict, phidp_dict = kdp_proc.kdp_schneebeli(radar,band='X')
+    valid_kdp = kdp_dict['data'][0][np.isfinite(kdp_dict['data'][0])]
```
jjhelmus: The kdp value here seem very unusual.  Please investigate.

PR #607, opened 2016-10-14
BUG: Read in altitude form ARM sonde files

### LINKED ISSUE #606: Altitude data available from ARM sonde data stream
I'm working on some VAD code and came across this potential issue. Altitude information is available in ARM sonde data streams, yet heights are computed from pressure measurements in [line 59](https://github.com/ARM-DOE/pyart/blob/master/pyart/io/arm_sonde.py#L59) of `read_arm_sonde`. Just curious why this is.  Here's a plot of the differences between the two for a test file [here](http://www.meteo.mcgill.ca/~kirk/). 

height_difference

According to the variable attributes, sonde altitude measurements should be in meters above mean sea level, which matches what's written in the docs of `HorizontalWindProfile` class.

jjhelmus:
Read in the altitude directly from ARM sonde files from the alt variable.
Previously the altitude/height was calculated from the pressure which would
give slightly incorrect values.

closes #606

### DIFF
--- a/pyart/io/arm_sonde.py
+++ b/pyart/io/arm_sonde.py
@@ -40,24 +40,7 @@ def read_arm_sonde(filename):
     profile_datetime = netCDF4.num2date(
         dset.variables['time'][0], dset.variables['time'].units)
 
-    # extract wind profile
-    pressure = dset.variables['pres'][:]
-    # pressure to altitude:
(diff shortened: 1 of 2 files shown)


jjhelmus:
Failing CI is due to ContinuumIO/anaconda-issues/#1158 and not related to the changes at hand.  Merging.

PR #608, opened 2016-10-14
Added a CfRadial writer for SINARAME H5 files

### LINKED ISSUE #600: Reading Argentinian H5 files and saving in CfRadial
Hi everyone,

this is not an actual issue of PyART but an enhancement proposal separated in two parts:

I updated the sinarame_h5 reader to match the odim_h5 reader. It's mostly the same but there are some 'unique' features from the h5 files. I will soon submit a PR with this stuff.

The second thing is something @swnesbitt tried to pull in #480. Writing cfradial from the h5 files is something I found really useful and I know some people who can use this as well.

This particular issue I found was on the way the original bufr files are created:  the variables are in separated files, they have different gate bins, strategies applied, etc. It's a huge mess to work with this data as is. I know this is a _very_ specific case, but I came up with a script (adapting something Steve has done) that saves all the variables and their metadata in one cfradial file. Should this be added as another function in the `sinarame_h5` file? Or, as discussed in the PR, it will fit better as an example or in the wiki?

I'm polishing that script (because it's in spanish right now) and I'll upload it a gist so you have a better idea of what I'm talking about. In the meantime let me know what you think of this addition.

mrugna:
I believe I didn't forget anything for the dev documentation. Let me know if something's wrong with the addition.

This fully closes #600.

### DIFF
--- a/pyart/aux_io/__init__.py
+++ b/pyart/aux_io/__init__.py
@@ -43,6 +43,7 @@ from .edge_netcdf import read_edge_netcdf
 from .odim_h5 import read_odim_h5
 from .gamic_hdf5 import read_gamic
 from .sinarame_h5 import read_sinarame_h5
+from .sinarame_h5 import write_sinarame_cfradial
 from .rainbow_wrl import read_rainbow_wrl
 
 __all__ = [s for s in dir() if not s.startswith('_')]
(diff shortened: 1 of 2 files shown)


jjhelmus:
LGMT, failing CI is unrelated.  Thanks for the contribution @mrugna!

PR #610, opened 2016-10-17
BUG: Read rainbow files with only starting angle

### LINKED ISSUE #609: reading rainbow version 5.43.10 data
I've been able to read some version 5.34.16 rainbow files but am unable to do so for a more current rainbow version (5.43.10).  Was wondering if support for these files will be included at some future date.  I've attached an example of the newer version.  

[2015111700500600dBZ.zip](https://github.com/ARM-DOE/pyart/files/530587/2015111700500600dBZ.zip)

Here's the error I get:

In [25]: file = '/home/disk/bob/olympex/xband/rainbowFiles/RAW_DOPVOL1_A/DBZ/2015111700500600dBZ.azi'

In [26]: radar = pyart.aux_io.read_rainbow_wrl(file)                                                 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-26-6b5fa6247414> in <module>()
----> 1 radar = pyart.aux_io.read_rainbow_wrl(file)

/home/disk/shear2/brodzik/.local/lib/python2.7/site-packages/pyart/aux_io/rainbow_wrl.pyc in read_rainbow_wrl(filename, field_names, additional_metadata, file_field_names, exclude_fields, **kwargs)
    297         moving_angle[ssri[i]: seri[i]+1], angle_start, angle_stop = (
    298             _get_angle(slice_info['slicedata']['rayinfo'],
--> 299                        angle_step=angle_step, scan_type=scan_type))
    300 
    301         # time

/home/disk/shear2/brodzik/.local/lib/python2.7/site-packages/pyart/aux_io/rainbow_wrl.pyc in _get_angle(ray_info, angle_step, scan_type)
    385         if angle_step is None:
    386             raise ValueError('Unknown angle step')
--> 387         angle_start = np.array(ray_info['data']*bin_to_deg, dtype='float64')
    388         if scan_type == 'rhi':
    389             ind = (angle_start > 225.).nonzero()

TypeError: list indices must be integers, not str

jjhelmus:
Fix issue in reading rainbow files where only the starting angle is provided in
the file.

closes #609

### DIFF
--- a/pyart/aux_io/rainbow_wrl.py
+++ b/pyart/aux_io/rainbow_wrl.py
@@ -384,7 +384,7 @@ def _get_angle(ray_info, angle_step=None, scan_type='ppi'):
     else:
         if angle_step is None:
             raise ValueError('Unknown angle step')
-        angle_start = np.array(ray_info['data']*bin_to_deg, dtype='float64')
+        angle_start = np.array(ray_info[0]['data']*bin_to_deg, dtype='float64')
         if scan_type == 'rhi':
             ind = (angle_start > 225.).nonzero()
             angle_start[ind] -= 360.


jjhelmus:
@jfigui Does this change seem reasonable?  You wrote this function so I wanted to check before merging this change.

meteoswiss-mdr:
@jjhelmus ,

Are you sure about that? When I tested it with my own files the wradlib reader would return a 1 dimension array with the angle data when only angle start was there. Only when both angle start and angle stop where there would return a two dimensional array.

I think that the original code works all right. Have you tested the new one with one of the rainbow files with version 5.22.3 that I gave you?

meteoswiss-mdr:
By the way, my nickname has changed. From now one please refer to me as meteoswiss-mdr. This is the nickname from the Radar, Satellite and Nowcasting division at MeteoSwiss.

kmuehlbauer:
@jjhelmus @meteoswiss-mdr 

The problem has to do with indexing and the @meteoswiss-mdr code is missing something. 

The thing is, that the `rayinfo` can hold several types of info. Normally this is just the `startangle`. In some systems the `stopangle` is added. The provided data from @srbrodzik contains 6 of them (`startangle`, `stopangle`, 'timestamp`, `txpower`, `noisepowerh` and `noisepowerv`). 

As the current code just checks for two elements (`startangle`, `stopangle`) it wrongly assumes just one element (because there are 6) and is trying to read it. The proposed change by @jjhelmus assumes that the `startangle` is always in the first element. But this will let the `stopangle` unacknowledged, although it is there. And it will break the reader if the `rayinfo` contains only the `startangle`.

So a more sophisticated check is needed.

Just as a side note for devs and users. For getting a quick overview of the layout of a deeply nested python dictionary (as the wradlib rainbow reader is putting out) I recommend using `traitsui`:

http://wradlib.org/2016/03/treeview-with-traits/

pyart/aux_io/rainbow_wrl.py
```diff
@@ -384,7 +384,7 @@ def _get_angle(ray_info, angle_step=None, scan_type='ppi'):
     else:
         if angle_step is None:
             raise ValueError('Unknown angle step')
-        angle_start = np.array(ray_info['data']*bin_to_deg, dtype='float64')
+        angle_start = np.array(ray_info[0]['data']*bin_to_deg, dtype='float64')
```
kmuehlbauer: The `else` path should only be taken when the `ray_info` is just a dict (consisting of blobid, depth, rays, refid).

The check `if (len(ray_info) == 2):` (Line 375) works only if the `ray_info` contains only `startangle` and `stopangle`.

So, we should check if `ray_info` is a list first. If not we can take the `else` path. If it is a list we iterate over the elements and check `rayinfo[iterator]['@refid']` for it's contents (eg `startangle`, `stopangle`). Then we fill the found data into the right arrays. Makes sense?

meteoswiss-mdr: @kmuehlbauer 
Thanks for clarifying the problem. I see the issue now. You have a newest version of the Rainbow file. I would love to have this information in my own files! :).
I think the best approach is to have a chain of elif:

if (len(ray_info) == 1):
    treatment of case with angle_start only
elif (len(ray_info)==2):
    treatnment of case with angle_start and angle_stop
else:
   treatment of case with all the additional information

I assume that the ray information is given in an specified order so we can keep using indices to access the information in ray_info anyway.

Does this approach make sense to you?

kmuehlbauer:
@meteoswiss-mdr 
Checking for actual content (and not relying on index) is better IMHO. But we can extract the `refid` fields before and iterate over the known ones (which we can declare as default fields). If some unknown `refid` is found, we can raise a warning and make the user report that back to pyart-dev. This is a more elegant way to be prepared for future additions or changes in element order.

meteoswiss-mdr:
@kmuehlbauer ,
You are right that checking on the content is a better policy but I would have to think about how to implement it. I am quite busy at the moment but feel free to pull a solution.

kmuehlbauer:
@jjhelmus 

One quick solution resolving only the angle-stuff is to add the following lines:

```python
    def _extract_angles(data):
        angle = np.array(data, dtype='float64')
        if scan_type == 'rhi':
            ind = (angle > 225.).nonzero()
            angle[ind] -= 360.
        return angle

    try:
        angle_start = _extract_angles(ray_info['data'] * bin_to_deg)
        if angle_step is None:
            raise ValueError('Unknown angle step')
        angle_stop = angle_start + angle_step
    except TypeError:
        angle_start = _extract_angles(ray_info[0]['data'] * bin_to_deg)
        angle_stop = _extract_angles(ray_info[1]['data'] * bin_to_deg)
```

instead of these lines:

```python
    if (len(ray_info) == 2):
        angle_start = np.array(ray_info[0]['data']*bin_to_deg, dtype='float64')
        if scan_type == 'rhi':
            ind = (angle_start > 225.).nonzero()
            angle_start[ind] -= 360.
        angle_stop = np.array(ray_info[1]['data']*bin_to_deg, dtype='float64')
        if scan_type == 'rhi':
            ind = (angle_stop > 225.).nonzero()
            angle_stop[ind] -= 360.
    else:
        if angle_step is None:
            raise ValueError('Unknown angle step')
        angle_start = np.array(ray_info['data']*bin_to_deg, dtype='float64')
        if scan_type == 'rhi':
            ind = (angle_start > 225.).nonzero()
            angle_start[ind] -= 360.
        angle_stop = angle_start+angle_step
```
This is based on the same assumptions as the current code, but will not break if there are more ray_info elements. 

If we think about retrieving also the other ray_info-elements, we might add some more generic getter-function.

srbrodzik:
I like this proposed change because the code won't break if the definition of ray_info changes.  It probably would be nice to grab all the other ray_info elements but this mod allows me to look at the data I have.  Thanks.

meteoswiss-mdr:
Very elegant solution!

@srbrodzik , your angle information is in the first two elements of ray_info right?

@jjhelmus, I think you can implement this solution for now until we get something smarter.

kmuehlbauer:
@meteoswiss-mdr, @jjhelmus, @srbrodzik 

I opened a PR with the proposed change (#613).

jjhelmus:
closing in favor of #613

PR #611, opened 2016-10-17
CI: Supress git commit output after building docs

jjhelmus:


jjhelmus:
Change required to shorten CI log so that the build passes.  For example see [this build](https://travis-ci.org/ARM-DOE/pyart/builds/168376152).

Issue #483, opened 2016-02-10
support for cartopy for mapping


swnesbitt:
I am increasingly favoring Cartopy as a mapping package over Basemap.  I find it to be much higher performance, has more meteorological community backing, and has lots of built in GIS capabilities, as well as integration with web mapping services like google and NASA.

From an architecture standpoint, I am wondering what the best way to implement this would be.  It would be nice to be able to select the mapping backend in functions like `plot_ppi_map` through a call to a `map_backend` keyword, and then allow a dictionary to be passed that could control the functionality of the mapping, and return handles to the plotting so that further cartopy functionality could be applied (i.e. overlay of shapefiles).

I am willing to take a crack at this, just interested in some general feedback on the idea.

scollis:
Completely concur. and yes. I do think having a backend keyword would be the best way to implement this. To start with Matplotlib will be the default but if cartopy becomes more fully featured it can eventually switch. An error should be raised if the requisite package does not exist. ie cartopy will not be a compulsory dependency for Py-ART

jjhelmus:
:+1: Having support for plotting on maps with Cartopy would be great.
 
In terms for implementation, if I was going to work on this I would begin by creating a separate class which uses Cartopy for mapping say **RadarMapDisplayCartopy**.  Get that working first and then try to merge the two classes with the backend being an optional argument.  I think this approach would be a bit easier than trying to work out all the details in the **RadarMapDisplay** class from the get go.

scollis:
We discussed this at the NCAR SEA.. Our discussion was prompted by MetPy support for Cartopy and station plots.. 
One issue is that Cartopy is not yet in the main channel of Conda. While we have some dependencies that are not in conda they are very optional and in direct support to ARM programs (eg CyLP). 

I will contact the Cartopy leads and check on their plans and if they intend to get a Conda package up.. If the answer is yes then we can think about adding this to the roadmap (increasing chances for DoE support through Argonne). Otherwise this remains in the domain of a user contribution and will need a conditional import.

jjhelmus:
There are a number of cartopy conda packages on Anaconda.org, including the ones in the [conda-forge](https://anaconda.org/conda-forge/cartopy) channel which are likely the "best".  The concern is that cartopy is not in the default channel nor installed in the anaconda meta-package.  Both of these are managed by Continuum and getting cartopy included in either would have to have a change on their end.

scollis:
So,
[...]

PR #612, opened 2016-10-18
support for cartopy for mapping

jsignell:
After the discussion in #483.


pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        if not _CARTOPY_AVAILABLE:
+            raise MissingOptionalDependency(
+                "Cartopy is required to use RadarMapDisplayCartopy but is " +
+                "not installed")
+
+        # initalize the base class
+        RadarDisplay.__init__(self, radar, shift=shift)
+
+        # additional attributes needed for plotting on a basemap.
+        self.basemap = None
```
jjhelmus: Can we change the name of this attribute.  It does not refer to a Basemap instance so the name is poor.  Perhaps crs since this will be a Cartopy coordinate reference system?

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            central_longitude=lon_0, central_latitude=lat_0,
+            standard_parallels=(lat_0, ))
+        self.proj = proj
+
+        # create the basemap if not provided
+        import matplotlib.pyplot as plt
+
+        if hasattr(ax, 'projection'):
+            basemap = ax
+        elif hasattr(ax, 'projection'):
```
jjhelmus: What exactly is this second `hasattr` statement doing?  It seems like this would always be False

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            field, sweep, mask_tuple, filter_transitions, gatefilter)
+        x, y = self._get_x_y(sweep, edges, filter_transitions)
+
+        # mask the data where outside the limits
+        if mask_outside:
+            data = np.ma.masked_outside(data, vmin, vmax)
+
+        proj = cartopy.crs.LambertConformal(
+            central_longitude=lon_0, central_latitude=lat_0,
+            standard_parallels=(lat_0, ))
```
jjhelmus: The creator or proj should be delayed until needed if no projection argument is provided.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        self.proj = proj
```
jjhelmus: This can be removed.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        # create the basemap if not provided
+        import matplotlib.pyplot as plt
+
+        if hasattr(ax, 'projection'):
+            basemap = ax
+        elif hasattr(ax, 'projection'):
+            basemap = basemap
+        else:
+            if projection is None:
+                basemap = plt.axes(projection=proj)
```
jjhelmus: Create the proj object in this if block

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        ax : Axis
+            Axis to plot on. If provided, must have a Cartopy crs projection.
+        fig : Figure
+            Figure to add the colorbar to. None will use the current figure.
+        lat_lines, lon_lines : array or None
+            Locations at which to draw latitude and longitude lines.
+            None will use default values which are resonable for maps of
+            North America.
+        projection : cartopy.crs class
+            Map projection supported by cartopy.
```
jjhelmus: Add text indicating that the default is a Lambert conformal projection centered at lat_0.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+                               crs=cartopy.crs.PlateCarree())
+
+        # add embelishments
+        if embelish is True:
+            # shpfile = cartopy.io.shapereader.gshhs(resolution)
+            # shp = cartopy.io.shapereader.Reader(shpfile)
+            # basemap.add_geometries(
+            #    shp.geometries(), cartopy.crs.PlateCarree(), edgecolor='black',
+            #    facecolor='none')
+            #basemap.drawstates()
```
jjhelmus: Remove dead lines

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        edges : bool
+            True will interpolate and extrapolate the gate edges from the
+            range, azimuth and elevations in the radar, treating these
+            as specifying the center of each gate.  False treats these
+            coordinates themselved as the gate edges, resulting in a plot
+            in which the last gate in each ray and the entire last ray are not
+            not plotted.
+        embelish: bool
+            True by default. Set to false to supress drawing of coastlines
+            etc.. Use for speedup when specifying shapefiles.
```
jjhelmus: This should be worded to indicate that the embelishment is only added when the projection allows it.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            #    facecolor='none')
+            #basemap.drawstates()
+            if basemap.projection in [cartopy.crs.PlateCarree(),
+                                      cartopy.crs.Mercator()]:
+                gl = basemap.gridlines(xlocs=lon_lines, ylocs=lat_lines,
+                                       draw_labels=True)
+                gl.xlabels_top = False
+                gl.ylabels_right = False
+
+            elif basemap.projection == proj:
```
jjhelmus: check against crs.LambertConformal here

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            basemap.coastlines(resolution=resolution)
+            basemap.add_feature(states_provinces, edgecolor='gray')
+
+        # plot the data and optionally the shape file
+        # we need to convert the radar gate locations (x and y) which are in
+        # km to meters as well as add the map coordiate radar location
+        # which is given by self._x0, self._y0.
+        pm = basemap.pcolormesh(
+            x * 1000., y * 1000., data,
+            vmin=vmin, vmax=vmax, cmap=cmap, norm=norm, transform=proj)
```
jjhelmus: I think transform=self.basemap (or the new attribute name) is needed here.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        if 'ax' in kwargs:
+            kwargs.pop('ax')
+        self._check_basemap()
+        angle = np.linspace(0., 2.0 * np.pi, npts)
+        xpts = range_ring_location_km * 1000. * np.sin(angle)
+        ypts = range_ring_location_km * 1000. * np.cos(angle)
+        self.plot_line_xy(xpts, ypts, line_style=line_style, **kwargs)
+
+# Thses methods are a hack to allow gridlines when the projection is lambert
+# http://nbviewer.jupyter.org/gist/ajdawson/dd536f786741e987ae4e
```
jjhelmus: Do we have permission from the original author to include these?  Would they allow this work to be BSD licensed?

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+def find_side(ls, side):
+    """
+    Given a shapely LineString which is assumed to be rectangular, return the
+    line corresponding to a given side of the rectangle.
+
+    """
+    import shapely.geometry as sgeom
```
jjhelmus: This should be moved to the top, although also must be allowed to fail.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+    yticks, yticklabels = _lambert_ticks(ax, ticks, 'left', lc, te)
+    ax.yaxis.tick_left()
+    ax.set_yticks(yticks)
+    ax.set_yticklabels([ax.yaxis.get_major_formatter()(ytick) for
+                        ytick in yticklabels])
+
+def _lambert_ticks(ax, ticks, tick_location, line_constructor, tick_extractor):
+    """Get the tick locations and labels for a Lambert Conformal projection."""
+    import shapely.geometry as sgeom
+    from copy import copy
```
jjhelmus: Move to top, and allow to fail

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+        if mask_outside:
+            data = np.ma.masked_outside(data, vmin, vmax)
+
+        proj = cartopy.crs.LambertConformal(
+            central_longitude=lon_0, central_latitude=lat_0,
+            standard_parallels=(lat_0, ))
+        self.proj = proj
+
+        # create the basemap if not provided
+        import matplotlib.pyplot as plt
```
jjhelmus: Move to top of module

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            Axis to plot on. If provided, must have a Cartopy crs projection.
+        fig : Figure
+            Figure to add the colorbar to. None will use the current figure.
+        lat_lines, lon_lines : array or None
+            Locations at which to draw latitude and longitude lines.
+            None will use default values which are resonable for maps of
+            North America.
+        projection : cartopy.crs class
+            Map projection supported by cartopy.
+        area_thresh : float
```
jjhelmus: This argument is not used in the method, remove.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -0,0 +1,510 @@
...
+            Coastline or lake with an area smaller than area_thresh in
+            km^2 will not be plotted.
+        min_lat, max_lat, min_lon, max_lon : float
+            Latitude and longitude ranges for the map projection region in
+            degrees.
+        width, height : float
```
jjhelmus: These arguments are not used in the method, remove.

jjhelmus:
Thanks for submitting this @jsignell.  I've provided a review with a number of suggestions on how this can be improved.  Let me know if you would like any assistance with these changes.

pyart/correct/phase_proc.py
```diff
@@ -949,7 +949,7 @@ def phase_proc_lp(radar, offset, debug=False, self_const=60000.0,
                   overide_sys_phase=False, nowrap=None, really_verbose=False,
                   LP_solver='cylp', refl_field=None, ncp_field=None,
                   rhv_field=None, phidp_field=None, kdp_field=None,
-                  unf_field=None, window_len=35, proc=1):
```
jsignell: Yes sorry I forgot I had that

pyart/graph/radarmapdisplay_cartopy.py
scollis: http://ajdawson.github.io/contact.html

jsignell: @ajdawson,  I included your gridlines for Lambert Conformal that I found [here].(http://nbviewer.jupyter.org/gist/ajdawson/dd536f786741e987ae4e) Is that ok?

ajdawson: @jjhelmus @jsignell - This is fine by me.

pyart/graph/radarmapdisplay_cartopy.py
jsignell: `transform` here should be the original projection of the data, which is always Lambert Conformal right? The map will be drawn in the projection of the `self.basemap` object, so we just need to know what projection we are coming from

jjhelmus:
@jsignell I think the issue is that the original gridded data is not defined with a projection, rather it is a set of points with Cartesian distances from a given latitude, longitude and altitude.  I've had discussions over if this type of gridded data has a associated map projection which have not been conclusive.  If such data does have an associated projection it would be a Azimuthal Equidistant projection not a Lambert Conformal.  

The `RadarMapDisplay` class makes no assumption about the projection to associate with the gridded data.  It leave it to the user to specify the map projection.  Lambert Conformal ('lcc') is default projection but only because some default is needed.  Basemap then can be used to transform the gridded Cartesian data into this projection.

   I think something similar can be done with Cartopy, leave the specification of the projection entirely to the user and using Lambert Conformal only as the default which gets entirely overridden when a user specifies a different projection.  Moving the creation of LambertConformal projection into the `if projection is None` block will solve some of this but the other uses of `self.proj` need to be worked out.

jsignell:
I am fine with switching the default to Azimuthal Equidistant, but I still don't understand how you propose to have a projection-less original grid. The units on all projections aren't meters or km, so you can't just hand the x, y over without saying what it mean. For instance in cartopy people use PlateCarree a lot, and the units are degrees of lat, lon. So we need to pass something that will transform the x, y coordinates (in meters or km) into the desired map projection.

Does that make sense?

jjhelmus:
As a concrete example the following will plot the unit test figure using a Mercator projection using the `RadarMapDisplay` class.  Only a Mercator projection is used.

``` Python
import pyart
import numpy as np

radar = pyart.io.read_cfradial(pyart.testing.CFRADIAL_PPI_FILE)
fig = plt.figure()
display = pyart.graph.RadarMapDisplay(radar, shift=(0.1, 0.0))
display.plot_ppi_map(
    'reflectivity_horizontal', 0, colorbar_flag=True,
    title="Fancy PPI Map", mask_tuple=('reflectivity_horizontal', -100),
    resolution='c', min_lon=-100, max_lon=-93, min_lat=33, max_lat=38,
    projection='merc',
    mask_outside=True)
display.plot_point(-95, 35, label_text='TEXT')
display.plot_range_rings([15, 30])
display.plot_line_geo(np.array([-95, -95]), np.array([33, 38]))
plt.show()
```

I think something along the lines of the following should work with the `RadarMapDisplayCartopy` class and never create a different projection:

``` Python
import pyart
import numpy as np
import cartopy

fig = plt.figure()
display = pyart.graph.RadarMapDisplayCartopy(radar, shift=(0.1, 0.0))
lat_0 = display.loc[0]
lon_0 = display.loc[1]
projection = cartopy.crs.Mercator(
                central_longitude=lon_0,
                min_latitude=33.0, max_latitude=38.)
display.plot_ppi_map(
    'reflectivity_horizontal', 0, colorbar_flag=True,
    title="Fancy PPI Map", mask_tuple=('reflectivity_horizontal', -100),
    projection=projection,
    min_lon=-100, max_lon=-93, min_lat=33, max_lat=38,
    mask_outside=True)
display.plot_point(-95, 35, label_text='TEXT')
display.plot_range_rings([15, 30])
display.plot_line_geo(np.array([-95, -95]), np.array([33, 38]))
plt.show()
```

jjhelmus:
> The units on all projections aren't meters or km, so you can't just hand the x, y over without saying what it mean

This is exactly why I really appreciate Basemap, all non-cylindrical Basemap class instances convert between Geographic (lon/lat) and Map Projection (x/y) coordinates.  Cartopy does not seem to specify what the units of the projections are.

pyart/graph/radarmapdisplay_cartopy.py
```diff
@@ -108,7 +108,8 @@ def plot_ppi_map(
             lat_lines=None, lon_lines=None, projection=None,
             min_lon=None, max_lon=None, min_lat=None, max_lat=None,
             width=None, height=None, lon_0=None, lat_0=None,
-            resolution='110m', shapefile=None, edges=True, gatefilter=None,
+            resolution='110m', shapefile=None, shapefile_kwargs={},
```
jjhelmus: shapefile_kwargs should be None and an empty dictionary created in the function.  [Mutable default arguments are dangerous](https://python-guide.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments)

dopplershift:
It's not about specifying the units, it's about having an associated projection object, which when inverted, gives you lon/lat coordinates corresponding to the projected coordinates, at least that's how I look at it. (For PlateCarree(), the inverted projection is a noop.)

jsignell:
I don't understand why it is a bad idea to give the x, y a basic projection (like Azimuth Equidistant) which doesn't ever show up in the map.

jjhelmus:
How about if the `proj` parameter is renamed to `grid_projection` and added as an optional parameter to the class constructor (the `__init__` method).  If this is not specified the default is to use a `crs.AzimuthalEquidistant` projection with central longitude and latitude at the radar.  This object is used to project from the grid coordinates to lon/lat coordinates.

jjhelmus:
> It's not about specifying the units, it's about having an associated projection object, which when inverted, gives you lon/lat coordinates corresponding to the projected coordinates, at least that's how I look at it. (For PlateCarree(), the inverted projection is a noop.)

I agree completely, but without knowing the units on the projections included in Cartopy I do not know how anyone can have confidence to say that their data uses that projection.

jjhelmus:
> I don't understand why it is a bad idea to give the x, y a basic projection (like Azimuth Equidistant) which doesn't ever show up in the map.

It can be a bad idea because you end up doing two projections which may distort the data for example:  
grid (meters) ----[Azimuthal Equidistant]---> lat/lon ----[Mercator]----> pixel locations

Where as with Basemap I can do

grid (meters) ---[Mercator]----> pixel locations

That said I do not think that two projections may be needed to use Cartopy.

dopplershift:
CartoPy uses Proj4 and doesn't specify units ever. Therefore, from the Proj4 docs:
```
Units

Horizontal units can be specified using the +units= keyword with a symbolic name for a unit (ie. us-ft). Alternatively the translation to meters can be specified with the +to_meter keyword (ie. 0.304800609601219 for US feet). The -lu argument to cs2cs or proj can be used to list symbolic unit names. The default unit is meters.
```

jjhelmus:
@jsignell I think this looks good.  Can you rebase this against the master branch so that it does not have conflicts with the base branch and then apply the norm/vmin/vmin fix to the class that was done in #618?  I'm happy to do these for you if you prefer.

jsignell:
@jjhelmus if you could do those things I would be super happy. I tried to rebase to get rid of that unrelated first commit, and got very confused and gave up.

jjhelmus:
@jsignell Not a problem, should be merged in a few minutes.  Thanks for seeing this through!

jsignell:
@jjhelmus Thanks for helping me along :)

jjhelmus:
Manually merged in commit fb1c62300a3569b8988ca64e9cf8abd2a06ef3ee after rebasing against master.

PR #616, opened 2016-10-25
ENH: location data available in HorizontalWindProfile class

kirknorth:
Adding latitude and longitude parameters/attributes to `HorizontalWindProfile` class. This information is now parsed from ARM radiosonde datasets in `read_arm_sonde`. I also considered adding `sample_time` as an optional parameter/attribute to the class, e.g., 1D array of datetimes, but thought otherwise. Let me know if I should add this as well, seeing as this information is available from standard soundings.

### DIFF
--- a/pyart/core/wind_profile.py
+++ b/pyart/core/wind_profile.py
@@ -32,9 +32,9 @@ class HorizontalWindProfile(object):
 
     Other Parameters
     ----------------
-    latitude : array-like, 1D
+    latitude : array-like, 1D, optional
         Latitude in degrees north at each height sampled.
-    longitude : array-like, 1D
+    longitude : array-like, 1D, optional
         Longitude in degrees east at each height sampled.
 
     Attributes
--- a/pyart/core/wind_profile.py
+++ b/pyart/core/wind_profile.py
@@ -30,6 +30,13 @@ class HorizontalWindProfile(object):
     direction : array-like, 1D
         Horizontal wind direction in degrees at each height sampled.
 
+    Other Parameters
+    ----------------
+    latitude : array-like, 1D
+        Latitude in degrees north at each height sampled.
+    longitude : array-like, 1D
+        Longitude in degrees east at each height sampled.
+
     Attributes
     ----------
     height : array, 1D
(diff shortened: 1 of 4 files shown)


pyart/core/wind_profile.py
```diff
@@ -39,20 +46,22 @@ class HorizontalWindProfile(object):
         Horizontal wind speed in meters per second at each height.
     direction : array, 1D
         Horizontal wind direction in degrees at each height.
-    u : array, 1D
+    u_wind : array, 1D
```
kirknorth: Corrected type u -> u_wind

kirknorth:
@jjhelmus I'll add a unit test shortly, or at least update a previous one.

pyart/core/wind_profile.py
```diff
@@ -90,3 +99,14 @@ def v_wind(self):
         # V = -cos(direction) * speed
         v_wind = -np.cos(np.deg2rad(self.direction)) * self.speed
         return v_wind
+
+    def _parse_location_data(self, latitude, longitude):
```
jjhelmus: Keeping this private makes sense.

pyart/core/wind_profile.py
```diff
@@ -30,6 +30,13 @@ class HorizontalWindProfile(object):
     direction : array-like, 1D
         Horizontal wind direction in degrees at each height sampled.
 
+    Other Parameters
+    ----------------
+    latitude : array-like, 1D
```
jjhelmus: add `, optional` to this since the argument is optional

jjhelmus:
Thanks for the addition @kirknorth.  Add `, optional` to the docstring of the new parameters and this should be good to go.  

I'm fine adding the PyCharm files to `.gitignore` since we include other editor temp file in there.  You may want to add these to your global ignore file so that they do not need to be added to every project.  GitHub has instructions on [how to do create a global .gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

kirknorth:
Thanks for the tip about creating a global .gitignore @jjhelmus! I'll push one more commit message shortly.

jjhelmus:
LGTM, merging once the CI pass.

PR #617, opened 2016-10-25
ENH: Added local install option for radx

scollis:
Added a keyword argument to pyart.aux_io.read_radx radx_dir which allows for the
user to define a local (non-root) install of RadX. This should be very handy for
users who just want to drop in the pre-compiled radx binaries. Tested on OSX
sierra with CPOL RAPIC files.

### DIFF
--- a/pyart/aux_io/radx.py
+++ b/pyart/aux_io/radx.py
@@ -39,7 +39,7 @@ def read_radx(filename, radx_dir = None, **kwargs):
     """
     # test for non empty kwargs
     _test_arguments(kwargs)
-    if radx_dir != None:
+    if radx_dir is not None:
         executable = os.path.join(radx_dir, 'RadxConvert')
     else:
         executable = 'RadxConvert'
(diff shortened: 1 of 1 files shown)


pyart/aux_io/radx.py
```diff
@@ -36,12 +39,16 @@ def read_radx(filename, **kwargs):
     """
     # test for non empty kwargs
     _test_arguments(kwargs)
+    if radx_dir != None:
```
kirknorth:
```python
if radx_dir is not None:
```

scollis:
Thanks @kirknorth ! Changed.. I am kind of rusty

jjhelmus:
Humm, not sure why Travis isn't triggering.  Merging anyways.

scollis:
Thanks @jjhelmus should have run pep8 first

Dr Scott Collis
Argonne National Laboratory
Mb: +1 630 235 8025
Of: +1 630 252 0550

PR #620, opened 2016-10-27
Addition of two Kdp estimation methods

wolfidan:
Dear all,

(Original request is at the end)

Sorry for the long delay, I was quite busy these weeks and I wanted to ask the original co-author of the IDL code (Jacopo Grazioli) for advice and help and he was in vacation and conference so...

We went carefully through the code and fixed several issues, one of them major. I am now confident that this code is in line with the original code written by Marc Schneebeli who designed the method.

Note that unfortunately for the Kalman filter method it is not possible to perform a simple geometric check using a linear Psidp, since it is not an exact method. First of all it pads the signal on both side with gaussian noise and also assumes that the measurements of phase are noisy (with standard deviation = 2). Hence, one can not expect to get a constant KDP.

However to show that the methods are globally consistent I estimated Kdp during one day of strong precipitation (29 September 2012, south of France) with PPI scans from an X-band radar (around 200 scans) using the three methods (Vulpiani with 20 gates window size, Kalman and Maesaka with standard setup). An example for one scan is given here.

ex_kdp
ex_phidp

It can be seen that the Kalman method generally gives less extremes and looks smoother, all three methods reproduce similar reconstructed Phidp.

In terms of overall scatter-plots (below), it can be seen that the methods are globally consistent as they are spread around the 1:1 line (dashed black line), however of course there is quite some variability between the methods, for example, the Kalman filter methods  tends to produce less extremes.

comparison_vulp_maesaka
comparison_kalman_maesaka

Best regards,

Daniel Wolfensberger

_Original request_:

As my lab's first contribution to the pyart project I would like to suggest adding two methods for KDP estimation, to the retrieve module. I have added these methods in the suggested kdp_proc.py file. The two methods are

kdp_vulpiani : a simple method for estimating kdp with moving windows (Vulpiani and al., 2012)

kdp_schneebeli : the ensemble Kalman filter method by Schneebeli and al. (2014).

One advantage of these methods over the Maesaka method is that they can retrieve negative KDP. These two methods have very similar calling procedures to the already implement method kdp_maesaka. Another addition is the function filter_psidp which can be optionally used prior to estimating Kdp to remove short subsequences of psidp, unravel angles with large discontinuities, use a median filter and mask psidp when rhohv is smaller than a given a threshold.

I provide a comparison of the three methods in the tests/compare_kdp_proc.py file for a real profile of measured psidp (on an X-band radar) as well as a comparison for one PPI (see below).


pyart/retrieve/kdp_proc.py
```diff
@@ -21,11 +33,994 @@
...
+    Parameters
+    ----------
+    radar : Radar
+        Radar containing differential phase field.
+
+    psidp_field : str, optional
+        Total differential phase field. If None, the default field name must be
+        specified in the Py-ART configuration file.
+    rhohv_field : str, optional
+        Cross_ orrelation ratio field. If None, the default field name must
```
jjhelmus: Cross correlation

pyart/retrieve/kdp_proc.py
```diff
@@ -21,11 +33,994 @@
...
+PADDING = 50 # Noise padding of the psidp signal (before and after signal)
+SHIFT = 13 # Shifting of the final signal
+
+from scipy import optimize, stats, interpolate, linalg, signal
+from pyart.retrieve import _kdp_proc
+from pyart.config import get_field_name, get_metadata, get_fillvalue
+
+def kdp_schneebeli(radar, gatefilter=None, fill_value = None, psidp_field=None,
+                kdp_field=None, phidp_field=None, band = 'C', rcov = 0, pcov = 0,
+                prefilter_psidp = False, filter_opt = {}, parallel = True):
```
jjhelmus: The default for filter_opt should be None with the empty dictionary generated in the function is None is detected.  [Mutable default options are dangerous.](https://python-guide.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments)

pyart/retrieve/kdp_proc.py
```diff
@@ -21,11 +33,994 @@
...
+            b1, b2, kdp_th)
+    kdp002f = kdp_dummy_b2[PADDING:nrg + PADDING]
+    
+    #Generate the ensemble of Kalman filters estimates in backward and
+    #Forward directions    
+    #--------------------Loop on scalers-------------------------------
+   
+    for i, sc in enumerate(SCALERS):     
+        #Forward
+        kdp_dummy_f2,phidp_dummy_f2, error_kdp = _kdp_estimation_forward_fixed(psidp,
```
jjhelmus: Use _ for the dummy arguments which are not used, e.g. `kdp_dummy_f2, _, error_kdp = _kdp_estimation_forward_fixed(`.  Also error_kdp is not used in this function.

jjhelmus:
@wolfidan Thanks for submitting this and updating your PR.  I'm excited to get these new Kdp processing function into Py-ART but have a few changes that I would like to see to the code.  I made some comments in the submitted Review, let me know if you have any questions or would like help addressing any of these.  

Additionally, Py-ART follows the style conventions in [PEP8](https://www.python.org/dev/peps/pep-0008/).  I  saw a number of places where your code does not follow these conventions, could you change these?  I uses the [pycodestyle](https://pypi.python.org/pypi/pycodestyle) tool to find places which do not meet these conventions.

pyart/retrieve/kdp_proc.py
```diff
@@ -10,6 +12,16 @@
 .. autosummary::
     :toctree: generated/
 
+    kdp_schneebeli
```
wolfidan: Fixed: I added them to the __init__.py file

pyart/retrieve/kdp_proc.py
```diff
@@ -21,11 +33,994 @@
...
+    F : 4x4 float array 
+        Forward state prediction matrix [4x4]
+    F_transposed: 4x4 float array 
+        Transpose of F
+    H_plus : 4x3 float array
+        Measurement prediction matrix [4x3]
+    c1, c2,b1,b2: floats
+        the values of the intercept of the relation c  = b*Kdp - delta.
+        This relation uses b1, c1 IF kdp is lower than a kdp_th and b2, c2 
+        otherwise kdp_th
```
wolfidan: I added the following doc:

    kdp_th: float
        the kdp threshold which separates the two Kdp - delta regime
        i.e. the power law relating delta to Kdp will be different if Kdp is
        larger or smaller than kdp_th
    mpsidp: float
        final observed value of psidp along the radial (usually also the max value),
        needed for inverting the psidp vector

pyart/retrieve/kdp_proc.py
```diff
@@ -21,11 +33,994 @@
...
+    F : 4x4 float array 
+        Forward state prediction matrix [4x4]
+    F_transposed: 4x4 float array 
+        Transpose of F
+    H_plus : 4x3 float array*np.nan
+        Measurement prediction matrix [4x3]
+    c1, c2,b1,b2: floats
+        the values of the intercept of the relation c  = b*Kdp - delta.
+        This relation uses b1, c1 IF kdp is lower than a kdp_th and b2, c2 
+        otherwise kdp_th
```
wolfidan: See last point

pyart/retrieve/test_kdp_proc.py
```diff
@@ -0,0 +1,69 @@
+""" Unit tests for pyart.retrieve.kdp_proc module. """
```
wolfidan: I removed the file as it was basically a duplicate of the same file in the tests directory

pyart/retrieve/tests/compare_kdp_proc.py
```diff
@@ -0,0 +1,192 @@
+""" Unit tests for pyart.retrieve.kdp_proc module. """
```
wolfidan: Changed it to 

""" 
Unit tests for pyart.retrieve.kdp_proc module. 
Can be used to test the Vulpiani and Maesaka estimation methods for the ideal
case of a linear psidp profile with specified slope
"""

wolfidan:
Sorry for the long delay,

I tried to correct all issues you raised,

Best regards,

Daniel Wolfensberger

scollis:
hey @wolfidan We will get to this.. Things are complicated right now by staffing changes and the upcoming ARM meeting.

pyart/retrieve/tests/compare_kdp_proc.py
```diff
@@ -0,0 +1,192 @@
...
+    phidp_dict_mae = 0.5 * (phidpf_dict_mae['data'][0] + 
+                            phidpr_dict_mae['data'][0])
+    plt.plot(ranges,phidp_dict_mae)
+    plt.plot(ranges,phidp_dict_vulp['data'][0])
+    plt.plot(ranges,phidp_dict_schnee['data'][0])
+    plt.plot(ranges,prof_psidp.fields['differential_phase']['data'][0])
+    
+    plt.xlabel('Range [m]')
+    plt.ylabel(r'Diff. phase [deg]')
+    plt.legend(['Maesaka','Vulpiani','Schneebeli','Real Psidp'],loc=0)
```
jjhelmus: Would plt.show be helpful here to bring up a window with the plot when this is run from the console?

pyart/retrieve/tests/compare_kdp_proc.py
jjhelmus: This file is not a unit test so should not be labeled as such.  How about:

A script for comparing the e Vulpiani, Maesaka and Schneedbeli methods for KDP estimation with sample data.

jjhelmus:
@wolfidan There are still a few minor changes that I would like to suggest.  In addition there are a number of style issues with the `kdp_proc.py` module.  Can you run a tool like [pycodestyle](https://pypi.python.org/pypi/pycodestyle) on this file and make the changes needed to conform to the [PEP8](https://www.python.org/dev/peps/pep-0008/) standard?

pyart/retrieve/tests/compare_kdp_proc.py
wolfidan: Ok I fixed this

pyart/retrieve/tests/compare_kdp_proc.py
wolfidan: I added the plt.show() at the end

wolfidan:
I have fixed all PEP8 issues I could find (I used the PEP8 issues highlighting of the spyder IDE) in all the three files I have modified

scollis:
Thanks Daniel!

I am at the ARM-ASR meeting this week, I will take a look when I get back.

scollis:
@Rumpkie Could you please download this PR and check with pycodestyle if the code meets PEP8 we are good to go.. Sorry for taking so long

scollis:
@wolfidan : can you push a commit with those two small changes made? If I do not hear anything by friday I will accept and fix the PeP8 with a separate PR

scollis:
Great Work @wolfidan ! This is very much appreciated. Merging

PR #623, opened 2016-11-21
DOC: Update INSTALL.rst file

jjhelmus:
* Update supported Python versions
* Remove minimum versions from other packages
* Update URL for netcdf4-python

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -5,28 +5,27 @@ Building and installing Py-ART
 Required Dependencies
 =====================
 
-Py-ART requires the following software, earlier version may work but have 
-not been extensively tested.
+Py-ART requires the following software.
 
-* Python__ 2.6.x or newer
+* Python__ 2.7.x, 3.4.x or 3.5.x
 
 __ http://www.python.org
 
-* NumPy__ 1.6.x or newer
+* NumPy__
 
 __ http://www.scipy.org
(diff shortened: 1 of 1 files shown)


dopplershift:
Do you no longer have specific versions of numpy/scipy you support?

jjhelmus:
In all truthfulness I don't know what the minimum versions of numpy/scipy are for Py-ART.  Older versions should work fine or will give a loud error when trying to use some "new" API feature but I have not extensively tested how far back this can go.  

Checking a few version of numpy with conda, I found that NumPy 1.9 is required for the unit tests to pass, but the failures from older versions (1.8 and 1.7) are likely caused by test which expect an over ambitious precision and an minor API change in matplotlib which I doubt would effect actual use of the package.  NumPy 1.6 fails a few more tests but against mostly due to minor precision issues and a few API changes. 

Is it worth putting in the time to track down and fix any bugs so that exact minimum version of Py-ART dependencies can be specified?  It definitely could be done, and a Travis job could be added to run tests in this minimum environment, but I'm not sure if the payoff would be worth the effort.

dopplershift:
Makes sense...I was just curious.

jjhelmus:
I think I'll leave it as is, if there is a call for more details on minimum versions this can be revisited.

PR #625, opened 2016-12-13
improving gamic reader

TulipaSilva:
Every gamic file has the atributte named 'range_samples'. Some times it is equal 1, but sometimes not.
This pull request fixes the range_step when the range_samples is not equal 1.


jjhelmus:
Looks good! Thanks for the improvement @TulipaSilva

PR #647, opened 2017-04-12
added Raster keyword for happier vector file output

swnesbitt:
This PR addresses the issue of large file size outputs in `plot_ppi*` displays when creating vector graphics for publications.  If saving to pdf, eps, or svg output, `pcolormesh` was rendering many polygons, and not clipping those polygons to the plot window, meaning that for high resolution data, the output file sizes were huge (sometimes exceeding 100 megabytes), and editing the resulting images in graphics software such as Illustrator was difficult.  Such large files are also a world of pain in the publication process.

The PR introduces the raster keyword, which, when set to True, will cause the artist (`pcolormesh`) to use `pm.set_rasterized(True)`.  Behavior is described here: [http://matplotlib.org/api/_as_gen/matplotlib.artist.Artist.set_rasterized.html](http://matplotlib.org/api/_as_gen/matplotlib.artist.Artist.set_rasterized.html)

When saving the file, setting the `dpi` keyword in `plt.save_fig()` changes the resolution of the raster image in the plot.

An example pdf (5.6 MB) file is below from a multi-panel NEXRAD level 2 sweep.

[KILX20160316_020459_V06.gz.pdf](https://github.com/ARM-DOE/pyart/files/916604/KILX20160316_020459_V06.gz.pdf)

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -234,7 +234,7 @@ class RadarDisplay(object):
             colorbar_flag=True, colorbar_label=None,
             colorbar_orient='vertical', edges=True, gatefilter=None,
             filter_transitions=True, ax=None, fig=None,
-            ticks=None, ticklabs=None, **kwargs):
+            ticks=None, ticklabs=None, raster=None, **kwargs):
         """
         Plot a PPI.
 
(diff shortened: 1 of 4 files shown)


scollis:
Thanks Steve! Will take a look once I am back in Chicago

scollis:
Did you alternate between False and None between plot_rhi and plot_ppi @swnesbitt ?

scollis:
Thanks @swnesbitt ! This looks good.. Merging

PR #650, opened 2017-05-17
H5Py OSError: File close degree doesn't match fix

vlouf:
I had a problem when I tried to open an ODIM HDF5 file. This bug comes from H5Py, not pyart. This error is explained here:
https://github.com/h5py/h5py/issues/218
I just made a little change in pyart.aux_io.odim_h5.py to fix this bug.


scollis:
Thanks @vlouf This looks like a good but unfortunate fix.

PR #651, opened 2017-05-19
FIX: Coef in B-Matrix

scollis:
There was no way to set the coef for relation between KDP and Z in the LP Phase
processing. It was hard set to C-Band. Some how we dropped this in a version.


scollis:
Checked working and tests passed.. Merging

PR #656, opened 2017-05-31
AttributeError fix with python 3.6 in cfradial.py

vlouf:
This is a fix for a problem I had after an update (netcdf4 1.2.8 np112py36_0).

`type(netCDF4.chartostring(sweep_mode['data'][0])[()])` is `numpy._str` and there is no `decode` attribute in numpy._str

Because all strings are already in unicode in python we can drop the `decode('utf-8')`.

### DIFF
--- a/pyart/io/cfradial.py
+++ b/pyart/io/cfradial.py
@@ -189,7 +189,12 @@ def read_cfradial(filename, field_names=None, additional_metadata=None,
         ray_angle_res = None
 
     # first sweep mode determines scan_type
-    mode = netCDF4.chartostring(sweep_mode['data'][0])[()].decode('utf-8')
+    try:
+        mode = netCDF4.chartostring(sweep_mode['data'][0])[()].decode('utf-8')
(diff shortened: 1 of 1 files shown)


zssherman:
@vlouf The travis check was failing because when the test does
`assert np.all(str_array == [b'azimuth_surveillance'])` str_array actually equals `['azimuth_surveillance']` no byte at the beginning, due to the updated netCDF. I'll have a fix for that, but the request will have a review from Cory and Scott soon. Thanks again for the fix.

Rumpkie:
@scollis I think it would be good for @zssherman to update the CI test so that this passes, right now the expected result only passes the current py-art way of processing the string. Fix the test and then merge this PR. Thoughts?

scollis:
Yes.. @zssherman : please update the unit test in a separate PR. Put a check for netCDF version in there.. for old one use the byte (ie b'blah') and for new add a test that will pass with the test file which is complaint. After that we are good to Merge

PR #660, opened 2017-06-19
TST: Fix for failing test due to assert of an older result.

zssherman:
The assert equal for radar.sweep_mode was failing due to the b byte in
front of the 'azimuth_surviellance. Netcdf chartostring already decodes.
Adding try and except AssertionError to test both versions of netCDF.


scollis:
Thanks! I will check to see if CI passes.

scollis:
oops @zssherman: you need to make a clean branch to make this change as there is a bunch of other commits in this

zssherman:
Ah yep the merge, my bad. I'll redo.

PR #665, opened 2017-06-22
TST: Fix for numpy 1.13 update.

zssherman:
Tests involving gate_x_y_z were failing most likely due to numpy 1.13. The radar gate arrays are now float32 in the newer version of numpy. Older version, gate arrays are float64. Still working on why the arrays dtype changed.

The atol values for the numpy.testing.assert_allclose tests were
changed for lower precision to better reflect the dtype change. The
tests changes work for both the newer and older versions of numpy.


scollis:
Thanks for this essential PR @zssherman 
Users: We are still unsure of exactly what has changed that is causing this.. We know that we can see the difference between Numpy 1.12 and 1.13 but from the change logs over in Numpy land we can not work out what specific change was made. I have a feeling that it changed the upcasting when you multiply float32 by float(int).. If anyone knows we would love input.. for now we have just changed the tolerance on the assert_isclose to account for the fact range data is now 32 bit (which is more than enough precision).

scollis:
Tests pass.. For some reason the "Build" is failing.. We need to get to the bottom of this. Merging

PR #668, opened 2017-07-06
Fixed issue with range()

wolfidan:
Hi,

In my Kdp retrieving routines I have had some issues with using the standard Python function range. I have solved this by using numpy's arange instead. This is a very minor change.

Cheers,

Daniel


scollis:
Thanks @wolfidan .. BTW this is where unit tests help :) Merging.

PR #672, opened 2017-07-19
MAINT: Move calculate_velocity_texture to pyart.retreive

rcjackson:
calculate_velocity_texture is moved to pyart.retrieve.


scollis:
Thanks @rcjackson !

PR #673, opened 2017-07-25
DOC: Noticed link to wradlib was to the old bit bucket page, no longer exists.

zssherman:
Link in README was outdated and the link has been updated to wradlib's docs latest page.


scollis:
Thanks @zssherman merging

PR #687, opened 2017-09-20
Change graph titles and filename to use time at beginning of sweep

hichris1234:
Use the time at the beginning of a sweep to name a graph/filename, not the beginning time of a radar.

This seems to make more sense, as the beginning of the sweep could be minutes from the beginning of the radar.


josephhardinee:
I think this should probably be something that is configurable (Probably a kwarg). It is often very useful to see the file time instead (Knowing which volume you are using, and matching volumes up across plots).

hichris1234:
Fair point, @josephhardinee. Should I go ahead and add that as a parameter to e.g. the plotting/file saving functions?

josephhardinee:
Let's let @scollis chime in, but I think it would make sense for the plotting at least.

scollis:
Yep! Sorry been busy.. I would like to see it as a keyword with a default.. Basically make it like a strftime string

hichris1234:
@scollis I'm not quite following that, like the parameter would determine what parts of the begin datetime and what parts of the beginning of the sweep datetime would be used?

scollis:
Sorry for the delay @csnardi 
What I am saying is that we have a datetimeformat=None in the kwargs
then 
if datetimeformat == None:
    datetimeformat = 'SOMECOOLDATETIMEDEFAULT'

csnardi:
No worries @scollis; I've kept meaning to work on this but forgot.

Do you think that just accepting a `strftime` string would make sense for this parameter? And we could also have boolean parameter to specify whether the current sweep or the first sweep should be used to generate that time?

scollis:
Exactly.

csnardi:
I wasn't able to test `plot()` itself to verify that I added the parameters correctly, but I added in test for the helper methods and `generate_filename()`.

scollis:
@rcjackson Can you review please. We want to get this in as part of Picasso release. @csnardi Ready to go?

csnardi:
Yep, should be good; all the test I added are passing and I believe the changes to `plot()` should be correct.

scollis:
Great! @rcjackson is just going to take a real quick look. Good news is we are cutting a new release this week so it will make it into the next Conda-forge package.

rcjackson:
I just tested this feature on an example file. It works great! I say it's ready to go.

scollis:
Sweet.. Merging.

PR #688, opened 2017-09-23
ENH: Drop FORTRAN dependancy for computing Steiner classification

vlouf:
I have translated the FORTRAN code for computing the Steiner classification in full Python. I didn't remove the original FORTRAN, now if `if not _F90_EXTENSIONS_AVAILABLE:` is True, instead of raising an error it calls the new python code. 

I think you could take the decision to completely drop the FORTRAN dependancy. But for now, it can be useful for people who are installing pyart through conda or pip (and therefor can't call the `steiner_conv_strat`.

Note that the PYTHON code has identical performance compared to the FORTRAN code (I'm using `numba`).


vlouf:
The module `numba` is required. That's why it failed.

vlouf:
Not that slower, it might be 1s instead of 100ms. With numba the function execution time is equal to the one of the old fortran routine.

But yes, I share similar feelings about numba, it's really impressive when you can use it, but it's a complete black box, you don't know what it's doing.

vlouf:
I don't know how to create an empty decorator that take argument, so it was easier to drop numba dependency...

scollis:
@rcjackson please review.

rcjackson:
Valentin's code:
valentins fork conv strat

Current code:
default conv strat

Same grid, same settings, same output, and his new code is reasonably fast.

scollis:
Ok. I am accepting. If anyone complains we can roll back.

scollis:
Thanks @vlouf !!

PR #692, opened 2017-10-13
FIX: Python 3 issue with KAZR ingest

scollis:
iteritems no longer supported for dict in Python 3.. simple fix


scollis:
Tests have passed. Minor fix so I am merging

Issue #690, opened 2017-09-29
Code Stalls When Using pyart.correct.dealias_fourdd


pjmarinescu:
Hello, 

I have been using pyart to read in and dealias both NEXRAD Level 2 and CSU-CHILL radar velocity data. I have been using the pyart.correct.dealias_fourdd function. It works great for the NEXRAD data, but the code stalls for the CSU-CHILL data. The CSU-CHILL data I am using is not in the traditional CHL format, but has been converted to uf files (these files are ~30 MB). [I was able to make a sample file available, which can be found here: http://vandenheever.atmos.colostate.edu/PJM/CHL20160717_222347.uf.gz]

From my own tests, I believe the code is stalling at the following line:
_rsl_interface._label_volume(rsl_volume, radar)

One difference between the data is that the CSU-CHILL data is sector scans, while the NEXRAD data is full sector PPIs. 

Do you have any ideas what may be causing the code to stall with the CSU-CHILL uf files?

I have included below the lines that read in the data, as well as the call to the dealias_fourdd.

file = '/Volumes/42/CHILL/ORIG/20160717/CHL20160717_222347.uf'
radar = pyart.io.read_uf(file,field_names={'DZ':'DBZ', 'VE':'VEL'})
dealias=pyart.correct.dealias_fourdd(radar,sounding_heights=aint, sounding_wind_speeds=wsint,\
                                      sounding_wind_direction=wdint,vel_field='VEL')

Please note that I have tried several different read commands (pyart.io.read_uf, pyart.io.read, not including the field_names specification) with the same code stall. Also, I use the same sounding data for the NEXRAD and CHILL data, and since it works with the NEXRAD data, I am assuming that the sounding data is not the issue.

Thanks for your time,

Peter

freemansw1:
Just to follow up on this- I've been working on debugging this now and have narrowed it down to the while loop in the findRay function of dealias_fourdd.c. It appears that under certain circumstances, the while loop can never find a ray that is sufficiently close, so it gets caught in an infinite loop. 

I have been able to somewhat resolve the problem by replacing the relatively complex findRay function with a much simpler version and just putting closeness checks into continuity_dealias. Working now on a sanity check before I do a PR, any thoughts on this? I don't see a reason to keep that while loop in the findRay function and not replace it with a for loop.

scollis:
Hey Sean,
Is this in Py-ART or 4DD itself?



--
Scott Collis

[...]

PR #693, opened 2017-10-18
Fixing infinite loop in 4DD algorithm

freemansw1:
This PR is to address #690 (pinging @pjmarinescu ). I replaced the findRay function with a completely new and simplified function. I then check in the continuity_dealias function to make sure that the spacing is reasonable. This seems to fix the error, although I would definitely appreciate a second look. 

On our first look at the data, it looks like for the specific file that we're working with, it looks mostly okay, with some small oddities that could also be related to needing to QC the original data. 

As an aside, I don't think that my spacing variable makes a lot of sense- it's identical to the one in the original code, but the indices (0 and 50) seem arbitrary or specific to some radar. I'm not quite sure what that's doing, so I'm open to changes. For reference, the spacing is defined as: `spacing = fabs(sweep2->ray[0]->h.azimuth - sweep2->ray[50]->h.azimuth);`. 

Hopefully this works to resolve this issue- I'm open to any feedback on this, it's been quite a spin up to try to debug this code.

### DIFF
--- a/pyart/correct/src/dealias_fourdd.c
+++ b/pyart/correct/src/dealias_fourdd.c
@@ -332,7 +332,7 @@ static void continuity_dealias(
             float azold = sweepc->rv->ray[ray_index]->h.azimuth;
             float diffaz_newold = fabs(aznew-azold);
             if(diffaz_newold > 180.0)
-                diffaz_newold = 360.0-diffaz_newold
+                diffaz_newold = 360.0-diffaz_newold;
             if(fabs(aznew-azold)>spacing/2){
                 valid_above_ray = 1;
             }
--- a/pyart/correct/src/dealias_fourdd.c
+++ b/pyart/correct/src/dealias_fourdd.c
@@ -270,79 +270,27 @@ static void init_sweep(Sweep *sweep, float val)
 static int findRay(Sweep *sweep1, Sweep *sweep2, int ray_index) 
 {
     int numRays, rayIndex1;
-    float az0, az1, diffaz;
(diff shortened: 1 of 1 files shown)


scollis:
Thanks @freemansw1 ! We will check this out.

freemansw1:
So I've done some more investigation into the spacing function- it's divided by 50 later, so it appears to be essentially an average of the azimuthal width of the rays over 50 rays. This may have been the original issue- for sector scans especially. To resolve this, I've changed it to be an average of four rays and added a comment as such. I'm not sure that this is any less arbitrary, but at the least, I doubt that there will be any use case where the number of rays is fewer than 4 (although given that the failure may not be pretty, maybe it should fail earlier?).

scollis:
Hey @rcjackson can you have a look at review this pull request in the next week or two for me?

rcjackson:
I've had a few cases that produced an infinite loop in 4DD or even a segfault and it was due to small errors in the azimuthal angles from CPOL that would make the spacing negative. This should fix that. 

I wonder if a better way is to determine the spacing over all rays and take the mode of them. It would be at least less arbitrary.

rcjackson:
I tested this with my dealiasing notebook and it works fine. For now, we can keep the spacing @freemansw1 suggested. It looks like CI passes too.

scollis:
Just to be 100% sure I am going to close and re-open to force CI to run again.

scollis:
CI passed.. Happy Day! Merging and thank you @freemansw1

Issue #696, opened 2018-01-16
The circular reference in Radar Class


hhuangwx:
The LazyLoadDict used in the Radar class causes a circular reference, which will make the memory used by the class never collected by the automatic gc system (known as the memory leakage).  
![image](https://user-images.githubusercontent.com/11309009/34994685-4f062d8e-fb0f-11e7-9669-c8b9ceb48b94.png)

If you are agree with my opinion that the LazyLoadDict should not utilized, I can help to fix the above problem.

scollis:
Thanks @hhuangwx ! @zssherman take a look at this

scollis:
And yes.. a PR would be very welcome..

scollis:
Ha! Your PR found a bigger issue as well thanks to the CI. Numpy's np.fromstring is depricated. 
We need to fix this first. It will not impact CF-Radial.. but it will impact the binary readers.

zssherman:
Closing this for now.

PR #697, opened 2018-01-17
fix of circular reference in Radar and Grid Class

hhuangwx:
fix of #696 

"The LazyLoadDict used in the Radar class causes a circular reference, which will make the memory used by the class never collected by the automatic gc system (known as the memory leakage)."

The LazyLoadDict is removed and the relevant parts are refactored in the Radar and Grid Class in this PR. The other codes using LazyLoadDict (e.g. the mdv class) are not changed.

### DIFF
--- a/pyart/core/grid.py
+++ b/pyart/core/grid.py
@@ -253,7 +253,6 @@ class Grid(object):
         self.point_latitude = get_metadata('point_latitude')
         self.point_latitude['data'] = geographic_coords[1]
 
-
     def init_point_altitude(self):
         """ Initialize the point_altitude attribute. """
         self.point_altitude = get_metadata('point_altitude')
--- a/pyart/core/grid.py
+++ b/pyart/core/grid.py
@@ -13,13 +13,9 @@ An class for holding gridded Radar data.
 .. autosummary::
     :toctree: generated/
 
-    _point_data_factory
-    _point_lon_lat_data_factory
-    _point_altitude_data_factory
 
 """
 
-import warnings
 
 import numpy as np
 
@@ -30,7 +26,6 @@ except ImportError:
     _PYPROJ_AVAILABLE = False
 
 from ..config import get_metadata
-from ..lazydict import LazyLoadDict
 from .transforms import cartesian_to_geographic
 from .transforms import cartesian_vectors_to_geographic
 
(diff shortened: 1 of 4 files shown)


scollis:
Thank you @hhuangwx ! Much appreciated! 
@zssherman please review. @hhuangwx will this kill our ability to load a file without reading the moments?
We will take a look! I wonder if we are better off leaving the lazy load methods in and only having them used when a partial load is done.. These were implemented by @jjhelmus

hhuangwx:
I think the delay_field_loading used in read_cfradial is not affected by this PR. The circular reference is caused by fact that the functions used by LazyLoadDict point back to the Radar or Grid Class again. In the delay_field_loading function, the function used by LazyLoadDict points to the ncobj, which will not cause the circular reference. It'll be more elegant if anyone figures out a way to fix the circular references without removing the lazy loading in Radar and Grid.

hhuangwx:
I think the potential solution for the delayed loading is to use functions with the "property" decorators as the attributions (e.g., gate_x). The computations to get the attributions are delayed until we call/use the attributions. Actually, the LazyLoadDict shares the same philosophy with this idea.

scollis:
Thanks @hhuangwx @zssherman needs to investigate why the CI failed as well. It is very important we maintain compatibility with users existing workflows and not break them

dopplershift:
Why do you think circular references cause memory leaks? Python has a garbage collector it runs whose sole purpose in life is to collect circular references and avoid memory leaks:

https://pythoninternal.wordpress.com/2014/08/04/the-garbage-collector/

scollis:
Closing and opening to force run unit tests

hhuangwx:
@dopplershift You're right about the ability that the standard library "gc" can collect the garbage if we explicitly call it. However, "Python doesn’t collect such cycles automatically " according to the document of the gc library. That's the reason I think we should break the reference cycles to avoid the manual gc procedures.

image

scollis:
@hhuangwx are you able to run the nosetests locally? I see a lot of issues on the CI.

hhuangwx:
I'll try the nosetests. @scollis

jjhelmus:
Neither the Radar class nor the LazyLoadDict class have a `__del__` method so the Python garbage collector will have no issue collecting the cyclic references automatically.

If there is a desire to break the circular reference, passing a `weakref.proxy` instance to the factory function should work.

scollis:
Thanks @jjhelmus ! 
@hhuangwx I want to know more about the specific problems this is causing you. We appreciate your contribution but LazyLoadDict was designed to enable partial loading of radar files and the building of data models where calculations could be performed later. What you are proposing is a pretty big change so we need a careful review.

hhuangwx:
Thanks @jjhelmus @scollis. I'm kind-of busy these days. I'll perform additional test about the memory when I'm free.

scollis:
Absolutely no hurry. 

-sent from a mobile device-

scollis:
I am going to close this for now. If anyone gets back to this we can reopen

PR #704, opened 2018-01-18
FIX: last part of fromstring fix

scollis:
Added _sigmetfile.c generated from _sigmeetfile.pyx


scollis:
Again, the same issue with 2.7 that is not actually a failed test..

PR #706, opened 2018-01-23
EDIT: Adding new version of dealiasing code

rcjackson:
This is the new version of the dealiasing code that includes the option to minimize the difference between the folded regions and a sounding for certain scans that the region based dealiasing algorithm overcorrects.

### DIFF
--- a/pyart/correct/region_dealias.py
+++ b/pyart/correct/region_dealias.py
@@ -27,10 +27,13 @@ import warnings
 import numpy as np
 import scipy.ndimage as ndimage
 
+from scipy.optimize import fmin_l_bfgs_b
+
 from ..config import get_metadata, get_fillvalue
 from ._common_dealias import _parse_fields, _parse_gatefilter, _set_limits
 from ._common_dealias import _parse_rays_wrap_around, _parse_nyquist_vel
 from ._fast_edge_finder import _fast_edge_finder
+
 
 # Possible future improvements to the region based dealiasing algorithm:
 #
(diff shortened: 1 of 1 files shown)


scollis:
Very welcome PR @rcjackson ! CI is failing on your new function however..
```
test_region_dealias.test_dealias_reference_velocity_field ... RUNNING THE L-BFGS-B CODE
           * * *
Machine precision = 2.220D-16
 N =            4     M =           10
At X0         0 variables are exactly at the bounds
/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/numpy/ma/core.py:5147: RuntimeWarning: Mean of empty slice.
  dtype=dtype, **kwargs)[()]
/home/travis/miniconda3/envs/testenv/lib/python3.6/site-packages/numpy/core/_methods.py:73: RuntimeWarning: invalid value encountered in true_divide
  ret, rcount, out=ret, casting='unsafe', subok=False)
At iterate    0    f=  1.12717D+01    |proj g|=  5.00000D+00
At iterate    1    f=  1.12483D+01    |proj g|=  5.56359D+00
 Bad direction in the line search;
   refresh the lbfgs memory and restart the iteration.
           * * *
Tit   = total number of iterations
Tnf   = total number of function evaluations
Tnint = total number of segments explored during Cauchy searches
Skip  = number of BFGS updates skipped
Nact  = number of active bounds at final generalized Cauchy point
Projg = norm of the final projected gradient
F     = final function value
           * * *
   N    Tit     Tnf  Tnint  Skip  Nact     Projg        F
    4      2     48      9     0     3   5.564D+00   1.125D+01
  F =   11.248300552368162     
ABNORMAL_TERMINATION_IN_LNSRCH                              
 Line search cannot locate an adequate point after 20 function
  and gradient evaluations.  Previous x, f and g restored.
 Possible causes: 1 error in function or gradient evaluation;
                  2 rounding error dominate computation.
 Cauchy                time 0.000E+00 seconds.
 Subspace minimization time 0.000E+00 seconds.
 Line search           time 0.000E+00 seconds.
 Total User time 0.000E+00 seconds.
FAIL
```

rcjackson:
The unit test catches an edge case that broke the optimization loop due to it converging to the wrong local minima. A change in the initial state from the zero vector to a vector of ones seems to fix the issue.

scollis:
Thanks @rcjackson ! I see all green ticks now.. 

Two Questions:
1) Can the user still run region dealias without this new feature?
2) do you think this will break anyone else's code?

rcjackson:
1. Yes, if you don't provide a reference velocity field then the new feature will not be on.

2. There will be a difference in the velocity field if they are providing a reference velocity field, so if they are providing that they may want to be aware of the change.

scollis:
Very nice @rcjackson I will accept this PR. We will make note of this when we release Py-ART V1.9-Picasso

Issue #701, opened 2018-01-17
Small edit to the contributors guide


scollis:
The way we tell people to run unit tests will not work..
We need to tell users to run
`python setup.py build_ext --inplace`

PR #715, opened 2018-03-01
DOC: Added pyart build in place install for nosetests and added import order list.

zssherman:
This fixes issue #701


scollis:
Looks great @zssherman thanks for the docfix

PR #733, opened 2018-05-06
Fix TypeError when displaying products without standard names

wave100:


scollis:
Thanks for the PR! 
can you post the error message encountered? 
I can't see a situation when a string would not be passed 
https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/radardisplay.py#L1413

wave100:
Thanks for taking a look at it!

I've only been running into errors in a pretty narrow use case - Running pyart.io.read() on NEXRAD Level 3 products with file_field_names=True. I've been working with data from the NWS FTP server.

Initially, I was running into the error in #734. Once I got past that, I tried plotting my data with a RadarMapDisplay. I can't get an exact error message right now (getting basemap running in a virtualenv is proving difficult) but I'll see if I can try it on another machine tomorrow. Let me know if you need anything else!

wave100:
After doing some more digging, I've been able to isolate the root cause of #733 and #735 to line 96 of nexradl3_read.py. The fix is to change it from:

`field_name = filemetadata.get_field_name(msg_code)`

to:

`field_name = str(filemetadata.get_field_name(msg_code))`

This accounts for the case where filemetadata.get_field_name returns an int if the file_field_names kwarg is True.

scollis:
Interesting.. Since some users may be used to using ints instead of strings with level 3 I will accept your PR rather than enforcing a field name to always be a string

PR #734, opened 2018-05-06
Fix bug in sign calculation

wave100:
"b0b1000000000000000" ->  "0b1000000000000000"


scollis:
Thanks @wave100 !! @rcjackson Do you have some level3 data we can test this on? 
All tests have passed and this is a small change so don't spend a huge amount of time.. @jjhelmus as the code's author if you have a view I would appreciate it if you have time to post

scollis:
Thanks @jjhelmus !

rcjackson:
It works on an example level 3 file from KLOT I just downloaded with no issues.

scollis:
Perfect! Thanks @wave100 for the nice bugfixes!

PR #736, opened 2018-05-09
FIX: netcdftime was changed to cftime. Causes travis to fail. Changed import name to reflect new name.

zssherman:


zssherman:
@scollis Changing to 64 fixed all issues. There are a few other things to look at with CI, but this will fix it for now.

scollis:
Accepting

PR #746, opened 2018-05-22
allow reading from MDV with unknown vertical level

gamaanderson:
While reading MDV warns for unknown 'vlevel_type', set units to 'unknown' and continue


scollis:
Thanks @gamaanderson !

scollis:
Looks great.. merging

PR #748, opened 2018-05-29
FIX: Fixes a bug where set_rasterized is always True.

zssherman:
The if raster statement checks for 'if not None' False will count as this so set_rasterized will be True.


scollis:
@nguy can you provide a quick comment as it touches airborne ?

zssherman:
@scollis Sorry completely forgot to tag him.

nguy:
I don't see any issues with this. Thanks @zssherman and @scollis !

scollis:
Looks good.. merging

Issue #749, opened 2018-06-01
ValueError in pyart.retrieve.map_profile_to_gates


vlouf:
Error when using `pyart.retrieve.map_profile_to_gates`

```python
Traceback (most recent call last):
  File "/miniconda3/envs/radar/lib/python3.6/site-packages/cpol_processing-1.0-py3.6.egg/cpol_processing/processing/radar_codes.py", line 454, in snr_and_sounding
    radar)
  File "/miniconda3/envs/radar/lib/python3.6/site-packages/pyart/retrieve/gate_id.py", line 74, in map_profile_to_gates
    fld = np.ma.masked_equal(f_interp(z + radar.altitude['data'][0]),
  File "/miniconda3/envs/radar/lib/python3.6/site-packages/scipy/interpolate/polyint.py", line 78, in __call__
    x, x_shape = self._prepare_x(x)
  File "/miniconda3/envs/radar/lib/python3.6/site-packages/scipy/interpolate/polyint.py", line 90, in _prepare_x
    x = _asarray_validated(x, check_finite=False, as_inexact=True)
  File "/miniconda3/envs/radar/lib/python3.6/site-packages/scipy/_lib/_util.py", line 236, in _asarray_validated
    raise ValueError('masked arrays are not supported')
ValueError: masked arrays are not supported
```

I have this error with 2 different versions of scipy, 1.0.0 and 1.1.0

vlouf:
More details:
```python
ValueError                                Traceback (most recent call last)
<ipython-input-44-d46e0cfba1f5> in <module>()
----> 1 pyart.retrieve.map_profile_to_gates(temperatures, my_profile['height'].filled(np.NaN), radar)

/miniconda3/envs/radar/lib/python3.6/site-packages/pyart/retrieve/gate_id.py in map_profile_to_gates(profile, heights, radar, toa, profile_field, height_field)
     72         heights[:toa], profile[:toa], bounds_error=False,
     73         fill_value=get_fillvalue())
---> 74     fld = np.ma.masked_equal(f_interp(z + radar.altitude['data'][0]),
     75                              get_fillvalue())
     76 

/miniconda3/envs/radar/lib/python3.6/site-packages/scipy/interpolate/polyint.py in __call__(self, x)
     76 
     77         """
---> 78         x, x_shape = self._prepare_x(x)
     79         y = self._evaluate(x)
     80         return self._finish_y(y, x_shape)

[...]

PR #750, opened 2018-06-01
Fix MaskedArray Error in gate_id.py #749

vlouf:
See  #749


scollis:
Thanks @vlouf !! Assigning @zssherman to do a quick review.

pyart/retrieve/gate_id.py
```diff
@@ -58,14 +58,18 @@ def map_profile_to_gates(profile, heights, radar, toa=None,
...
+        z = z.filled(np.NaN)
 
     # find toa is not provided
     if toa is None:
         ismasked = np.where(np.ma.getmaskarray(profile))[0]
         if len(ismasked) == 0:
             toa = None
         else:
-            toa = ismasked.min()
+            toa = ismasked.min()
```
zssherman: @vlouf The fix looks good to me! I would just suggest removing the whitespace after this line to be pep8 :)

PR #757, opened 2018-08-02
FIX: Fixed an issue of nexrad level 2 throwing an error when no msg5 …

scollis:
…present

Py-ART used msg5 data to set the target angle for the radar, fixed_angle in cf-radial.
Now for level2 data with no msg5 it will set all fixed angles to 0.0

I played with setting a warning.. probably a better way to do it -SC

### DIFF
--- a/pyart/io/nexrad_level2.py
+++ b/pyart/io/nexrad_level2.py
@@ -21,6 +21,8 @@ pyart.io.nexrad_level2
 
 """
 
+
+
 # This file is part of the Py-ART, the Python ARM Radar Toolkit
 # https://github.com/ARM-DOE/pyart
 
@@ -79,7 +81,7 @@ import struct
 from datetime import datetime, timedelta
 
 import numpy as np
-
+import warnings
 
 class NEXRADLevel2File(object):
     """
@@ -176,9 +178,15 @@ class NEXRADLevel2File(object):
 
         # pull out the vcp record
         msg_5 = [r for r in self._records if r['header']['type'] == 5]
+
+        wstring1 = 'No MSG5 detected. Setting to meaningless data \n'
+        wstring2 = 'Rethink your life choices and be ready for errors'
+
         if len(msg_5):
             self.vcp = msg_5[0]
         else:
+            # There is no VCP Data.. This is uber dodgy
+            warnings.warn(wstring1+wstring2)
             self.vcp = None
         return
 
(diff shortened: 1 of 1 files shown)


pyart/io/nexrad_level2.py
```diff
@@ -176,9 +178,15 @@ def __init__(self, filename):
 
         # pull out the vcp record
         msg_5 = [r for r in self._records if r['header']['type'] == 5]
+
+        wstring1 = 'No MSG5 detected. Setting to meaningless data \n'
```
zssherman: @scollis I would maybe remove these lines and added the message within the warnings.warn, see below.

pyart/io/nexrad_level2.py
```diff
@@ -176,9 +178,15 @@ def __init__(self, filename):
...
+        wstring2 = 'Rethink your life choices and be ready for errors'
+
         if len(msg_5):
             self.vcp = msg_5[0]
         else:
+            # There is no VCP Data.. This is uber dodgy
+            warnings.warn(wstring1+wstring2)
```
zssherman: I would put

```
warnings.warn("No MSG5 detected. Setting to meaningless data. "
                        Rethink your life choices and be read for errors.")

File: pyart/io/nexrad_level2.py [hunk 3]
```diff
@@ -426,7 +434,10 @@ def get_target_angles(self, scans=None):
         if scans is None:
             scans = range(self.nscans)
         if self._msg_type == '31':
-            cut_parameters = self.vcp['cut_parameters']
+            if self.vcp is not None:
+                cut_parameters = self.vcp['cut_parameters']
+            else:
+                cut_parameters = [{'elevation_angle' : 0.0}]*self.nscans
```
zssherman: for pep8 and consistency I would change remove the space before the colon and added spaces around the multiplication:
```
cut_parameters = [{'elevation_angle': 0.0}] * self.nscans
```

scollis:
Thanks @zssherman will do

pyart/io/nexrad_level2.py
```diff
@@ -426,7 +434,10 @@ def get_target_angles(self, scans=None):
         if scans is None:
             scans = range(self.nscans)
         if self._msg_type == '31':
-            cut_parameters = self.vcp['cut_parameters']
+            if self.vcp is not None:
+                cut_parameters = self.vcp['cut_parameters']
+            else:
```
zssherman: @scollis I would possibly move the warning here instead. It seems to only effect the fixed angle data, and pytest throws warnings on the testing nexrad file, but if the warning is moved here, it still catches and does what it is suppose to do, but doesn't throw the warning in the testing suite.

pyart/io/nexrad_level2.py
scollis: So no comma?

pyart/io/nexrad_level2.py
scollis: Actually I disagree. I want it when the MSG5 is found to me missing in case anyone adds anything else.

pyart/io/nexrad_level2.py
zssherman: No, the two statements are still printed together.

pyart/io/nexrad_level2.py
zssherman: Ah makes sense, yeah disregard.

scollis:
Commit meeting review waiting for tests to run

scollis:
Merging

PR #758, opened 2018-08-09
Recompile cython code with cython 0.28.4

ngoldbaum:
This allows me to build pyart on python3.7.

I mentioned to Bobby and Zach here at the sprint that you may want to think about adding cython as a build dependency so that these files are regenerated when you first clone the repo or when a .pyx file changes, so you don't need to recompile and check in the .c file to the repo. You can also set it up so that as part of the process of building the sdist the cython files get compiled so people who are installing the sdist from pypi don't need cython installed. Users who install wheels would also of course not need cython installed: only developers would need cython installed.

I was initially going to do that for pyary myself following yt's setup.py but it looks like pyart doesn't use setuptools so that would be harder than I thought it would be. I've opted to instead just update the cython code as-is and kick the can down the road.


scollis:
Very much appreciated @ngoldbaum! Assigning @zssherman to review. Zach can you pull this down and install and check gridding and unwrapping works (@rcjackson can help) 

CI passed so this is good 

We can open an issue to discuss generating .c on the file from pyx. When @jjhelmus included cython files cython was pretty new and including the C files insulated is against cython going away.. I would be interested in @jjhelmus's views on this

dopplershift:
I throw out there that netcdf4-python removed the `.c` file and only includes the `.pyx` in the repo. For some reason, if Cython went away, you could always go back to it.

Issue #743, opened 2018-05-21
Need to raise a warning in radardisplaycartopy when passed axes dont have projection


scollis:
See https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/radarmapdisplay_cartopy.py#L235

Twice I have wondered why radardisplaycartopy ignored passed axes... At the moment if the axes don't have a projection we override.. this can make users pull hair out... we need to raise a warning here

scollis:
Specifically this line here:
https://github.com/ARM-DOE/pyart/blob/master/pyart/graph/radarmapdisplay_cartopy.py#L262

This line needs a warning added before

scollis:
I am assigning to @jhemedin as his first Py-ART work
 Jason you will need to fork Py-ART to your account

Clone it

Make a branch to to work on 

Add in the required code

and then submit a Pull Request.. Consider this some training. I don't expect it to be done quickly :)

zssherman:
@jhemedin Is the issue fixed?

jhemedin:
@zssherman yes, PR #761 addressed this issue

zssherman:
Cool beans, closing.

PR #761, opened 2018-08-13
FIX: Added a warning messsage when an axes is defined without a projection

jhemedin:
This is in response to issue #743. A warning is now raised when a user doesn't define a projection when setting the axes when creating a map display. 

When an axes is set but no projection is defined, the program will override the defined axes and change it to the default along with a default projection. A user will now be warned when this happens.

### DIFF
--- a/pyart/graph/radarmapdisplay_cartopy.py
+++ b/pyart/graph/radarmapdisplay_cartopy.py
@@ -13,6 +13,8 @@ for drawing maps.
 
 """
 
+import warnings
+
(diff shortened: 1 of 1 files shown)


zssherman:
@scollis @jhemedin Rerunning travis test for 3.6 failed due to connection error with Travis. Should pass in a few minutes.

scollis:
Thank you @jhemedin and congratulations on becoming a Py-ART contributor

PR #764, opened 2018-08-24
ENH: Gatefiltered LP codes

scollis:
Added new methods in phase_proc that will accept a gate filter. I also cleaned
a lot of issues with carry through in the LP sub functions. Will need to
eventually depricate the normal codes and rename


scollis:
@zssherman and @rcjackson I will end up doing a documentation clean up in a future PR. The LP codes are a mess and need a good refactor.

scollis:
@rcjackson tested and no issues.. Merging

PR #765, opened 2018-09-03
Don't crash on broken nexrad NL2

striges:
Some nexrad L2 files have inconsistent gate number in their headers and actual sweeps. I cannot give an example at this time but I've encounter this issue many times in daily operational run when processing CDX/Batch scans. An easy but not perfect solution is to take whichever variable that gives minimal gate number to avoid a crash.

### DIFF
--- a/pyart/io/nexrad_level2.py
+++ b/pyart/io/nexrad_level2.py
@@ -536,8 +536,8 @@ class NEXRADLevel2File(object):
             msg = self.radial_records[msg_num]
             if moment not in msg.keys():
                 continue
-            ngates = msg[moment]['ngates']
-            data[i, :ngates] = msg[moment]['data']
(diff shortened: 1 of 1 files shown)


scollis:
Thanks @striges @zssherman please review

zssherman:
@scollis I ran it with nexrad level 2 and had no issues. ngates had no changes between both versions, but that was with a nexrad without the issue. Testing it on a file that has the error mentioned above might be ideal, but we would need to find one that fails. @striges Do you know a day and nexrad radar site that has a file that has had this error before? I can find the actual time from a nexrad boto loop that I've done before, but just need a day and site. Unless @scollis this seems like a reasonable fix and we don't have to go that far.

striges:
Try this

KGYX19990725_104721

scollis:
@zssherman did you try the timestep from @striges ?

scollis:
Good to go.. Thanks @striges

PR #767, opened 2018-09-06
FIX: Fixes an issue that, when python-coveralls is installed.

zssherman:
When python-coveralls is installed, the coverage dependency version is downgraded which causes issues with pytest-cov.


zssherman:
@scollis All checks are good. After this pull request, I'll get the micro release out.

PR #776, opened 2018-10-15
BUG: prevent IndexError in check_outliers

juhi24:
Previously, `boundary_conditions_maesaka` would crash with IndexError if `phi_near` was all zeros. The proposed fix prevents this by skipping the outlier check if `phi_near_valid` is empty.

I was wondering why `kdp_maesaka` was failing for most of my RHIs during batch recalculation of KDP. It seems this bug was triggered when there were only weak radar echoes.

### DIFF
--- a/pyart/retrieve/kdp_proc.py
+++ b/pyart/retrieve/kdp_proc.py
@@ -1501,10 +1501,12 @@ def boundary_conditions_maesaka(
 
     # check for outliers in the near range boundary conditions, e.g., ground
     # clutter can introduce spurious values in certain rays
-    if kwargs.get('check_outliers', True):
 
-        # do not include missing values in the analysis
-        phi_near_valid = phi_near[phi_near != 0.0]
+    # do not include missing values in the analysis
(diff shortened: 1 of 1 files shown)


scollis:
Thanks for the Pull Request @juhi24 @rcjackson can you grab A RHI and give this a spin.

scollis:
@kirknorth if you are still watching as well.

juhi24:
Here's one RHI that should trigger the exception https://b2drop.eudat.eu/s/rTwji7YQxT7WCFQ

juhi24:
Steps to reproduce:
```python
import pyart
radar = pyart.io.read('/path/to/201504221810_IKA.RHI_HV.raw')
rhi = pyart.retrieve.kdp_maesaka(radar)
```

zssherman:
The failing appveyor is unrelated to the pull request. Its a basemap python2.7 issue.

scollis:
Ok. Given @zssherman's assertion and given we hope to soon be rid of both Basemap and Python2.7 I will merge. Thank you @juhi24 !

PR #777, opened 2018-10-15
FIX: An attempt to fix basemap pyproj issue.

zssherman:
Install the packages from the conda-forge channel.


scollis:
Thanks @zssherman merging.

PR #778, opened 2018-10-18
ENH: Nearest neighbor in map_to_grid

rcjackson:
@zssherman and I had a good sprint this afternoon. We can now use the nearest neighbor weighting function in the cython version of the gridding code. This makes using that function a lot faster.

Also, we fixed an issue where the map_to_grid function was only using the Barnes weighting function no matter what the user specified.


scollis:
Some really nice work here. @rcjackson as discussed, since this changes some behavior  of Py-ART please notify the Py-ART users email list so folks know what to expect. Since this was worked on by our two main developers I am accepting this PR without an external review.. (or more so I reviewed it).

Issue #780, opened 2018-11-05
Add an `include_fields` argument to pyart.io.read


vlouf:
Currently the arguments exclude_fields is available in pyart.io.read() and pyart.io.read_grid().

The opposite would be nice when, for example, we only want to read the reflectivity in an heavy file containing lots of different fields, it would make the loading faster:

```python
radar = pyart.io.read(input_file, include_fields=['reflectivity'])  # Radar now only contains the reflectivity field
```

However, if you plan to switch to xarray as data model, then you probably won't need it.

scollis:
Feel free to ask @zssherman for help as well.

scollis:
I am happy to prioritize DIGR resources while @rcjackson is visiting @vlouf and Joshua

josephhardinee:
I thought PyART did lazy loading from Netcdf files which would mean no speed increase?

scollis:
@josephhardinee Its also useful for memory issues. If you know you are not going to use many of the vars you can exclude them from hard loading.. like C-SAPR2 :)

josephhardinee:
I mean I thought they were only loaded when  accessed (lazy load), or at least I thought I remember Jonathan implementing that a while back.

josephhardinee:
Or do you mean for binaries, load the fields but drop them before finalizing object which would hit memory initially, but then drop down in size after the load is complete.

rcjackson:
From what I read in the docs, pyart.io.read only does lazy loading if delay_field_loading is set to True, and that is not supported for every file format:

delay_field_loading : bool

True to delay loading of field data from the file until the ‘data’ key in a particular field dictionary is accessed. In this case the field attribute of the returned Radar object will contain LazyLoadDict objects not dict objects. Not all file types support this parameter.

Therefore, we would want something that is compatible for every file format.

PR #781, opened 2018-11-06
ADD: include_fields parameter

rcjackson:
In response to #780 raised by @vlouf I have added an include_fields parameter to pyart.io.read and pyart.aux_io.read functions. If it is set to None, all of the fields that are not in exclude_fields will be loaded. Otherwise, it will only load the fields that are in include_fields.

I have not tested it for any of the aux_io cases as we don't have unit tests set up for those, but it works for all of the example files in the Py-ART testing suite.

### DIFF
--- a/pyart/aux_io/arm_vpt.py
+++ b/pyart/aux_io/arm_vpt.py
@@ -18,7 +18,8 @@ from ..core.radar import Radar
 
 
 def read_kazr(filename, field_names=None, additional_metadata=None,
-              file_field_names=False, exclude_fields=None):
+              file_field_names=False, exclude_fields=None,
+              include_fields=None):
     """
     Read K-band ARM Zenith Radar (KAZR) NetCDF ingest data.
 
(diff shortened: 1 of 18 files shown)


scollis:
Looks like you are passing through to read_rsl which does not like it

rcjackson:
It's an input to FileMetadata in config.py along with exclude_fields (which most of the read routines use). The user was already allowed to supply exclude_fields to read_rsl. Should we just fix them both to be None (i.e. just loading all fields), and not allow the user to try to exclude fields with read_rsl?


Bobby

________________________________

zssherman:
@rcjackson I think the issue is just the parameter in read_rsl. There is no include_fields parameter defined in the read_rsl thats then fed into into the FileMetaData and, which looks like the cause of the failure. The reason appveyor and passed is because the rsl tests are skipped. My guess is a while back they were set to skip because rsl doesn't play nicely with Windows.

scollis:
I'm going to use this as a guinea pig to see if our CI is all good again.

scollis:
And the tests pass! Merging and thanks @rcjackson

PR #786, opened 2018-11-27
MAINT: Removal of basemap from core modules and some fixes.

zssherman:
I removed basemap from the core modules, which pyproj from basemap is being used in these modules. I set the code to just use proj from pyproj until I integrate cartopy projections into grid.py etc. Also fixed missing import in grid.py of an exception and added a temporary try statement to remove a deprecation warning in lazy_dict.py. The try statement will be removed once we remove python 2.7 support.

### DIFF
--- a/pyart/_debug_info.py
+++ b/pyart/_debug_info.py
@@ -99,10 +99,10 @@ def _debug_info(stream=None):
         cvxopt_version = "MISSING"
 
     try:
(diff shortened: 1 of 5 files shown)


scollis:
@rcjackson can you take a quick second eye over this.

scollis:
Ditto! Accepting despite the GDAL issue which is unrelated to this PR

PR #790, opened 2018-11-28
FIX: CI fix. explicit add of GDAL for py2.7

scollis:
Our CI breaks for Python 2.7 at the geotiff creation. This seems due to a dependency and packaging issue with GDAL. Current env seems to pull GDAL from a Cartopy dep. Now we try specifically installing GDAL

### DIFF
--- a/continuous_integration/environment-2.7.yml
+++ b/continuous_integration/environment-2.7.yml
@@ -4,6 +4,7 @@ channels:
   - defaults
 dependencies:
   - python=2.7
+  - gdal
   - numpy
   - scipy
   - matplotlib


kmuehlbauer:
@scollis Just a heads up: Eveyrtime you conda install something into this env it might pull in numpy from defaults. As it does in this case here too. You can circumvent that by adding  numpy (and sometimes scipy) with this conda install command.

scollis:
Thanks @kmuehlbauer ... I will fix that when it breaks. :) For me this is not only interesting for testing but also advising our users how to set up an install. 

Its funny that this is an issue in Py2.7 only. Perhaps related to, as you said, conda-forge not adding new 2.7 packages.. As I said in the email.. 2.7 days are numbered.

PR #791, opened 2018-11-28
MAINT: Change of graph modules to reflect favoring of Cartopy.

zssherman:
RadarMapDisplay now has cartopy mapping and basemap now is found in radarmapdisplay_basemap if user's still want to use it, but they will get a warning. Updated all unit tests and __init__.py file to reflect changes and also docstrings. GridMapDisplay is now GridMapDisplayBasemap, which will also throw a warning if user has basemap and wants to use. Working on a gridmapdisplay.py module that uses Cartopy. Also removed temporary cartopy fix from pull request #770 because this issue was fixed in Cartopy.

PR #793, opened 2018-11-29
New and updated plot_ppi examples for basemap and cartopy

jhemedin:
Updated the plot_ppi_basemap_with_rings.py with the RadarMapDisplayBasemap function and added a new example called plot_ppi_cartopy_with_rings.py which uses the updated RadarMapDisplay function with cartopy. 

Example will not work until PR #791 is accepted


scollis:
Very nice. Merging

PR #798, opened 2018-11-29
CI: Added cartopy install to appveyor adapted travis.yml for 3.7 and added 3.7 environment.

zssherman:
New 3.7 environment, does not test basemap, only cartopy. From recipe build is now 3.7 instead of 3.6. Cartopy install added to appveyor.yml.


zssherman:
Actually nvm, closing. From what I'm finding online only 3.7-dev is supported with travis, so for now will stick with normal build.

PR #804, opened 2018-12-12
ENH: Faster optimization loop in region based dealiasing + extra continuity term

rcjackson:
I have reimplemented the optimization loop in the region based dealiasing to be faster and to have an extra term that adds a cost for correcting a region too far from the one with the closest mean velocity (i.e. it keeps regions of similar velocity together to better handle cases that may deviate far from the sounding).

### DIFF
--- a/pyart/correct/region_dealias.py
+++ b/pyart/correct/region_dealias.py
@@ -224,19 +224,14 @@ def dealias_region_based(
         # anchor unfolded velocities against reference velocity
         if ref_vdata is not None:
             sref = ref_vdata[sweep_slice]
-            #mean_diff = np.ma.mean(sref - scorr)
-            
-            #global_fold = round(mean_diff / nyquist_interval)
-            #if global_fold != 0:            
-            #    scorr += global_fold * nyquist_interval
 
             # Anchor specific regions against reference velocity
             # Do this by constraining cost function due to difference
             # from reference velocity and to 2D continuity
             new_interval_limits = np.linspace(scorr.min(), scorr.max(), 10)
             labels_corr, nfeatures_corr = _find_regions(
-                scorr, sfilter, new_interval_limits) 
-            
+                scorr, sfilter, new_interval_limits)
+
             bounds_list = [(x, y) for (x, y) in zip(-6*np.ones(nfeatures_corr),
                            5*np.ones(nfeatures_corr))]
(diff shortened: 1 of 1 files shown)


vlouf:
@scollis No I don't.

scollis:
Looks like usual problem @rcjackson .. 
When setting tolerances think about physical reasonableness. precision varies by machine

rcjackson:
Nah, I had enforced the continuity term a bit too strong...when I tested it I ran test_dealias.py instead of test_region_dealias.py. Easy to get those two confused!

rcjackson:
I mainly did this to make this code more usable for dealiasing the XSAPR sector scans. My goal is to hopefully have an example of those in my PyDDA talk for AMS (like I had in the abstract), but if not I have other stuff too.

scollis:
Testing ftw 

-sent from a mobile device-

scollis:
@zssherman can you check this works

zssherman:
@scollis Sounds good, I will check it.

pyart/correct/region_dealias.py
```diff
@@ -162,7 +162,7 @@ def dealias_region_based(
     # perform dealiasing
     vdata = radar.fields[vel_field]['data'].view(np.ndarray)
     data = vdata.copy()     # dealiased velocities
-
+    mean_vel_last = np.nan
```
zssherman: From what I'm seeing, this isn't used anywhere.

pyart/correct/region_dealias.py
```diff
@@ -445,20 +467,30 @@ def _cost_function(nyq_vector, regions, vels_slice,
...
+
+        # Regional continuity
+        vels_without_cur = np.delete(vels_slice_means, reg)
+        diffs = np.square(vels_slice_means[reg]-vels_without_cur)
+        the_min = np.argmin(diffs)
+        add_value2 = (vels_slice_means[reg] +
+                      np.round(nyq_vector[i])*v_nyq_vel -
+                      vels_without_cur[the_min])
+
+        if(np.isfinite(add_value2)):
```
zssherman: I believe there is 3 of these extra parenthesis throughout the code, I would just make it:
if np.isfinite(add_values2)

scollis:
Restarting AppVeyor

scollis:
Thanks all! Merging

Issue #772, opened 2018-10-02
fixed_angle for NEXRAD with negative elevation


striges:
NEXRAD now enables low-level tilts scan below 0.5 in SAILS/MRLE/N0R,  even a negative deg at -0.2. pyart will wrap it to 359.8, which may cause issue when sorting tilts by elevations, and actually read odd.

```
// global attributes:
                :_NCProperties = "version=1|netcdflibversion=4.6.1|hdf5libversion=1.10.1" ;
                :comment = "359.7803,359.7803,0.4834,0.4834,0.8789,0.8789,1.3184,1.3184,359.7803,359.7803,1.8018,2.4170,3.1201,3.9990,5.0977,6.4160,7.9980,10.0195,11.9971,14.0186,16.6992,19.5117" ;
                :scan_id = 215 ;
                :site_name = "KMAX" ;
                :title = "" ;
                :original_container = "NEXRAD Level II" ;
                :source = "" ;
                :version = "1.3" ;
                :instrument_name = "" ;
                :references = "" ;
                :vcp_pattern = 215L ;
                :volume_start_time = "20181002173057" ;
                :Conventions = "CF/Radial instrument_parameters" ;
                :institution = "" ;
                :field_names = "cross_correlation_ratio, differential_reflectivity, reflectivity, differential_phase" ;
                :history = "" ;
data:

 fixed_angle = 359.7803 ;
```

striges:
I am working on a patch now as it is impacting operational run, I may send a pull request once I finish verification.

scollis:
Much appreciated!



Scott Collis


On Oct 2, 2018 at 1:17 PM, striges <notifications@github.com> wrote:


I am working on a patch now as it is impacting operational run, I may send
a pull request once I finish verification.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ARM-DOE/pyart/issues/772#issuecomment-426378016>, or mute
the thread

nguy:
Anything new on this issue?

scollis:
Sorry! Life has been complex. Hey @jhemedin This is yours now! That's what you are doing to start the week First thing: Grab the KAMX file for 20181002173057 and take a look

nguy:
What, it's not like you've been to Alaska or something...oh wait!

FYI, KMUX has quite a few of these as well in case you need more test cases.

scollis:
[...]

PR #809, opened 2019-02-06
FIX: added argument to deal with negative fixed_angles

jhemedin:
Temporary fix to issue #772. Before 360 would be added to any negative angle and would read funny, now if any angle is above 45 degrees 360 will be subtracted. For example a fixed angle of -0.2 use to be outputted as 359.2, now it will be output as -0.2. Will be testing to get a more permanent fix for a future PR.


rcjackson:
Looks like Travis goofed in trying to download Python, so I would restart the build. Otherwise, it worked with the test case, and the CIs that actually downloaded Python passed.

nguy:
Any plans to rerun the CI and see if this gets merged?

scollis:
Yeah.. I think I know what is up with CI.. I will merge this

PR #811, opened 2019-02-15
Masked array check in simulated_velocity

rcjackson:
This fixes an error I was getting when using the newest scipy for interpolating the sounding to gates regarding using masked arrays. Now there is a check to see if there are masked arrays before doing the interpolation.


scollis:
I need to fix testing on py 3.4 and 3.5.. as it passes for 3.6 I am accepting

PR #812, opened 2019-02-15
BUG: Region based dealiasing cost function would crash when there was only one region

rcjackson:
I am fixing an issue where the sounding constraint on the region based dealiasing would crash if there was only one region.


scollis:
Thanks @rcjackson !

scollis:
Versioning issues on 3.4
```pkg_resources.VersionConflict: (pytest 3.2.2 (/home/travis/miniconda3/envs/testenv/lib/python3.4/site-packages), Requirement.parse('pytest>=3.6'))```

and for some reason pytests is segfaulting on 3.5 (!) 

3.6 works and since this is a minor fix I am going to accept

PR #817, opened 2019-03-04
Atteunation correction fixes

rcjackson:
I have done the following changes to Py-ART that needed to be done in the course of running CMAC2.0 on the X-SAPR data:

* Reorgainzed documentation for the MeteoSwiss attenuation correction codes and added a gatefilter to each of them.
* There was an issue where the phase processing code would crash if the freezing level was below the radar. For the LP phase processing, when the freezing level is below the radar it will default the mask to the smallest possible window size that can be provided to the filter in the LP processing code (6th gate in the ray and closer).


scollis:
3.4 failing but I know that is an issue with out testing more than anything else
Merging

PR #818, opened 2019-04-02
CI: Fix for Python 3.4 issue with pytest. 

zssherman:
pytest-cov needs to be lower then version 2.6 to work with pytest and python 3.4.


scollis:
Thankyou Thankyou Thankyou! I LOVE green ticks!

Issue #819, opened 2019-04-02
Possible bug plotting CSAPR-2 HSRHIs


swnesbitt:
In plotting CSAPR-2 HSRHIs from CACTI, I am having issues with plotting RHIs at 90° azimuth with `display.plot_rhi()`.  This behavior is repeatable amongst many files, but only at this azimuth angle.  Seems like a geometry issue.  
Example is here:
![junk](https://user-images.githubusercontent.com/3288592/55437035-c1e1f100-5552-11e9-9ccb-9287bac036d7.png)
Interestingly, `radar.get_gate_x_y_z(sweep=3)` seems to extract ok values:
![get_gate](https://user-images.githubusercontent.com/3288592/55436859-3c5e4100-5552-11e9-8dcd-21cb5b3a6099.png)
I'll keep digging...

scollis:
Thanks Steve. Well look into it 

-sent from a mobile device-

swnesbitt:
Yeah, the calculation for the x-axis value in the function `plot_rhi()` has a bug:

<img width="430" alt="Screen Shot 2019-04-02 at 2 39 14 PM" src="https://user-images.githubusercontent.com/3288592/55438003-49306400-5555-11e9-9001-05e68109e99a.png">

scollis:
What's strange is that it is working for some elevations past 90 degrees where sign(y) = -1.0

Can you plot the elevations for that azimuth?

swnesbitt:
Ask and you shall receive
<img width="389" alt="Screen Shot 2019-04-02 at 2 53 05 PM" src="https://user-images.githubusercontent.com/3288592/55438758-636b4180-5557-11e9-8bba-dbfb7634d1ef.png">

scollis:
grabbing data

scollis:
ok @swnesbitt I have the dodge to end all dodges why we patch this:
```
radar_file = '/data/CACTI/csapr2/corcsapr2cfrhsrhiM1.a1.20181126.120715.nc'  
radar = pyart.io.read(radar_file)
radar.azimuth['data'] += .01
my_d = pyart.graph.RadarDisplay(radar)
my_d.plot_rhi('reflectivity', 3)
```

scollis:
@zssherman please add a line that, when abs(radar.fixed_angle['data'][sweep] == 90) (or y.all == 0) use sign(x)

scollis:
Thank you @swnesbitt This was a nice find. Old code is, old code..

[...]

PR #820, opened 2019-04-03
FIX: A fix suggested by @scollis for the bug in rhi plotting found by @swnesbitt

zssherman:
A fix for issue #819 @scollis @swnesbitt.


scollis:
Thanks @zssherman !

PR #821, opened 2019-04-08
Master

tjlang:
On some user systems, using single instead of double quotes with the
-t_srs keyword with gdalwarp will cause an error. This update adds a
use_doublequotes Boolean keyword to the write_grid_geotiff function,
which allows the end user to address this error when they have their
system call gdalwarp from the command line (i.e., warp=True). See
updated documentation for usage.


tjlang:
This fix addresses the following issue:
https://groups.google.com/forum/#!topic/pyart-users/BvFNaNKoJXg

tjlang:
AppVeyor build failed?

scollis:
Could be all kinds of issues @zssherman look into it

zssherman:
@tjlang @scollis From digging around, looks like in the appveyor build, possibly the wrong python path is being used, appveyor.yml might need a revamp. I'll see what I can do to fix it. I was wondering too, do we want the double quotes as an argument or maybe have it as a try and except error?

tjlang:
@zssherman I am concerned with that option because the error actually occurs with gdalwarp, not with Python. Since I couldn't reproduce the user's error, and had to rely on their word on what was happening, it seemed safer to just have it as keyword option in the function.

zssherman:
@tjlang Ah gotcha, makes sense. We can go the argument route.

scollis:
restarted CI

scollis:
@zssherman its failing on 3.4 because we are not testing on 3.4 .. might just be a travis config issue right?

zssherman:
@scollis, yeah you might have to accept it. Then after the doc pull request runs, can accept that one and we'll be all green light

PR #824, opened 2019-04-11
MNT: Removed fortran from pyart.

zssherman:
With the new python version of the echo steiner code. We can remove the fortran code. This will help with compiling etc. Unit tests needs to be tweaked, so no merge until then and review.


zssherman:
@vlouf I'm trying to switch your code over to the default echo class code and remove fortran, but having a test fail. I'm trying to pinpoint where the zeros are coming from, do you possibly know what would cause this failure in the calculation?

scollis:
Thanks @zssherman merging

PR #828, opened 2019-04-19
ADD: new GridMapDisplay class that uses Xarray and Cartopy.

jhemedin:
Changed the GridMapDisplay class to use Xarray and it's plotting functions as well as using Cartopy for getting a basemap  background. Also created unit test for function.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,7 +31,7 @@ install:
   - "activate matrix_env"
 
   # Install Py-ART dependecies
-  - "conda install -y -q numpy scipy matplotlib netcdf4 pytest pytest-cov cython"
+  - "conda install -y -q numpy scipy matplotlib netcdf4 pytest pytest-cov cython cartopy xarray"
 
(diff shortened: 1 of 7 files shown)


zssherman:
@rcjackson Hey, I added you as a reviewer if you have time. I suggested PEP8 changes, fixes, etc. Everything looks good to me, but just wanted a second set of eyes.

jhemedin:
The get_dataset just creates an xarray dataset that is used for plotting. I can add a docstring if you think it is necessary.

rcjackson:
For now we can just keep it as is.


Bobby

________________________________

zssherman:
@scollis When you get this, seeing as your on travel, but once this is merged, I'll begin the release process.

scollis:
In flight Wi-Fi ftw 

-sent from a mobile device-

zssherman:
Hmm can't figure out why 3.5 is failing...

PR #831, opened 2019-04-22
Force typing to match sweep indices elsewhere.

nguy:
`radar.extract_sweeps` changes the `sweep_start_ray_index` and `sweep_end_ray_index` datatype from int32 to int64. To keep consistent with the rest of the PyART library, the array['data'] numpy arrays are typed to int32


scollis:
Thanks Nick! 

-sent from a mobile device-

PR #832, opened 2019-04-22
FIX: Skip empty scans and incomplete MSG5's

rcjackson:
This addresses #826 which was caused by incomplete message 5's and extra empty scans that were written in the file. Py-ART will now simply skip over both of these and throw a warning if there is an odd-sized MSG5.

### DIFF
--- a/pyart/io/nexrad_level2.py
+++ b/pyart/io/nexrad_level2.py
@@ -594,6 +594,9 @@ def _get_record_from_buf(buf, pos):
             warnings.warn("Encountered incomplete MSG5. File may be corrupt.",
                           RuntimeWarning)
             new_pos = pos + RECORD_SIZE
+    elif msg_type == 29:
+        new_pos = _get_msg29_from_buf(buf, pos, dic)
+        warnings.warn("Message 29 encountered, not parsing.", RuntimeWarning)
     elif msg_type == 1:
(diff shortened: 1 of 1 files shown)


scollis:
Thanks Bobby! 

-sent from a mobile device-

rcjackson:
@dopplershift let us know that there is a special meaning to a header size of 65535 that was not factored into our code when message 29 came up. I added a routine to just parse the message size and then skip over it.

pyart/io/nexrad_level2.py
```diff
@@ -584,14 +587,30 @@ def _get_record_from_buf(buf, pos):
...
-        new_pos = _get_msg1_from_buf(buf, pos, dic)
+        new_pos = _get_msg1_from_buf(buf, pos, dic)       
     else:   # not message 31 or 1, no decoding performed
         new_pos = pos + RECORD_SIZE
 
     return new_pos, dic
 
+def _get_msg29_from_buf(buf, pos, dic):
+    msg_size = dic['header']['size']
+    if msg_size == 65535:
```
dopplershift: Just to be clear, while this format change was made to accommodate message type 29, there's nothing in the NEXRAD ICD that indicates that this special sizing only applies to message 29. It's just that 29 is the only data we get currently that seems to use this.

PR #837, opened 2019-04-26
Fix for reading ECCC ODIM files

wxd00d:
Current ODIM reader assumes the number of bins between sweeps is constant, as long as range start and scale are constant between sweeps. This doesn't hold true for new data from ECCC S-Band radars, range can be variable between sweeps. Proposed fix checks all sweeps for the `nbins` attribute, and finds the max.  Axis 1 of the numpy array for field data `fdata` is set to `max_nbins` rather than just `nbins` read from the first sweep. If a sweep has less bins than the max, these are set to `NaN`s. File is attached for testing.
[2018050823_12_ODIMH5_PVOL6S_VOL_CASRA.h5.zip](https://github.com/ARM-DOE/pyart/files/3122169/2018050823_12_ODIMH5_PVOL6S_VOL_CASRA.h5.zip)

### DIFF
--- a/pyart/aux_io/odim_h5.py
+++ b/pyart/aux_io/odim_h5.py
@@ -335,7 +335,7 @@ def read_odim_h5(filename, field_names=None, additional_metadata=None,
             sweep_nbins = sweep_data.shape[1]
             fdata[start:start + rays_in_sweep, :sweep_nbins] = sweep_data[:]
             # set data to NaN if its beyond the range of this sweep
-            fdata[start:start + rays_in_sweep, sweep_nbins:max_nbins].data = np.nan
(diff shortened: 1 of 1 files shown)


zssherman:
Hi @wxd00d, thank you for the pull request! I did get a attribute error with the code, but if you change the line:
fdata[start:start + rays_in_sweep, sweep_nbins:max_nbins].data = np.nan
to:
fdata[start:start + rays_in_sweep, sweep_nbins:max_nbins] = np.nan
It works.
odim_ref

wxd00d:
Ok, I've made the requested changes.
Thanks!
-Brandon

zssherman:
@scollis This checks out for me, does this look good to you?

scollis:
Yep. Thanks @wxd00d !!

PR #838, opened 2019-05-03
MNT: Doc cleanup, removal of unused variables, added docstrings. raise error fix and more.

zssherman:
I went through code and cleaned up most of the docs. I also fixed in kd proc a raise value error that was using a incorrectly name variable. Added some functions that were missing from __init__ and top of module doc strings, so they should now appear in the generated docs. PEP8 changes, such as fix of literal comparisons,  whitespace, unused variables function, wrong continuation indents and more. Added optional to the docs of optional parameters in functions that were missing it. Instead of check python version for imp module, changed to try and except.

Note: I haven't gotten to the graph or aux_io module yet. But I can do a separate pull request for that so this isn't larger then it already is.


scollis:
Looks good.. We good to merge @zssherman ?

zssherman:
@scollis Yeah this is good to go.

PR #840, opened 2019-05-06
FIX: A fix for a bug in grid io which had dimensions out of order.

zssherman:
In write point function, the z, y, x dimensions were flipped which was
different then the original data. NetCDF4 new broadcasted_shape function
caught this.


zssherman:
@scollis Looks like with netCDF4 new release and the fix, we are green lights again

Issue #808, opened 2019-01-25
Adding radar.get_gate_longitude() instead of using attributes radar.gate_longitude 


aschueth:
I am trying to map out the lat and lons for a single sweep. With radar.get_gate_x_y_z() I can specify the sweep and it will return the x,y, and z arrays for each bin in the sweep, however there is not a function like that for lat or lon. I have been trying to use the attributes, but it includes all the sweeps and is challenging to slice the data I want. Is there any way to add this?

scollis:
Yep. This is a good idea. At the moment the team is flat out so I will come back to this later

dara74k:
Hi,
Is it possible to be able to save the (lon, lat, radarvariable) as csv also?
Thanks

zssherman:
Has been addressed in #841 closing, and at @dara74k I know this is a late response, but I recommended looking into pandas. You can create and dataframe and add radar['latitude']['data'] the longitude and variable to a dataframe and then out the dataframe with dataframe.to_csv. There is more in the pandas documentation.

PR #841, opened 2019-05-07
ENH: Adding get_gate_lat_lon_alt function.

zssherman:
Mentioned in #808 by @aschueth, function allows the user to specify a sweep and can then obtain the gate latitude, longitude and altitude for that sweep.

### DIFF
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -516,7 +516,8 @@ class Radar(object):
         return antenna_vectors_to_cartesian(
             self.range['data'], azimuths, elevations, edges=edges)
 
-    def get_gate_lat_lon_alt(self, sweep, filter_transitions=False):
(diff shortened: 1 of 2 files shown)


zssherman:
Will commit unit tests shortly.

scollis:
Quick thought @zssherman : I wonder if we should have an option in the code to force the recalculation of the underlying data.. Right now if a user alters geometry (eg radar alt) you need to call a method of radar to force recalc

zssherman:
@scollis I added the reset coords function that will recalculate the gate coords as an option.

scollis:
@rcjackson any comments?

rcjackson:
This all looks good to me. Having the altitude and lat and lon in one function is certainly very helpful.

scollis:
Thanks @zssherman !! Merging

PR #846, opened 2019-05-22
ENH: Added argument in io/grid_io.py and core/grid.py  

jhemedin:
When creating a grid object and saving it, a user now has an option to add the ARM standard alt, lat, lon variables if they want them added. Works the same as adding the ARM standard time variables.


scollis:
super nice

zssherman:
@jhemedin @scollis This also looks good to me, had to rerun travis though, wasn't showing complete for some reason.

scollis:
Looks good.. merging

PR #856, opened 2019-07-24
FIX: Fixes colorbar issue for HomeyerRainbow colormap

zssherman:
Caused by difference of floats and integers subtraction between python versions.


scollis:
Great work.. merging

Issue #863, opened 2019-08-08
'constant_roi' argument should override 'roi_func' in gates_to_grid.py


vlouf:
In gates_to_grid.py there are 2 arguments:
`roi_func='dist_beam', constant_roi=500.,`

Right now, if you define a constant_roi, the default `roi_func` is still 'dist_beam' and so the constant_roi that is defined is useless. Defining a constant_roi should switch automatically the roi_func to 'constant'.

cc. @robwarrenwx 

I propose to change the default of constant_roi to None: `constant_roi=None` and then in the function `map_gates_to_grid` to have something like that:
```python
if constant_roi is not None:
    roi_func = 'constant'
else:
    constant_roi = 500
```

scollis:
Closed by #864

PR #864, opened 2019-08-12
MNT: Suggested changes of automatically setting roi_func to constant if constant_roi is used.

zssherman:
As suggested by @vlouf in #863 , roi_func uses constant if constant_roi is not None.


scollis:
@rcjackson take a look at this and advise if we can accept.. looks like a sensible change

rcjackson:
Looks good to me!

scollis:
Thanks @rcjackson and @zssherman and thanks @vlouf for the issue

PR #865, opened 2019-08-23
FIX: ODIM reader with nbins being an array

scollis:
Some how reading South African ODIM files yields an array for max_nbins.. doing a check for that..


scollis:
tests passed.. And since ODIM is on the Aux io module I am merging..

PR #868, opened 2019-10-01
DOC: Adding some statements on Anaconda compilers and some FAQs from pyart google groups.

zssherman:


zssherman:
Tests passed merging.

PR #873, opened 2019-10-23
ENH: Added imshow as option for plot_grid and fixed bugs in to_xarray.

zssherman:
Added imshow to plot_grid, need to work on way to get it in plot_longitudinal etc,, but figure pull request now for the fixes. Mask outside wasn't being used in gridmapdisplay ode, so fixed that. to_xarray coords are fixed.


zssherman:
Merging, test fail unrelated to code, will be added x and y plotting in separate pull request.

PR #876, opened 2019-10-28
FIX: Fixes fill value issue with netCDF4.

zssherman:
After discussion, assigning the fillvalue to be in a specific order is removed as not seeing it as needed. Should fix error with unit tests.


zssherman:
Everything passes, merging.

PR #878, opened 2019-11-11
FIX: Crash when only one region found in optimization loop for region…

rcjackson:
@scollis found a bug when dealiasing using the sounding technique where the sounding correction would crash if only one region was found. I went ahead and placed a special case when we only have one region to examine and anchor against the sounding. For this case, it will just check to see if the mean velocity from the region based correction deviates from the sounding by less than the Nyquist interval. If it does not, it will adjust the sweep by the amount of Nyquist intervals it is off by.


scollis:
only failing on 2.7.. merging

PR #879, opened 2019-11-11
CI: Removing basemap as dependency in unit tests, support ends soon 2020, fails CI with 2.7.

zssherman:
Basemap plotting code will still exist in pyart, but there will be no support for it.


zssherman:
@scollis if your cool with this, I'll merge

scollis:
Yes.. please do..

I will send communications on end of life for 2.7 soon

PR #886, opened 2019-12-17
FIX: Fixes possible geos update that broke point calculation for lambert ticks.

zssherman:
I honestly had an issue trying to pinpoint the origin of the issue, most like geos as shapely and cartopy have not had recent updates that relate. In a nutshell, from testing, it seems that geos class returned changed from GEOMETRYCOLLECTION to LINESTRING, and when empty, one has and one doesn't seem to have an xy attribute. Regardless the value return to is None, so
if an empty class appears, None is appended to the ticks list as originally intended in the old code.


zssherman:
Side note, once there is a new release of cartopy, should be able to axe any work around projection tick code.

scollis:
Great work @zssherman

scollis:
Just waiting on Py2.7 on Appveyor... I have a meeting to get to so Merging :)

PR #889, opened 2020-01-01
PY: Removing Python 2.7 support from CI and documentation.

zssherman:


zssherman:
Its working, merging, pull request to follow with removal of from future imports etc.

PR #890, opened 2020-01-03
PY: Removing all futures relating to Python 2.

zssherman:


guides/contributors_guide.rst
```diff
@@ -38,15 +38,15 @@ please cite:
...
-cited if you make use of them.  Refer to the **References** section in the
+cited if you make use of them. Refer to the **References** section in the
 documentation of the functions used for information on these citations.
 
 
 Install
 -------
 
 The easiest method for installing Py-ART is to use the conda packages from
-the latest release.  To do this you must download and install
```
rcjackson: Perhaps comment here that Python 2.7 is no longer supported so that we don't get people trying to run this on Python 2.7. I'd make the minimum recommended version to be Python 3.6+.

zssherman: Sounds good @rcjackson , I will add some info in the docs clarifying 2.7 is no longer supported.

zssherman:
All good, merging.

PR #891, opened 2020-01-04
fix(xarray): Gate xarray so its optional

zflamig:
Optional xarray makes it easier to use pyart in space constrained environments


zssherman:
Hi @zflamig thanks for catching this and making the changes! Looks good to me, merging.

PR #893, opened 2020-01-14
DOC: Update to docs to use rtd theme and remove python 2 from docs.

zssherman:
example2
example1

Still going to do minor edits.


zssherman:
example3

zssherman:
Still need to do some minor edits, so not merging yet.

zssherman:
@rcjackson  whenever you get time, know your busy with AMS. The docs now pass, but there are alot of warnings that I'm seeing. Not sure if you've seen these before?

rcjackson:
It looks like you have .rst files that are not being referenced to in the documentation. I would double check to ensure that you're not missing any of those modules.

zssherman:
Hmm all the rst files seem to be there in the source and generated directory, which is why I'm a little lost.

zssherman:
But the warning seems to be for every single file, which almost seems like a format issue or something, but not sure honestly.

doc/source/index.rst
```diff
@@ -2,11 +2,9 @@
 The Python ARM Radar Toolkit - Py-ART
 =====================================
 
-Latest Release
-==============
-
 .. toctree::
-    :maxdepth: 4
+    :maxdepth: 1
```
rcjackson: I think it's whining about this. Try setting it back to 4.

zssherman: Gotcha, will try that

zssherman: @rcjackson Dang, sadly did not work.

rcjackson: If you try to access one of those functions in the docs can you still get to it? I just would like to know how important these warnings actually are, so this will tell me.

zssherman: Yeah it seems I can navigate through all the functions and their attributes and everything are there. One thing I notices is params show up as gate_yarray instead of gate_y: array so wondering if the old docs all colons need a space removed from before the colon

rcjackson: In PyDDA I only put a space after the colon, not before. So that might be causing it to not work with this theme. I recall we ran into something similar with ACT.

zssherman: Gotcha, I'll go through and change all of that.

zssherman:
I thin this is almost done. Most if not all warnings are fixed, examples are up to date. The only thing I can't figure out is why the custom Py-ART exceptions are not documenting. My only guess is the _templates aren't designed for how those classes are set up.

zssherman:
@rcjackson @scollis Think this is as good as I can get it, if you both want to review one last time.

rcjackson:
From what you have showed me, this should be good to go.

zssherman:
Cool, merging.

PR #903, opened 2020-02-10
ENH: Updated nexrad for RDA build 19.0.

zssherman:
I think I got it. The only thing is unit test is based on RDA builds before 19.0, will need to redo the unit tests as it assert values from an older nexrad file from testing directory.

### DIFF
--- a/pyart/io/nexrad_level2.py
+++ b/pyart/io/nexrad_level2.py
@@ -513,10 +513,10 @@ class NEXRADLevel2File(object):
         nrays = len(msg_nums)
 
         # extract the data
-        if moment != 'PHI' or moment != 'ZDR':
-            data = np.ones((nrays, max_ngates), dtype='u1')
-        else:
+        if moment == 'PHI' or moment == 'ZDR':
             data = np.ones((nrays, max_ngates), dtype='u2')
+        else:
+            data = np.ones((nrays, max_ngates), dtype='u1')
         for i, msg_num in enumerate(msg_nums):
             msg = self.radial_records[msg_num]
             if moment not in msg.keys():
--- a/pyart/io/nexrad_level2.py
+++ b/pyart/io/nexrad_level2.py
@@ -511,20 +511,20 @@ class NEXRADLevel2File(object):
         # determine the number of rays
         msg_nums = self._msg_nums(scans)
         nrays = len(msg_nums)
-
         # extract the data
-        if moment == 'PHI' or moment == 'ZDR':
-            data = np.ones((nrays, max_ngates), dtype='u2')
-        else:
-            data = np.ones((nrays, max_ngates), dtype='u1')
+        set_datatype = False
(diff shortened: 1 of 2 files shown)


pyart/io/nexrad_level2.py
```diff
@@ -621,11 +622,11 @@ def _get_msg31_data_block(buf, ptr):
         dic = _unpack_from_buf(buf, ptr, ELEVATION_DATA_BLOCK)
     elif block_name == 'RAD':
         dic = _unpack_from_buf(buf, ptr, RADIAL_DATA_BLOCK)
-    elif block_name in ['REF', 'VEL', 'SW', 'ZDR', 'PHI', 'RHO']:
+    elif block_name in ['REF', 'VEL', 'SW', 'ZDR', 'PHI', 'RHO', 'CFP']:
         dic = _unpack_from_buf(buf, ptr, GENERIC_DATA_BLOCK)
         ngates = dic['ngates']
         ptr2 = ptr + _structure_size(GENERIC_DATA_BLOCK)
-        if block_name == 'PHI':
+        if block_name == 'PHI' or block_name == 'ZDR':
```
dopplershift: There's a field in the message that explicitly states the size of each bin in bits. Could you not just use that rather than hard-coding it here, which just puts you back in the same position if they update again?

zssherman: That makes sense, and sounds good to me. I will do some changes. Thanks for the feedback!

zssherman: @dopplershift Just so I got this right as I'm starting to confuse myself lol, still set dtype for the moments, but use something like int16 and not hard coding to set a specific moment?

dopplershift: Not quite. Here's a section of the specification on the Data Block:

image

Note the field "Data Word Size", it can (currently at least) have a value of 8 or 16. I'm proposing using that value to decide which dtype code is appropriate. [This](https://github.com/Unidata/MetPy/blob/40d5c12ab341a449c9398508bd41d010165f9eeb/src/metpy/io/_tools.py#L313-L321) is the code we use to do that in MetPy.

zssherman: Ahhh gotcha, that makes sense. Yeah the bit of code in that link helps clear things up. Thanks for helping clarify that!

dopplershift: No problem!

zssherman:
@dopplershift I did some changes, but honestly struggled with figuring out how to set the datatype as the data arrays were set before the msg call so not sure if my changes are the best way around it, but when you have time, if you suggest another way around it. I was able to find the word size for message 31 and was able to use it to set the data type instead of how it use to hard code dtypes using field names. Seemed to work with both older formats and the new format with have uint8 for ZDR in older files and uint16 for the new file with really no changes in performance.

zssherman:
Merging.

PR #911, opened 2020-03-13
ADD: Adding a basic ingest using Py-ART's testing object example notebook.

zssherman:
Title says it all, I know some users have netCDF4 data that isn't sometimes in cfradial or format for Py-ART to read, so this is an example on ways around that. Ideally a ingest added to the io or aux module would be preferred for changes in formats etc, but sometimes that is not realistic and that data might be unique.


zssherman:
@rcjackson I added you as a reviewer and since this is a notebook, here is a link to the actually generated notebook to review:

rcjackson:
I feel that this example should be in the sphinx docs as a tutorial on making custom Radar objects from noncompliant datasets. That way, people would more readily see it and I think it would look better that way. Would that be too much for you to integrate this over to the docs?

zssherman:
I agree, I'll read up on how to do that.

zssherman:
@rcjackson I figured out how to generate notebooks, so I did the changes, here is a screenshot.

Screenshot from 2020-03-18 14-06-14

rcjackson:
It looks like a recent update to cftime is breaking the CI… I would make a separate PR fixing that issue next.
https://github.com/Unidata/cftime/issues/143

zssherman:
Gotcha I'll take a look at the cftime changes

zssherman:
Tests pass now, I will add more notebooks in a separate pull request, so merging for now.

Issue #910, opened 2020-03-03
Link to MPL in setup.py


scollis:
See this tweet:
https://twitter.com/story645/status/1234869245449818112
Hey folks who maintain libraries built on top of 
@matplotlib
 that are packaged on 
@pypi
, I need a huge to me favor. 
If you'd include a 'Framework::Matplotlib' classifier in your

 http://setup.py if it exists, please comment at

scollis:
@zssherman can you look after this

zssherman:
Yep, I'll do the changes.

zssherman:
This was done in #912 closing.

PR #912, opened 2020-03-18
MNT: Adding MPL framework classifier, addresses #910.

zssherman:
@scollis , the test will probably fail, but its not related, it due to new cftime release.


zssherman:
Tests pass, merging

PR #913, opened 2020-03-20
FIX: A fix for the update in one of the datetime modules that broke CI

zssherman:
At the moment isoformat attribute is not found, so setting isoformat using strftime. Also set calendar in fetch sonde time.

### DIFF
--- a/examples/plotting/plot_rhi_cfradial.py
+++ b/examples/plotting/plot_rhi_cfradial.py
@@ -12,8 +12,8 @@ print(__doc__)
 # Author: Jonathan J. Helmus (jhelmus@anl.gov)
 # License: BSD 3 clause
 
-import netCDF4
 import matplotlib.pyplot as plt
+import netCDF4
 import pyart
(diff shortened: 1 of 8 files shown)


rcjackson:
CI passes, I'm merging.

vlouf:
How did you fix that? It's a problem on my end with a lot of scripts relying on cftime. I had to change from
`radar_start_time = netCDF4.num2date(radar.time['data'][0], radar.time['units']).isoformat()`
to:
`radar_start_time = cftime.num2date(radar.time['data'][0], radar.time['units'], only_use_cftime_datetimes=False, only_use_python_datetimes=True).isoformat()`

zssherman:
@vlouf I did a temporary fix using strftime to fix pyart. I was waiting on the cftime developers to get back to me. The way you did it is what they suggested and is what I need to update the code to. Setting the parameter to false as a new update changed the return of python datetimes in num2date.

PR #914, opened 2020-03-23
MNT: New warnings on literals, and a fix for said warnings.

zssherman:
The parameter nowrap wasn't being used correctly and in kazr spectra 'is not' should not have been used for checking a float.


zssherman:
Tests pass merging.

Issue #916, opened 2020-03-23
Error in basic ingest example


louiswicker:
I am trying to write a ingest for MRMS radial velocity, and I was using the example in

http://arm-doe.github.io/pyart-docs-travis/notebooks/basic_ingest_using_test_radar_object.html

The n_azim and n_gates arguments to pyart.testing.make_empty_ppi_radar are reversed.

scollis:
Thanks @louiswicker !
@rcjackson can you take a look at this.. @IreneCrisologo can you help as this will be very useful for us.

zssherman:
Ah yep, I see that, my bad, I will do a pull request with the fix

zssherman:
Fixed and new notebooks are on the site now, closing.

PR #917, opened 2020-03-23
FIX: A fix for reversed ngates and nrays.

zssherman:
A fix for #916 the nrays and ngates in make_empty_ppi_radar were reversed. I fixed the revered inputs, the rest of the notebook should be fine as when the data for range and azimuth are populated they overwrite the incorrect numbers, but this fixes the incorrect empty object.


rcjackson:
This is a minor fix with CI passed. I am going to merge now.

Issue #551, opened 2016-06-30
Support for NEXRAD Level 3 product 176, Digital Instantaneous Precipitation Rate (DPR)


jjhelmus:
Py-ART cannot read NEXRAD Level 3 Digital Instantaneous Precipitation Rate (DPR) files (product 176).  They are stored as the `Generic Radial Product Format` which is not supported by the nexrad_level3.py module.

This product stores data in Generic Data Package packets (packet code 28) which is defined in Figure 3-15c on page 3-132 of "INTERFACE CONTROL DOCUMENT FOR THE RPG TO CLASS 1 USER" RPG Build 13.0, Document Number 2620001T.  This packet serializes the data using the External Data Representation (XDR) format, see Appendix E of the above document and RFC 1832.  

Reading this data should be possible with the [**xdrlib** module in the standard library](https://docs.python.org/3/library/xdrlib.html)

dopplershift:
There was some weirdness in the ICD too--MetPy has a working implementation: https://github.com/metpy/MetPy/blob/master/metpy/io/nexrad.py#L2082

jjhelmus:
@dopplershift Funny I was just look at that code :smile:

dopplershift:
Let me know if you find any issues with it--bugs, docs, whatnot.

jjhelmus:
Adding support for the product is now a high priority but when I get to it I let me know if I find anything.

dcedgren:
Just a note that I ran across this issue wanting to bring in DPR Level-III data into Py-ART. Looked into it briefly and I don't know what I'm doing well enough for me to try to use the Metpy code, and Metpy does not currently appear to have an output function.

So mostly just a quick note to say I'd appreciate support for this if you ever get around to it, or else any recommendations for a (not too technical) alternative.

scollis:
@dopplershift : Did you ever provide this in MetPy? 

One day we should discuss some bridges between MetPy and Py-ART

dopplershift:
We have the decoding working, we just don't have much of a data model around them; very ad-hoc data structures with dicts of dicts of lists of dicts...

scollis:
Is the data range-az-el data?

dopplershift:
Looks like it, yeah. Here's what I played with in MetPy:
```python
from metpy.io import Level3File
f = Level3File('testdata/nids/KOUN_SDUS84_DPRTLX_201305202016')
f.sym_block[0][0]
f.sym_block[0][0]['components']
```

[...]

PR #919, opened 2020-03-31
ENH: Support for Product 176 (DPR) - Instantaneous Precipitation Rate

dcedgren:
Added support for reading Nexrad Level III files of Product 176 in accordance with discussion in Issue #551.

I relied heavily on the Metpy implementation to read the Generic Radial Product format. But as I got into it I decided I was close enough to understanding the Pyart implementation that I could probably modify it with only short sections of additional code (as opposed to, say, reading in the file using Metpy and assigning bits and pieces of dictionaries and lists to line up with a pyart.core.radar object). It will be more maintainable this way.

In general I left the existing code and made product 176 work alongside it. Nothing of substance has changed for the code to read other products in.

The key piece I used from the Metpy implementation I used was the Level3XDRParser class. I went back and forth about whether or not to add Metpy as a dependency and import the class or to copy the 100-150 lines of code into Pyart. In the end I decided to copy them since per Dopplershift In Issue #551 a final data structure has yet to be decided for Metpy and it seems possible something could break here in the future. I think it seems reasonable if the maintainers want to go the other way instead. Also check that I did the license thing right.


dcedgren:
Is this something we'll want to add a new test for?

zssherman:
@dcedgren Thanks for submitting this! I agree, that maybe once the data structure is decided, the dependency can be added. @scollis Would do you think on the dependency part? @dcedgren a test would be preferred to test a Product 176 for this, but I don't think anything crazy as a lot of the nexrad level III is already tested. License looks good to me.

dcedgren:
Ok, added a test to read in an example DPR file. I just kept the same format as the other tests, so there are more tests than are probably necessary. Fingers crossed I did this all right.

zssherman:
Thanks for adding that! Seems to be failing due to a missing parenthesis syntax error line 202 of test nexrad level 3

dcedgren:
Hmm. I don't think it was a parenthesis. I think it was a period at the end of the line. The prior tests have periods at the end of the function definitions, but I've never seen that before. I just echoed it but I think that's what it didn't like.

dcedgren:
This pull request will fail because instrument altitude in the testing has been changed to meters rather than feet.

Once the new feet vs. meter pull request is merged it should pass and as far as I'm concerned should be ready to be incorporated.

dcedgren:
Will close PR and reopen to retrigger checks, which should pass now that feet vs. meters discrepancy has been resolved.

dcedgren:
Ok, that didn't work. I'll try adding a dummy commit instead.

zssherman:
Ah so @dcedgren its a conflict issue. I think it was from the pr #926

zssherman:
@dcedgren I'll do a PR to your branch shortly to fix it

dcedgren:
Yes, this was an issue where we were changing 2 things at the same time so I made the testing line up with the way it would eventually end up. It's actually PR #924. I think if we just rerun the test it should pass now that PR #924 has been merged.

zssherman:
Weird, it still shows conflicts, but they should have been resolved, is there maybe a marker still hidden somewhere? I don't see anything though which is weird

dcedgren:
Hmm. I may not have understood exactly what was going on here. I assumed the checks just had to be rerun, but I guess there is a conflict with another more recent pull request and Github can't figure out how to sort it. Sorry, still trying to learn the lingo.

zssherman:
There we go,  expect appveyor to fail, but its unrelated to this PR, but once travis passes, we should be groovy

dcedgren:
pyart/io/tests/test_nexrad_level3.py:196: AssertionError
0.244 == 0.244

I don't understand why Travis failed.

zssherman:
Weird... They match if i'm reading that correctly lol, im going to restart the build

dcedgren:
So it's the last line of the test, and of the module, that fails. That makes me suspicious, but I don't know what could cause that.

Also, how did you restart the build? That was what I was trying to do earlier.

zssherman:
I'm not sure either, it looks like it shows the values matching, which is odd. To restart the test, you go into details for the testing such as travis on this page. And once your GitHub is linked to travis, after going to the details page in the top right, there is a retart build option

dcedgren:
Ok, Travis passed. It should be good to merge.

zssherman:
Thanks again for doing this @dcedgren ! Merging.

dcedgren:
Absolutely! It's my first real open-source contribution and it's been fun. Thanks for your help.

PR #921, opened 2020-03-31
ENH: Remove disutils in favor of setuptools.

zssherman:
Removed disutils in favor of setuptools. Removed setup.py files. Pull request still in progress. Might add cython build in a separate pull request.


zssherman:
Requires cythonizing etc, I'll get back to this at a later date.

Issue #922, opened 2020-04-01
Nexrad Level 3 - Elevation in Feet, not meters


dcedgren:
For Product #176 (DPR) files the altitude is read into Pyart as meters, not feet as per the Nexrad documentation.

I suspect this may also be an issue with other products. I will confirm that this is an issue with other products and, if so, convert from feet to meters when the unit setting is meters (as is the default).

zssherman:
Hi @dcedgren I'll submit a pull request, so essentially the altitude needs to be multiplied by 0.3048?

dcedgren:
Correct.

But only if the user's units are set to meters. If they've changed them to feet we wouldn't want to convert.

Testing will also need to be changed. Including for the PR I'm working on. I intentionally made this wrong.

zssherman:
Ah gotcha, hmm, ok I'll close my PR, got to find a way to check if there units are already in meters.

zssherman:
@dcedgren It currently gets the height from self.prod_descr, which is unpacked from buffer.

dcedgren:
As far as I'm aware the Nexrad file will always be in feet. (Maybe someday we'll eventually manage to switch entirely to metric...)

BUT, it looks like the units might be switched by the user in default_config.py. If someone had switched this to feet instead I assume we wouldn't want to convert to meters. Unless this is never meant to be changed. You tell me.

We could easily check units at line 118 of nexradl3_read.py.

if altitude['units'] == 'meters':

zssherman:
Yeah, there users can change the config to their liking for field naming etc, but in terms of units I would believe meters is what we want too keep @scollis your thoughts? The gridding code and more do everything in meters.

zssherman:
Done in #924 merging

PR #924, opened 2020-04-01
FIX: This converts the nexrad altitude from feet to meters.

zssherman:
@dcedgren this should address #922 for nexrad level 3

### DIFF
--- a/pyart/graph/radardisplay.py
+++ b/pyart/graph/radardisplay.py
@@ -882,6 +882,13 @@ class RadarDisplay(object):
         # Get data and coordinate information
         az = self._radar.azimuth['data']
         el = self._radar.elevation['data']
+        if el[0] is None:
+            raise ValueError(
+                "Elevation is set to None. CR raster plotting is unavailable "
+                "for this dataset. Elevation can be set to None due to not "
+                "being present per Halfword 30 Table V of the ICD for NEXRAD "
+                "level 3 data.")
+
         rng = self._radar.range['data']
         data = self._radar.fields[field]['data']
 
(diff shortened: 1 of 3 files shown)


zssherman:
@scollis This would address the altitude units being feet and converting to meters, but would we want users to set altitude units to feet to keep those units? Would this cause issues in gridding and coordinate conversions?

dcedgren:
@zssherman Just a note that you may be confusing altitude and elevation. I think you meant this most recent commit to go on the other pull request.

I would vote against putting the warning on elevation being none where you did. If you put it in the IO module any time a user imports a Nexrad Level III product that's not in a certain list it's going to trigger the warning. If I didn't have context I think I might worry that something didn't import correctly with the text of the warning. Instead, why don't you make it an error and put it at the location where they actually try to plot an RHI with it, which would be undefined for that product?

dcedgren:
Ok, makes sense.

rcjackson:
It looks solid to me. Thankfully this was caught! Do we know if the previous version's gridding was impacted in any way?

zssherman:
@rcjackson I'm honestly not sure.

rcjackson:
I think it probably does, given that the altitudes are probably a factor of 3 too high. Therefore, I would put out a notice of this fix to pyart-users warning anyone using NEXRAD Level 3 data that the altitudes (in m) are off by a factor of ~3 due to this unit conversion. The users should regenerate any grids they made from NEXRAD L3 data as a precaution.

zssherman:
Sounds good, I'll put out a warning. And merge for now, thanks @dcedgren @rcjackson

Issue #923, opened 2020-04-01
Radar Elevation Misread for some Nexrad Level 3 Files


dcedgren:
Currently nexrad_level3.py assumes Halfword 30 of the Product Description block corresponds to the sweep elevation angle. In fact, it is a product-dependent parameter and this is only the case for some files. For some Nexrad Level 3 files there is not really a defined elevation. Product #176 (DPR), for example, and others.

There's a whole long list of product-dependent parameters in the ICD (Table V, which goes on for over 20 pages) and how to decode them from which halfwords. Some of these parameters are just file metadata, like compression information, but a lot of them are meteorological metadata, like start and end of precipitation or mean field bias or max velocity. It would be interesting to incorporate all these into Pyart, but it would be a fair bit of work. (Probably not hard, just a lot of putting the table into code.)

Unless someone is enthusiastic about this I recommend instead for the time being inserting logic that unless this file is one of the products for which halfword 30 corresponds to sweep elevation angle to set the radar object's elevation equal to zero or none.

zssherman:
Pull request has been merged, can reopen if needed. Closing for now.

PR #925, opened 2020-04-01
FIX: Elevation Angle for some products

dcedgren:
Many Nexrad Level 3 files have elevation angle defined by Halfword 30 of the product description block, but not all. (Elevation angle is not relevant for all products.)

Added a list of products for which elevation angle is defined by Halfword 30 per Table V of the ICD. For products for which it is not defined added logic to return data as None.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 .pydevproject
 *.rej
 .settings/
+.spyproject
 .*.sw[nop]
 .sw[nop]
 *.tmp
--- a/pyart/io/nexrad_level3.py
+++ b/pyart/io/nexrad_level3.py
@@ -191,7 +191,10 @@ class NEXRADLevel3File(object):
     def get_elevation(self):
         """ Return the sweep elevation angle in degrees. """
         hw30 = self.prod_descr['halfwords_30']
-        elevation = struct.unpack('>h', hw30)[0] * 0.1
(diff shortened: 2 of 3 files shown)


dcedgren:
Addresses Issue #923.

zssherman:
@dcedgren Thanks again for another change. The only thing is the data too None. I wonder if there should be a warning of some sort. Without elevation data, plotting etc I don't think  will no longer work. So I'm having a think on that one.

dcedgren:
@zssherman, I don't know what the trickle-down might be. Products without elevation angle should be able to be plotted as a PPI but would be undefined to be plotted as an RHI. None makes more sense to me, but if we need to change it to zero to avoid compatibility issues with other modules that seems ok, too.

zssherman:
@dcedgren I also think None would be the way to go. Only thing I would suggest is when elevation data is set to None a user warning stating that some functionality will be limited such as rhi plotting. @scollis are you good with this route?

scollis:
Yep! And thanks for the initiative!

dcedgren:
@zssherman this is something you can work on, right? Can you add a separate commit to this pull request?

zssherman:
@dcedgren ofcourse, I'll except the PR and submit a separate commit. Thanks again for the PR!

PR #926, opened 2020-04-02
ADD: Nexrad Level 3 Product Version Checking

dcedgren:
While I was combing through the ICD I realized there's a version number for each product included in the product description block that should be a good way to check that the product hasn't changed to a new version that is not supported by Pyart. If it has it issues a warning which hopefully should cause someone to take a look at the most recent ICD and verify there are no major changes.

I used the product version numbers in the table on pages 3-35 and 3-36 of the ICD FOR THE RPG TO CLASS 1 USER - Build 18.0.

I did not add version checking for the metadata parts that come before each product's data. I think there's a potential for these to change from build to build in a way that could break something, but also a greater potential to result in annoying warnings for no reason.

The good news is version numbers don't seem to change often. I compared Build 18 to Build 13, which is what this was originally based on, and found that they only increased the version number for products 165 and 172. Both were relatively minor. I checked that the new files uploaded successfully and they did. I couldn't find the parameters the ICD mentions that have changed for 172; I assume Pyart doesn't parse these. For 165 I added the Large Hail and Giant Hail classifications for user interpretation of the classification codes.

### DIFF
--- a/pyart/default_config.py
+++ b/pyart/default_config.py
@@ -831,6 +831,8 @@ nexrad_level3_metadata = {
                     '80: Big Drops (rain) (BD), '
                     '90: Graupel (GR), '
                     '100: Hail, possibly with rain (HA), '
+                    '110: Large Hail (LH), '
+                    '120: Giant Hail (GH), '
                     '140: Unknown Classification (UK), '
                     '150: Range Folded (RH)'),
         'coordinates': 'elevation azimuth range'},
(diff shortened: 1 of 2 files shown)


zssherman:
Hi @dcedgren , this looks good to me, I'm going to merge. Thanks for doing this!

dcedgren:
Not a problem. Now you've seen all the easy improvements I came across while doing what I set out to do. I figure once I got as deep in the code as I did I should probably take the next simple steps to improve what's a pretty obscure section of the codebase.

zssherman:
I appreciate the work on it. The nexrad code itself is from 2015 or earlier, so it did need some fine tuning.

scollis:
As do I ! Thanks @dcedgren

PR #929, opened 2020-05-01
Fix in filter_psidp in case of scalar mask

wolfidan:
I noticed that when writing a netCDF using write_cfradial for example, in the case a pyart field has a mask with all False (i.e. no masked values), it gets "compressed" to a single scalar boolean False. Parts of the Kdp processing codes couldn't handle that properly and assumed the mask to always be 2D. This fixes this particular case.


zssherman:
Hi @wolfidan , thanks for the PR! It looks good to me, merging.

PR #937, opened 2020-06-03
MNT: Attempt to fix appveyor path issue.

zssherman:
Since we are just renaming, using os, shutil behavior may have changed.


zssherman:
Tests pass merging.

Issue #940, opened 2020-07-01
gatefilter not working in `pyart.map.grid_from_radars`


swnesbitt:
I'm not having any luck getting a gatefilter to remove bad gates when passed to `pyart.map.grid_from_radars`.  I spoke to @rcjackson on Slack and he suggested using the `gatefilter.gates_excluded` to mask the data, but this does not seem to be functioning as intended.

![Screen Shot 2020-07-01 at 4 19 43 PM](https://user-images.githubusercontent.com/3288592/86292622-b1d21a80-bbb6-11ea-88f0-0e326a97b643.png)

scollis:
Can you try something crazy like excluding Z> 20

What did you construct the filter from?

-sent from a mobile device-

swnesbitt:
# filter out gates with reflectivity > 100 from both radars
gatefilter_1 = pyart.filters.GateFilter(radar1)
gatefilter_1.exclude_transition()
gatefilter_1.exclude_below('normalized_coherent_power', 0.3)
gatefilter_1.exclude_above('differential_reflectivity', 5)
pyart.correct.despeckle.despeckle_field(radar1, 'reflectivity', label_dict=None, threshold=-100, size=10, gatefilter=gatefilter_1, delta=5.0)
gatefilter_1.exclude_masked('reflectivity')
gatefilter_1.exclude_masked('velocity')

swnesbitt:
I tried excluding reflectivity > 20 - works in the plot, but not the grid again.

![Screen Shot 2020-07-01 at 4 30 09 PM](https://user-images.githubusercontent.com/3288592/86293402-26598900-bbb8-11ea-9805-8c0a8f1db04a.png)

swnesbitt:
Was able to workaround with
```
radar1.fields['reflectivity']['data']=np.ma.masked_where(gatefilter_1.gate_excluded,radar1.fields['reflectivity']['data'])
radar1.fields['velocity']['data']=np.ma.masked_where(gatefilter_1.gate_excluded,radar1.fields['velocity']['data'])
```
Off to `pydda` land!

scollis:
Cool beans. Zach can you use a gate filter on some data to see if it works on our end 

-sent from a mobile device-

zssherman:
Yeah, I'll take a look and see if I can figure out what's going on.

zssherman:
Yeah, @swnesbitt @scollis 
[...]

PR #946, opened 2020-07-21
FIX: Fix for missing gatefilter in map_to_grid.

zssherman:
When looking into the gatefilters issue #940, I noticed gatefilter in map_to_grid was broken due to the masked iterator behavior. Found a fix by flatten() the arrays to avoid the sequence error.


zssherman:
Passes merging.

PR #947, opened 2020-07-22
FIX: Gate to grid map blotchy

rcjackson:
Fixing the blotchiness in the nearest neighbor cython code.


zssherman:
Looks good to me, CI failure is unrelated. Merging.

PR #949, opened 2020-07-23
TST: Unit test of get_gate_area and some cleanup.

zssherman:


zssherman:
Need to figure out a way to fix appeyor, merging this as travis passes.

PR #950, opened 2020-07-24
FIX: add_gate_area now returns only positive values

rtanamachi:
Take absolute value of gate area to avoid returning negative areas. This covers cases where azimuth is decreasing instead of increasing. Thanks to John Krause for the suggestion.


zssherman:
Looks good to me thanks @rtanamachi! Merging.

Issue #955, opened 2020-08-19
dtype not preserved after gridding


vlouf:
The dtype of input fields is not preserved by the gridding operation. For example, in the original ppi radar, we have these fields:
```python
{'radar_echo_classification': dtype('int16'),
 'radar_estimated_rain_rate': dtype('float32'),
 'velocity': dtype('float32'),}
```

And after gridding everything is float32. Currently I have this little piece of code to correct the types of the input fields:
```python
intypes = {k: v['data'].dtype for k, v in radar.fields.items()}
for k, v in intypes.items():
    grid.fields[k]['data'] = grid.fields[k]['data'].astype(v)
```
But it would be nicer if the gridding preserved the type (or automatically correct it right after gridding).

scollis:
Thanks @vlouf Agree

zssherman:
@scollis @vlouf See #956

PR #956, opened 2020-08-19
ENH: Adding copy_field_dtype to grid mapper.

zssherman:
New parameter copys original radar fields dtype over to the grid.

@scollis @vlouf This should do what is requested in #955 . I added it as a parameter though to grid_from_radars.


zssherman:
@vlouf When you get a chance can you check with your case. This worked on some tests radars I had.

vlouf:
@zssherman yes the fix is working, thanks!

zssherman:
Cool! Merging.

Issue #977, opened 2021-02-18
pyart.aux_io.read_gamic() failing with h5py 3.1.0


cclopes:
When I try to open a file with `pyart.aux_io.read_gamic()` in an environment with h5py 3.1.0, the following error occurs:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-22674a19b649> in <module>
----> 1 pyart.aux_io.read_gamic("../dados/radar/SR/SRO-250--2017-03-14--18-30-22.mvol")

~/miniconda3/envs/tutoriais-python/lib/python3.8/site-packages/pyart/aux_io/gamic_hdf5.py in read_gamic(filename, field_names, additional_metadata, file_field_names, exclude_fields, include_fields, valid_range_from_file, units_from_file, pulse_width, **kwargs)
    135     # sweep_type
    136     scan_type = gfile.raw_scan0_group_attr('what', 'scan_type').lower()
--> 137     scan_type = scan_type.decode('utf-8')
    138     # check that all scans in the volume are the same type
    139     if not gfile.is_file_single_scan_type():

AttributeError: 'str' object has no attribute 'decode'
```

<details>
<summary>Environment that caused the error</summary>

```bash
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       1_gnu    conda-forge
arm_pyart                 1.11.3           py38h5c078b8_0    conda-forge
backcall                  0.2.0              pyh9f0ad1d_0    conda-forge
backports                 1.0                        py_2    conda-forge
backports.functools_lru_cache 1.6.1                      py_0    conda-forge
boost-cpp                 1.74.0               hc6e9bd1_2    conda-forge
brotlipy                  0.7.0           py38h497a2fe_1001    conda-forge
bzip2                     1.0.8                h7f98852_4    conda-forge
c-ares                    1.17.1               h36c2ea0_0    conda-forge
ca-certificates           2020.12.5            ha878542_0    conda-forge
cached-property           1.5.1                      py_0    conda-forge
cairo                     1.16.0            h7979940_1007    conda-forge
cartopy                   0.18.0          py38h2516ec2_10    conda-forge
certifi                   2020.12.5        py38h578d9bd_1    conda-forge
cffi                      1.14.5           py38ha65f79e_0    conda-forge
cfitsio                   3.470                hb418390_7    conda-forge
[...]

PR #978, opened 2021-02-18
FIX: Fix for h5py > 3.0 decoding of str.

zssherman:
Have a check for earleir versions of h5py in the meantime but changes
should work for newer and earlier versions


zssherman:
A fix for issue #977

zssherman:
Approving, error is unrelated to PR.

Issue #813, opened 2019-02-18
Adding additional colorbar functionality 


atmosmattbrewer:
Hello,
So I have never made edits to any package before but I was frustrated by the lack of changes to colorbar parameters that were in the original code, or at least from what I could figure out. Thus I decided to add a couple extra line to the radardisplay.py script under the plot_colorbar:

def plot_colorbar(self, **label_size =22,ticklabel_size = 22,fraction=0.046,** mappable=None, 		field=None, label=None,orient='vertical', cax=None, ax=None, fig=None,ticks=None, 			ticklabs=None):  
        Plot a colorbar.

        Parameters
       
        mappable : Image, ContourSet, etc.
            Image, ContourSet, etc to which the colorbar applied.  If None the
            last mappable object will be used.
        field : str
            Field to label colorbar with.
	**label_size : int
		Font size of the colorbar label. Default is 22
	ticklabel_size : int
		Font size for the tick label on the colorbar. Default is 22
	Fraction : int
		Fraction of original axes to use for colorbar. Default of 0.046 seems to work well from my testing**
        label : str
            Colorbar label.  None will use a default value from the last field
            plotted.
        orient : str
            Colorbar orientation, either 'vertical' [default] or 'horizontal'.
        cax : Axis
            Axis onto which the colorbar will be drawn.  None is also valid.
        ax : Axes
            Axis onto which the colorbar will be drawn. None is also valid.
        fig : Figure
            Figure to place colorbar on.  None will use the current figure.
        ticks : array
            Colorbar custom tick label locations.
        ticklabs : array
                Colorbar custom tick labels.
        
        if fig is None:
            fig = plt.gcf()
        if mappable is None:
            mappable = self.plots[-1]
        if label is None:
            if field is None:
                field = self.plot_vars[-1]
            label = self._get_colorbar_label(field)

        **cb = fig.colorbar(mappable, orientation=orient, ax=ax, cax=cax,fraction=fraction)
        cb.ax.tick_params(labelsize=ticklabel_size)**
        if ticks is not None:
            cb.set_ticks(ticks)
        if ticklabs:
            cb.set_ticklabels(ticklabs)
        **cb.set_label(label, fontsize = label_size)**

	

        self.cbs.append(cb)

I'm not sure how to contribute this to the source code or even if this is the best way to do it but it works well for what I needed. So I thought I'd share it here so maybe it could be added in the future.

scollis:
Thanks! We will take a look over this

[...]

PR #984, opened 2021-03-12
Added additional colorbar functionality

HumphreysCarter:
Added additional colorbar functionality from matplotlib, see https://github.com/ARM-DOE/pyart/issues/813.


kmuehlbauer:
@HumphreysCarter 

Since those keyword arguments are all fed to matplotlib `colorbar`, why not just use `**kwargs` to propagate this?

HumphreysCarter:
> 
> 
> @HumphreysCarter
> 
> Since those keyword arguments are all fed to matplotlib `colorbar`, why not just use `**kwargs` to propagate this?

That's true, I hadn't had much experience with `**kwargs` so that's probably why I overlooked that.

kmuehlbauer:
In the end that's a design decision. If `**kwargs` are used, the user can specify all kwargs which `plt.colorbar` can consume and has absolute freedom to tweak the colorbar as wanted . You still can give reasonable defaults within this funcion when a specific `kwarg` (eg. `aspect`) is not set. 

@zssherman It seems `plot_colorbar` is just a thin wrapper around `plt.colorbar` with some additions, so it might be a good idea to use `kwargs` here. Thoughts?

zssherman:
Thanks @kmuehlbauer for the feedback, and I do agree! @HumphreysCarter Thank you for the PR! I'm wondering if then we can keep the doc strings as it would help the user navigate the other features if they wish to change, but I do agree the kwargs would allow for more freedom.

HumphreysCarter:
Hi @zssherman sorry for my delay in getting this fixed. I removed the doc strings and followed @kmuehlbauer's recommendation of using kwargs. I believe I implemented them correctly?

zssherman:
@HumphreysCarter No worries! Looks like the error is plt.colorbar got tabbed possibly on accident.

HumphreysCarter:
Whoops! Thanks for pointing that out @zssherman.

zssherman:
No worries!

zssherman:
Thanks @HumphreysCarter for the PR! I'm going to accept and do a test fix in a new PR.

HumphreysCarter:
Awesome, thanks @zssherman!

PR #988, opened 2021-04-25
Change default weighting function to Barnes2 

Quba1:
`pyart.map.map_gates_to_grid` and `pyart.map.map_to_grid` emit deprecation warning that Barnes weighting function is deprecated, and suggest using Barnes 2. 

I propose changing the default weighting function to Barnes 2 in those two functions. Therefore less experienced PyART users will not be confused which function they should use. Advanced users still can change the function manually by kwargs. 

This PR also includes a few typo corrections in the documentation.

### DIFF
--- a/examples/mapping/plot_map_one_radar_to_grid.py
+++ b/examples/mapping/plot_map_one_radar_to_grid.py
@@ -20,7 +20,7 @@ import pyart
 RADAR_FILE = '110635.mdv'
 radar = pyart.io.read_mdv(RADAR_FILE)
 
-# mask out last 10 gates of each ray, this removes the "ring" around th radar.
+# mask out last 10 gates of each ray, this removes the "ring" around the radar.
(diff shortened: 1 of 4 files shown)


zssherman:
HI @Quba1 thanks for the PR. Yeah the original thought was to allow transition if individuals were using barnes in operations. I think it has been long enough now to do such a change, @scollis would you be fine if we switch the default now to barnes2?

scollis:
This is a great and timely conversation.
Here is what I suggest:
First put in a PR that ads a warning that is activated any time some one uses default and lets users know the default is changing on the 1st of June to Barnes2.
On June 1 accept this PR.

Also ping email list

This will avoid anyone's code doing unexpected things

zssherman:
Sounds good

Quba1:
Hi, just kindly reminding that this PR can possibly be merged.

zssherman:
@Quba1 Sorry, slipped the radar, merging thanks again for the PR!

PR #991, opened 2021-04-29
Add check if radars tuple is not empty

Quba1:
This is the same PR as [#990](https://github.com/ARM-DOE/pyart/pull/990) (reopened as I have forgotten to add the check to the `map_to_grid` function).

In some edge cases (in my case when radars tuple is created in a loop) the empty tuple can be passed to the `gates_to _grid` or `map_to_grid` functions. In that case`tuple index out of range` error occures when `radars[0].altitude['data']` is acessed, and it is not always obvious what is the cause of that error.

I propose to add a simple check if `radars` tuple length is equal to 0. If yes then ValueError exception is raised with a more meaningful error message indicating what the issue is.
The placement of that `if` statement is just after conversion of single `Radar` object into `radars` tuple, so that tuple length can be checked without worring about edge cases. 
In this solution also length of `gatefilters` tuple is implicitly checked as `gatefilters` must have the same length as `radars`. 
The check is also added to `grid_from_radars` function as there is no need to call the gridding functions if the radars tuple is empty.

### DIFF
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -72,6 +72,9 @@ def map_gates_to_grid(
     if isinstance(radars, Radar):
         radars = (radars, )
 
+    if len(radars) == 0:
+        raise ValueError('Length of radars tuple cannot be zero')
+
     skip_transform = False
     if len(radars) == 1 and grid_origin_alt is None and grid_origin is None:
         skip_transform = True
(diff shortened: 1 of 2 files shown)


zssherman:
Thanks @Quba1! Everything looks good, down the roqd we may need a new unit tests to pytest raise the error. Merging.

PR #1001, opened 2021-07-15
FIX: Correctly decode format from bytes to str

ReallyNiceGuy:
When running in python 3, the format was kept as bytes and it would fail to load the data.


zssherman:
Thanks @ReallyNiceGuy  for doing this, merging!

PR #1003, opened 2021-07-23
ENH: Default scan_rate to target_scan_rate

ReallyNiceGuy:
GAMIC's volume doesn't mandate the real scan_rate to be stored,
so we initialize scan_rate to the same value as the target_scan_rate.


zssherman (approved):
Thanks for doing this! Looks good to me! Merging.

PR #1014, opened 2021-10-11
ENH: GAMIC's volume doesn't mandate a pulse_width field

ReallyNiceGuy:
Set the default value of the pulse_width field to None and allow
a GAMIC volume to be loaded even if this field is missing.


zssherman (approved):
Sorry for the late response. Looks good to me, merging!

PR #1019, opened 2021-11-16
FIX: Set minimum for cost function gradient INF norm to be nyquist interval

rcjackson:
This is a bugfix for the functionality to use sounding data to fix velocities from region based dealiasing that are off by fixed nyquist velocities. The cost function will now have been determined to reach a minimum once the difference between any region and the reference velocity is less than a nyquist velocity (it was not stopping before this before, causing issues). In addition, the fact that the closest region must be less than one nyquist velocity is now a hard constraint coded in the gradient.


zssherman (approved):
Looks good to me, merging. Thanks @rcjackson !

PR #1020, opened 2021-11-19
ADD: Options to select ODIM HDF5 datasets

ritvje:
Hi! Here's another pull request from my personal modifications. At the Finnish Meteorological Institute, the radar volumes in ODIM HDF5 format store all elevation scans in the same file, but some of the scans have different amount of range bins and different range resolution. This makes it impossible to read the volumes with pyart without picking the correct scans into separate files first with some external tool. 

So as a workaround, I'm suggesting adding options `include_datasets` and `exclude_datasets` in `pyart.aux_io.read_odim_h5` to select the datasets that are read. The default options select all datasets, so old behaviour is retained.

For example (the volume file is available from [AWS](http://s3-eu-west-1.amazonaws.com/fmi-opendata-radar-volume-hdf5/2021/11/19/fivih/202111190500_fivih_PVOL.h5)):
```python
>>> radar = pyart.aux_io.read_odim_h5("202111190500_radar.polar.fivih.h5")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ritvanen/Documents/koodaus/pyart/pyart/aux_io/odim_h5.py", line 231, in read_odim_h5
    raise ValueError('range scale changes between sweeps')
ValueError: range scale changes between sweeps
>>> radar = pyart.aux_io.read_odim_h5("202111190500_radar.polar.fivih.h5", include_datasets=["dataset1", "dataset2"])
>>> print(radar.fixed_angle)
{'long_name': 'Target angle for sweep', 'units': 'degrees', 'standard_name': 'target_fixed_angle', 'data': array([0.3, 0.7], dtype=float32)}
```

### DIFF
--- a/pyart/aux_io/odim_h5.py
+++ b/pyart/aux_io/odim_h5.py
@@ -30,9 +30,9 @@ ODIM_H5_FIELD_NAMES = {
     'KDP': 'specific_differential_phase',
     'SQI': 'normalized_coherent_power',
     'SNR': 'signal_to_noise_ratio',
-    'VRAD': 'velocity', # radial velocity, marked for deprecation in ODIM HDF5 2.2
-    'VRADH': 'velocity', # radial velocity, horizontal polarisation
-    'VRADV': 'velocity', # radial velocity, vertical polarisation
+    'VRAD': 'velocity',  # radial velocity, marked for deprecation in ODIM HDF5 2.2
+    'VRADH': 'velocity',  # radial velocity, horizontal polarisation
+    'VRADV': 'velocity',  # radial velocity, vertical polarisation
     'WRAD': 'spectrum_width',
     'QIND': 'quality_index',
 }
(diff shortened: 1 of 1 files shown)


zssherman:
Reclosing and reopening, for some reason travis did not run.

zssherman:
With that fixed @ritvje I think this is a great fix. I have thought of this before and how I can maybe get these readers to still read these datasets with differing range bins. Might be easier once we start integrating Xarray which hopefully will begin soon.

ritvje:
Yeah, having different range bins would be great, but in the meanwhile I think this would be a good solution. And even with shared range bins, often reading the whole volume is not needed, so selecting only the required datasets saves also memory.  

Is there anything I should do regarding the failed checks? I'm not an expert, but it looks to me like the failed tests don't use this function.

zssherman:
@ritvje Looks like there was an update to matplotlib that broke the unittests, yeah unrelated to this PR, so merging.

zssherman:
Thanks again!

PR #1024, opened 2021-12-14
ADD: Adding a feature to map gates of approximately equivalent location from one radar geometry to another

rcjackson:
I have added a GateMapper class in the pyart.map module for making mapping gates of roughly equivalent location easy between radars. It even supports returning a Radar object with the field from the source radar mapped to the destination. 

nexrad_on_csapr2
nexrad


zssherman:
@rcjackson Awesome, thanks for doing this! The unit tests failing is unrelated to this. The only thing I would change is remove the whitespace after the unit test function. Once that's done, I'll approve and merge.

PR #1029, opened 2022-01-06
FIX: Fixes for overlapping axes in mapdisplay modules.

zssherman:
When parsing a default matplotlib axes for the mapdisplay modules in cartopy 0.19 or higher, a duplicate axes is formed with geoaxes is defined. Removed default defined matplotlib axes as this mapdisplay is utilizing cartopy anyways, so a default geoaxes is defined instead if user provides no projection or axes. Also change of ax.outline_patch to ax.spines['geos'] to removed deprecation warning.

### DIFF
--- a/pyart/graph/gridmapdisplay.py
+++ b/pyart/graph/gridmapdisplay.py
@@ -167,7 +167,6 @@ class GridMapDisplay(object):
         ds = self.grid.to_xarray()
 
         # parse parameters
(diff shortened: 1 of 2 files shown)


zssherman:
@rcjackson If you have a quick minute, for a review.

zssherman:
example of issue:
overlap

PR #1031, opened 2022-02-15
Add function to "image mute" radar object

lauratomkins:
This function is intended to split a field in a radar object into two separate fields based on a threshold from a different field. Specifically, I designed this function to split reflectivity into two separate fields based on a threshold from the correlation coefficient field to help detect regions of mixed precipitation. Manuscript in prep.


zssherman:
Hmm I can't figure out why an exact local copy is not failing but the CI here is.

scollis:
Thanks for the PR! This looks very interesting!

zssherman:
@lauratomkins I'm still trying to figure out why the tests are failing. However in the meantime, the things I noticed with a second review, is can you remove the dict standard setting. So:
mute_dict['standard_name'] and
non_mute_dict['standard_name']
as these only defined when an cfradial standard name from the cfradials tables is used.
As well as creating a separate radar for the mute_radar in the testing:
mute_radar = pyart.util.image_mute_radar(radar etc etc)
then reference that mute_radar in the asserts?

I'll keep looking to see if I can find why the tests are giving different results locally and here.

zssherman:
I think I figured it out, well partially lol, so for some reason when we read the file within the test function pytest fails when the full suite of tests run. It was passing for me because I was only running the tests in utils. Why pyests behavior is like this I'm not sure, I'll need to look into it more. @lauratomkins If you move the radar = pyart.io.nexrad read the testing file line above the def of the test function:
radar = pyart.io.read(pyart.testing.NEXRAD_ARCHIVE_MSG31_FILE)
def test_image_mute_radar():
    mute_radar = pyart.util.image_mute_radar(
            radar, field='reflectivity', mute_field='cross_correlation_ratio',
            mute_threshold=0.97)

This CI should now pass.

PR #1064, opened 2022-03-10
MNT: Removing pandas import and converters.

zssherman:
Was a temp fix for np.datetime64 for python3.8, removing this no longer should be an issue and or required.


mgrover1 (approved):
Looks good - thanks for putting this together @zssherman

PR #1071, opened 2022-03-14
Added RXM-25 reader

rtanamachi:


rtanamachi:
Thanks Zach! I may have committed this a bit prematurely; I meant to merge to my own forked main rather than the ARM-DOE main. Well, let's see if it works! I can provide some sample files to check; just let me know.

scollis:
First PR since GitHub actions! 

Thanks for the PR @rtanamachi .. Is this the format used by Ridgeline?

rtanamachi:
Yes, that's correct.

zssherman:
@rtanamachi Just a few recommendations. We can probably remove import sys, glob, and os as they aren't being used. I would also rename the parameter netcdf_file to filename as it would match other readers as well as match with the docstring. I'll add an exclude fields parameter separately to this. And can you send me a test file? Just want to do a test real quick.

rtanamachi:
> We can probably remove import sys, glob, and os as they aren't being used. I would also rename the parameter netcdf_file to filename as it would match other readers as well as match with the docstring.
Agree on both counts.
> I'll add an exclude fields parameter separately to this. 👍 
> And can you send me a test file? 
I will send you links via email, as they are rather voluminous!

zssherman:
Going to merge and do some minor changes. Thanks for the PR @rtanamachi !

PR #1108, opened 2022-04-14
add verbose flag to qvp code

dopplerchase:
While seeing all available radar tilts for each file is useful, I think it is unnecessary to have it print out every single time. So I have added a bool to the input where you can turn that print statement on-off, but default it is off. Feel free to switch it to default on. I just wanted control to turn it off so my progress bar worked properly.

### DIFF
--- a/pyart/retrieve/qvp.py
+++ b/pyart/retrieve/qvp.py
@@ -27,7 +27,6 @@ def quasi_vertical_profile(radar, desired_angle=None, fields=None, gatefilter=No
         None will result in wanted_angle = 20.0
     verbose : bool
         boolean flag to turn the printing of radar tilts on or off. 
-        Can be very annoying if you process alot of files. 
 
     Other Parameters
     ----------------
--- a/pyart/retrieve/qvp.py
+++ b/pyart/retrieve/qvp.py
@@ -8,7 +8,7 @@ import numpy as np
 from ..core.transforms import antenna_to_cartesian
 
 
-def quasi_vertical_profile(radar, desired_angle=None, fields=None, gatefilter=None):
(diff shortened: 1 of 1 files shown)


zssherman:
Thanks @dopplerchase for the PR! I agree this is a better way to handle the print statements. Only recommendation, can we remove the second sentence of the docstring for the verbose parameter? Besides that once the tests run and pass should be good.

pyart/retrieve/qvp.py
```diff
@@ -25,6 +25,9 @@ def quasi_vertical_profile(radar, desired_angle=None, fields=None, gatefilter=No
     desired_angle : float
         Radar tilt angle to use for indexing radar field data.
         None will result in wanted_angle = 20.0
+    verbose : bool
+        boolean flag to turn the printing of radar tilts on or off. 
+        Can be very annoying if you process alot of files.
```
mgrover1, line 30: Yeah - I would agree with @zssherman here, can you remove this line? The rest looks great 😁

mgrover1 (changes requested):
One small comment - other than that, looks great!

pyart/retrieve/qvp.py
dopplerchase: Git novice over here. How do I make a change to an exisitng pull request?

zssherman: or to your main

mgrover1: You'll want to make this change on your branch, commit, then push again

mgrover1 (approved):
Thanks for making those changes @dopplerchase ! Once the tests pass, this looks good to merge 🚀

PR #1122, opened 2022-04-20
MNT: Utilize conda for build due to trmm_rsl.

zssherman:
@mgrover1 I think this should work.

### DIFF
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -29,6 +29,7 @@ jobs:
 
       - name: Set TRMM RSL path
         run: |
+          mamba install -c conda-forge trmm_rsl
           export RSL_PATH=$CONDA/pkgs/trmm_rsl-1.49-3
 
       - name: Install pip dependencies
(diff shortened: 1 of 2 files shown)


mgrover1:
It's not available on windows which is why those are failing - we might need a `condo install -c conda-forge trmm_rsl` to the pypi workflow

.github/workflows/pypi-release.yml
```diff
@@ -8,23 +8,37 @@ jobs:
...
+          activate-environment: pyart-dev
+          auto-activate-base: false
+          mamba-version: '*'
+          use-mamba: true
+          miniforge-variant: Mambaforge
+          extra-specs: python=3.8
+
+      - name: Set TRMM RSL path
+        run: |
+          export RSL_PATH=$CONDA/pkgs/trmm_rsl-1.49-3
```
mgrover1, line 32: Add in a line here `conda install -c conda-forge trmm_rsl` instead of adding to the environment.yml for the CI

mgrover1 (changes requested):

zssherman:
Ah yep makes sense, changing that now.

mgrover1 (approved):
Good to go! The windows builds are passing now

PR #1144, opened 2022-04-25
DOC: PyART doc static image color schemes changed to colorblind frien…

jrobrien91:
…dly PyART scheme.

plot_ppi_mdv.py modified for this change.
plot_rhi_cfradial_singlescan.py created to create one rhi cfradial plot for the docs.


jrobrien91:
Just noticed that the default figure sizes for these images are slightly larger than the current docs. Please let me know if this is going to be an issue, and I'll force them to the current sizes. It's nice that GitHub will tell you pixel sizes!

PR #1148, opened 2022-05-02
MNT: Changing dependency check to Cartopy.

zssherman:
Running into errors with Xarray being a dependency.


mgrover1 (approved):
Thanks for adding this @zssherman !

PR #1149, opened 2022-05-03
TST: Adding bias unittests as well as attempt to fix rsl pathing issue.

zssherman:


zssherman:
I'm going to close for now and try different testing methods

PR #1151, opened 2022-05-04
FIX: Fixes an issue with mismatch arrays.

zssherman:
When using sonde profiles, when it resets the variable with the same name, get issues due to the shape mismatch. Creating an array with same shape as the no_nan array.

### DIFF
--- a/pyart/util/simulated_vel.py
+++ b/pyart/util/simulated_vel.py
@@ -6,7 +6,7 @@ Function for creating simulated velocity fields.
 import numpy as np
 from scipy.interpolate import interp1d
 
-from pyart.config import get_metadata, get_field_name
+from pyart.config import get_metadata
 
 
 def simulated_vel_from_profile(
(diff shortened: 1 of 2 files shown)


mgrover1 (changes requested):
Good catch - would you be willing to add a test here to verify this is working as expected?

zssherman:
Ah i forgot awhile back I added a skip because I ran into an issue, let me remove the skip and see if this fixes the failure from the past.

zssherman:
I sim_vel unittests should run again now

zssherman:
Now I remember why i skiped at the time, its not finding simulated_velocity despite it being in the config lol, I remember I was scratching my head for awhile on this one

mgrover1:
@zssherman looks like it's having an issue with some attribute?

zssherman:
Yeah we are getting incredibly odd behavior with this test. The attribute is a scipy error, that same code is called in other tests in pyart and does not fail. This test also fails when trying to find simulated velocity field name, neither of these fail locally, none of this makes sense at all

zssherman:
I think this is why i skipped it before, because the tests being pull by pytest for simulated velocity won't work correctly

mgrover1:
Good to merge @zssherman ?

PR #1152, opened 2022-05-04
FIX: Fix rounding issue with mask arrays, using np.round instead.

zssherman:


rcjackson:
Should we use np.ma.round if we're looking to deal with masked arrays. Or perhaps check for the mask and use np.round or np.ma.round depending on if there is a masked array.

zssherman:
@rcjackson I believe it might be always a mask away based on the masking in the simulated velocity profile code, so might be better to switch it to np.ma.round

PR #1154, opened 2022-05-05
Add Image muting plotting example to gallery

lauratomkins:
This script provides an example of how to use the `image_mute_radar` function.


zssherman:
@lauratomkins Thanks for doing this PR! The only thing I would change would be removal of the two lines of whitespace at the bottom of the script, besides that looks good to me!

zssherman:
Looks good! Merging.

PR #1158, opened 2022-05-19
MNT: Mispelling of embelish

zssherman:
Pointed out by David Wolff embellish was mispelled, this is a fix for that.


mgrover1 (approved):
Thanks for fixing this @zssherman

PR #1196, opened 2022-07-06
FIX: C-band values being set by x-band.

zssherman:
While creating unitests I noticed in the _mass_centers_table, the dictionary was incorrectly being set for c-band, x-band values were being set for c-band and x-band. One way around this is to create separate arrays instead of overwriting it. Seems to have fixed the issue.


mgrover1 (approved):
Thanks for catching this @zssherman

PR #1204, opened 2022-07-15
FIX: update join_radar to also append instrument parameters

wolfidan:
Hi dear Py-ART team,

The join_radar function in Py-ART has the problem that it does not merge instrument parameters. This is an issue for certain functions for example _dealias_region_based_, which expects a nyquist_velocity value for every sweep. I think there is currently no way to use this function on a multi sweep scan obtained with join_radar. In our MeteoSwiss version we use the function in this PR, which also merges the most important instrument parameters and is able to merge radar objects which have different variables (by merging only the variables that they have in common).
The code below for example fails for me with the current Py-ART version with error

_nyquist_vel = [radar.get_nyquist_vel(i, check_uniform) for i in 
return(nyq_vel[0])
TypeError: only size-1 arrays can be converted to Python scalars_

 but works with this fix.  
````

for i,f in enumerate(files):
    radar =  pyart.io.read_cfradial(f)
    if i == 0:
        mradar = radar
    else:
        mradar = pyart.util.join_radar(mradar, radar)
    
    
wind = pyart.correct.dealias_region_based(mradar, ref_vel_field = 'velocity')

````
Cheers,
Daniel

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ The Python ARM Radar Toolkit (Py-ART)
 
 |GithubCI| |CodeCovStatus|
 
-|AnacondaCloud| |CondaDownloads|
+|AnacondaCloud| |PyPiDownloads| |CondaDownloads|
 
 |DocsUsers| |DocsGuides|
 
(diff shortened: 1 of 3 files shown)


wolfidan:
Sorry warn function is missing, I'll fix that ASAP.

zssherman:
Thanks @wolfidan for doing this! Yeah once the warn is fix illl do another review.

zssherman:
@wolfidan Looks good, only thing is there is an added whitespace near the top making 3 spaces between the two functions. More nitpicking, but besides that everything looks good.

mgrover1 (changes requested):
@wolfidan thanks for adding this fix! One small suggestion - could you please add a test for this?

zssherman:
Doing a test on CI, why there was a close and reopen

wolfidan:
Thank you @mgrover1, I have added an unittest that check if a join of two radars with different numbers of sweeps and only one common field is possible.

mgrover1 (approved):
Great! Thanks @wolfidan

PR #1207, opened 2022-07-18
TST: Moving tests to test module within Py-ART main directory.

zssherman:


mgrover1:
@zssherman I am pretty sure this has to do with the way the test configuration file is being used.

zssherman:
Not sure about this new macos error

zssherman:
@mgrover1 This should be good to go now, the error is unrelated to the PR.

mgrover1 (approved):
Once the final tests pass, this looks great! Thanks for all this hard work @zssherman

zssherman:
@mgrover1 Oh just forgot, so would it be prune tests in manifest.in for ignore tests for the repo?

mgrover1:
@zssherman yes!!

zssherman:
@mgrover1 However, if they aren't included, how does conda forge test the package when its uploaded as a PR for a new release?

mgrover1:
We don't need to explicitly include it - see https://github.com/Unidata/MetPy/blob/main/MANIFEST.in

zssherman:
Ah gotcha, so just leave the manifest as is?

mgrover1:
@zssherman I think so.

PR #1208, opened 2022-07-18
FIX: Fix for only python 3.10 being used.

zssherman:


zssherman:
Issue for wrong python version during tests has been fixed, still library issue with MacOS for PIL however

zssherman:
Error seems to be addressed in https://github.com/conda-forge/libwebp-feedstock/issues/26
so looking through our installs to see if there is any reason why we are encountering this now.

zssherman:
@mgrover1 This fixes only python 3.10 being used for all tests, also fixes pinning issues. However the current error is related to the PR above, however it has been fixed but can't seem to get it to work and it recently broke, so im curious if another release of theirs broke something.

mgrover1 (approved):
Good catch @zssherman

mgrover1:
And the registry

zssherman:
For the other PR? Sorry I don't follow @mgrover1

mgrover1:
yup other PR sorry

PR #1214, opened 2022-07-21
TST: Fix median filter depreciation warnings as well as more tests for simple moments.

zssherman:


zssherman:
Some weird behavior going on with arrays matching or not generating the same on mac....

mgrover1:
Looks good now!

mgrover1 (approved):
Great! Thanks @zssherman

zssherman:
@mgrover1 Did some more changes for the default config and another depreciation warning,

PR #1215, opened 2022-07-22
ENH: Corrected Height Gates to be height above sea level

jrobrien91:
After analysis of CMAC2.0 processed radar files, the extracted radar column heights did not match the calculated CMAC2.0 heights. Additionally, get_fields_location failed with CMAC2.0 files. This was determined to be due to attempting to set the dimension of the DataSet with the same name as one of the CMAC2.0 fields (e.g. 'height').

Testing showed that the extracted radar column heights were not heights above sea level and were also interpolating gates based on the range edges. 

To fix this, the altitude of the radar was added to extracted radar column height gates and edges option on antenna_vectors_to_cartesian was set to 'False'. 

Check was added for a 'height' field within the radar object. 

'Base_time' and "time_offset" data variables were added to the DataSet to conform to the ARM standards. 'base_time' corresponds to the time units within the radar file. 'time_offset' are the times to the center of each height gate. 

Various DataSet attributes are set to update meta data for the extracted column.


jrobrien91:
Forgot to pull upstream changes before submitting, so my testing directories are messed up. I'll take a look at this tonight/tomorrow morning and commit changes

zssherman:
@jrobrien91 The unit tests just need to be updated to reflect the changes.

jrobrien91:
Forgot I wrote a height test. Sorry about that!

zssherman (approved):

PR #1216, opened 2022-07-22
fix writing IRIS sigmet files to UF

wx4stg:
When a radar object originally read in from sigmet file generated by IRIS is written to UF using `pyart.io.write_uf`, the range_start_m and range_step vars are single-object arrays. When these are subsequently rounded to integers by `round()`, the arrays give a traceback.

I fixed this by checking to see if these values were arrays, and if so, accessing the first value of these arrays. There may be a better implementation for fixing this, I'm not sure. I've attached an example IRIS sigmet file here (unzip the file first): [ADR220322013706.zip](https://github.com/ARM-DOE/pyart/files/9172328/ADR220322013706.zip)

### DIFF
--- a/pyart/io/uf_write.py
+++ b/pyart/io/uf_write.py
@@ -247,6 +247,10 @@ class UFRayCreator(object):
         range_start = self.radar.range['meters_to_center_of_first_gate']
         range_start -= (range_step / 2.)  # range bin center to edge
         range_start_km, range_start_m = divmod(range_start, 1000)
+        if isinstance(range_start_m, np.ndarray):
+            range_start_m = range_start_m[0]
+        if isinstance(range_step, np.ndarray):
+            range_step = range_step[0]
         header['range_start_km'] = int(range_start_km)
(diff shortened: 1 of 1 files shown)


zssherman:
Thanks for the PR @wx4stg, seems like a reasonable solution to me, is there any reason why range step might vary?

wx4stg:
I’m not sure honestly.

From a software perspective, the way the code is currently written, it expects a single value for these. In the sigmet files, it still produces a single value, just a float in a single-value ndarray. I have not yet come across a sigmet file with multiple values in this array, but I “only” have access to one radar so... not a terribly large sample size. 

From a hardware perspective, the start range should be related to how fast the radar’s duplexer can switch from tx to rx. I’m not sure if that’s constant or not. I have no idea what creates “gate spacing”.

wx4stg:
Also I just realized y’all have a contributors’ guide and I didn’t read it until just now... oops

zssherman:
Gotcha, sounds good, tests pass, going to merge! Thanks again for the PR!

Issue #1095, opened 2022-04-01
pyart.util.join_radar not tiling the text in sweep mode (and perhaps others correctly)


scollis:
On making a radar from 8 sweeps the array for sweep_mode is shape (192,) rather than (8,32)

mgrover1:
Closed via #1218

PR #1218, opened 2022-07-24
FIX: join_radar also joins pulse repititon times iparam

wx4stg:
In the current version of py-ART, joining two radar files using `pyart.util.join_radar(rdr1, rdr2)` produces a radar object that contains PRT data only from `rdr1`. 

This commit appends the PRT data from `rdr2`

This is necessary because if the joined radar object is then exported to universal format using `pyart.io.write_uf`, the rays from the second radar object don't have PRT data, causing an out-of-bounds array traceback.

### DIFF
--- a/pyart/util/radar_utils.py
+++ b/pyart/util/radar_utils.py
@@ -160,6 +160,11 @@ def join_radar(radar1, radar2):
                 np.append(
                     radar1.instrument_parameters['number_of_pulses']['data'],
                     radar2.instrument_parameters['number_of_pulses']['data']))
+        if 'prt' in new_radar.instrument_parameters:
+            new_radar.instrument_parameters['prt']['data'] = (
+                np.append(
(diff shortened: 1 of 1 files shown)


scollis:
Just sliding on in and saying thank you @wx4stg for the PRs!

wx4stg:
it may be more desired to refactor all of L148 through 163 to instead of join iparams fields individually by name, use a for loop through took look through all of the iparams in `radar1`, and if `radar2` has the same parameter, append the arrays of data.

However if some of the parameters have only a single value instead of an array, this could produce unexpected/invalid results... I don't have much experience which filetypes provide which parameters and what's considered valid

zssherman:
@wx4stg Thanks for the fix! I agree, a loop to append and look through the params is ideal. What I could do is accept this for now, and i can look into creating an actually loop, testing on the file in the previous PR.

wx4stg:
yeah, that would also fix #1095

PR #1227, opened 2022-07-27
ADD: addition of subset_radar function in radar_utils

wolfidan:
Dear Py-ART team,

In this PR I propose the addition of a ~~cut_radar~~ subset_radar function from our MeteoSwiss fork of the library (thanks @jfigui). It cuts a radar object to a new domain specified by bounds in range, azimuth and elevation, and returns a new object. It can be useful to make processing faster when one is interested only in a certain subdomain around the radar (typically shorter ranges and only a certain sector).
This function is used a lot in our [pyrad](https://github.com/MeteoSwiss/pyrad) library, which is  a framework for real time radar processing that relies on PyART. 
I also added a short test function that checks if a cut radar object has indeed coordinates that match with the specified bounds.
Cheers,
Daniel W.

### DIFF
--- a/pyart/util/radar_utils.py
+++ b/pyart/util/radar_utils.py
@@ -104,6 +104,203 @@ def to_vpt(radar, single_scan=True):
     return
 
 
+def cut_radar(radar, field_names, rng_min=None, rng_max=None, ele_min=None,
+              ele_max=None, azi_min=None, azi_max=None):
+    """
+    Cuts the radar object into new dimensions
+    Parameters
+    ----------
+    radar : radar object
+        The radar object containing the data
+    field_names : str or None
+        The fields to keep in the new radar
+    rng_min, rng_max : float
+        The range limits [m]. If None the entire coverage of the radar is
+        going to be used
+    ele_min, ele_max, azi_min, azi_max : float or None
+        The limits of the grid [deg]. If None the limits will be the limits
+        of the radar volume
+    Returns
+    -------
+    radar : radar object
+        The radar object containing only the desired data
+    """
+    radar_aux = copy.deepcopy(radar)
+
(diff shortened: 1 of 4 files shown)


pyart/util/radar_utils.py
```diff
@@ -104,6 +104,203 @@ def to_vpt(radar, single_scan=True):
     return
 
 
+def cut_radar(radar, field_names, rng_min=None, rng_max=None, ele_min=None,
```
mgrover1, line 107: This all looks great! One small suggestion - what would you think about changing the name here to subset_radar?

mgrover1 (changes requested):
One small suggestion - other than that, looks great! Thanks for adding tests!

wolfidan:
Thanks for the feedback @mgrover1, it's done

wolfidan:
Could you help me with this failing check, I have honestly no idea what failed here in the website deploy?

mgrover1:
@wolfidan we updated the trmm_rsl feedstock, changing the path in our CI. I added the fix here, should fix the problem. Thanks for making those changes!

mgrover1 (approved):

zssherman:
@mgrover1 looks like it's still not finding trmm_rsl, the full path on condaforge has a weird extension now before the 4 and after the package name

zssherman:
And it's different between osx and Linux for some reason... hmmm

mgrover1:
I updated the TRMM rsl path - it was incorrect...

wolfidan:
Thanks for your help!

mgrover1:
@zssherman I think we can go ahead and merge - let's fix the TRMM_RSL for linux in a different PR

PR #1235, opened 2022-08-11
ENH: Composite Reflectivity Function 

dopplerchase:
Hello pyart'ers, 

Since moving out west, the low level radar scans show lack luster echoes. Often times I use composite reflectivity on radar scope for quick looks at storms. One day, I wanted to plot composite Z with pyart, thinking there was a function to do this, but upon looking I could not find one. Hence I sought out to do this with some code. The trick I took here was sorting each sweeps azimuth such that the order of the matrix is the same for every single sweep. What this enables is the .max() call on the reordered matrix at the end.

Spot tests seem to work, but a full quantitative analysis has not been done.


tests/retrieve/test_comp_z.py
```diff
@@ -0,0 +1,72 @@
+""" Unit Tests for Py-ART's retrieve/qvp.py module. """
+
+import numpy as np
+from numpy.testing import assert_equal,assert_array_equal
+
```
mgrover1, line 5: I think you need another import copy here

mgrover1:
@dopplerchase I think you need to add it to the `pyart/retrieve/__init__,py` file to expose that function

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+Calculate the composite reflectivity 
+
+"""
+
+import copy
+import numpy as np
+from netCDF4 import num2date
+from pandas import to_datetime
+
+def composite_reflectivity(radar,rhv_filter=True,rhv_value=0.95,verbose=False):
```
mgrover1, line 11: I would make this more generic to take in any gatefilter, and if there isn't one, you can create one with the parameters you mentioned (for rho_hv)

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+    
+    """
+    Composite Reflectivity 
+    
+    Often a legacy product, composite reflectivity is: 
+    "A display or mapping of the maximum radar reflectivity factor at any altitude as a function of position on the ground." - AMS Glossary
```
mgrover1, line 17: Can you add citation here?

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+    to the physical spinning of the radar antenna. 
+    
+    Author: Randy J. Chase (@dopplerchase) 
+    
+    Note: this function could potenially be parallelized at the loop, but for now runs in ~5 seconds on my 16 GB Mac
+    
+    Parameters
+    ----------
+    radar : Radar
+        Radar object used.
```
mgrover1, line 33: specify pyart.core.Radar

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+    rhv_filter : bool
+        boolean flag to turn on correlation coef. filtering. 
+    rhv_value : float
+        value to consider 'meteorological'
+    verbose : bool
+        boolean flag to turn the printing of radar tilts on or off. 
+
+    Returns
+    -------
+    comp_dict : Dictonary
```
mgrover1, line 43: Can you have it return the field dict? So we can add it back to the radar object?

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+    
+    #loop over all measured sweeps 
+    for sweep in radar.sweep_number['data']:
+    
+        #get start and stop index numbers
+        s_idx = radar.sweep_start_ray_index['data'][sweep]
+        e_idx = radar.sweep_end_ray_index['data'][sweep]+1
+        
+        #grab radar data 
+        z = radar.get_field(sweep,'reflectivity')
```
mgrover1, line 57: Make sure to add a `reflectivity_field` argument so the user can specify this in the function

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+        
+        #filter out values if desired. Future we should use gate filters...
```
mgrover1, line 59: Yup! Making this a gate filter would be ideal

pyart/retrieve/comp_z.py
```diff
@@ -0,0 +1,109 @@
...
+        lat = lat[az_ids]
+        
+        #if the first sweep, store re-ordered lons/lats
+        if sweep == 0:
+            lon_0 = copy.deepcopy(lon)
+            lon_0[-1,:] = lon_0[0,:]
+            lat_0 = copy.deepcopy(lat)
+            lat_0[-1,:] = lat_0[0,:]
+        
+        #if 360 scan, upsample to super res
```
mgrover1, line 85: Why 720? Let's make this a parameter too - mentioning that default it will be at "super resolution" or 0.5 degrees

mgrover1 (changes requested):
Thanks @dopplerchase for putting this together!! This PR looks great - just a few comments and feedback. Let me know if you need any help it, including an example!

It would be nice to have an example in this PR too.

I think users are going to be **very excited** about this new functionality. Super helpful to the community, and thanks for adding the tests.

pyart/retrieve/comp_z.py
dopplerchase: Not quite because this is only 1 sweep. So the size of the matrix doesnt match any other field

mgrover1: Ohh good point... what if it returned a separate radar object? With a single field?

mgrover1:
@dopplerchase - I can help with some of these changes if you'd like (especially integrating the gatefilter in here). Is that something you would be interested in?

dopplerchase:
Yeah I have 0 experience implementing a gatefilter. It would be best if you can help with that

mgrover1:
> Yeah I have 0 experience implementing a gatefilter. It would be best if you can help with that

Happy to help :)

pyart/retrieve/comp_z.py
mgrover1: Nevermind! Using a dictionary would make sense and is consistent!

zssherman: Discussed this as the shape and info of the radar object changes, will go for radar object return.

pyart/retrieve/comp_z.py
zssherman: done

zssherman:
@dopplerchase With going the radar object route, with composite reflectivity, what typically is used for other fields for this data. Example, what would fixed angle be set to? Time array? etc just trying to figure out what to set these fields to, and not entirely familiar what is used as the defaults for a composite field.

mgrover1:
@zssherman just let me know when this is ready for a review!

zssherman:
@mgrover1 I still need to do the radar object return, was waiting on dopplerchase for the response on attributes and which values to use for time, azimuth etc, if we are doing one sweep we can return a radar.extract_sweeps.

dopplerchase:
Hey sorry guys, busy time for me right now. As for the attributes, could you elaborate more on what you need? or point me to an example?

zssherman:
@dopplerchase No worries, my question more pertained, if we return a radar object, for a composite radar, should the fixed angle be 0? elevations etc, wasn't sure how these values get defined for a composite field, or if we take the mean of these fields? Because one way, i this is one sweep, is just use the radar.extract_sweeps 0 sweep and have a radar set up for that, however if we are setting the azimuth using 360 or 720 if it will differ from the nrays of one sweep. Can we define the resolution from instrument parameters so it matches the original radar, or is it better to allow the user to set that?

mgrover1:
@zssherman how can I help here? I think this would be great to include in the upcoming release!

zssherman:
@mgrover1 I think the only thing left is plugging everything into a radar class similar to what's done in the ios like the nexrad or cfradial readers. I'm trying to remember what Scott said on Thursday of values to use. Think fixed angle as 0.

mgrover1:
> @mgrover1 I think the only thing left is plugging everything into a radar class similar to what's done in the ios like the nexrad or cfradial readers. I'm trying to remember what Scott said on Thursday of values to use. Think fixed angle as 0.

Yeah! Fixed angle of 0, treating it as a single sweep of a radar

zssherman:
@mgrover1 Tests failed due to a newer matplotlib version, I fixed it, but we will have to update the radarmapdisplay and gridmapdisplay plot_colorbar to make ax as a parameter and not optional.

mgrover1:
Okay - @zssherman @dopplerchase this should be ready to go. I added in the example in the last commit, which resulted in the following image:

test_image

zssherman (approved):

zssherman:
@mgrover1 only thing left i see is the reference

pyart/retrieve/comp_z.py
mgrover1: On second thought, I don't see a well-defined (or citable) definition out there...

mgrover1:
@zssherman after searching for a bit, I could not find a good reference. @dopplerchase any thoughts here? If we can't find one, I think we can go ahead and merge.

mgrover1 (approved):

Issue #1293, opened 2022-10-20
radardisplay does not recognize c-sapr "sector" plot type 


mvanlierwalq:
tryping to plot the csapr file houcsapr2cfrS2.a1.20220617.190001.nc, display.plot throws an error owing to unknown plot type "sector" : `ValueError: unknown scan_type sector`
can this be added to pyart?

mgrover1:
Yeah... currently "sector" is not one of these options, but we can consider adding it! Treating it the same as a PPI... which I think would be the right assumption @mvanlierwalq

zssherman:
Yeah agreed, it should be a simple fix

mvanlierwalq:
@zssherman thanks! will have to wait until tonight or tomorrow, having trouble connecting to my computer. will let you know.

mgrover1:
@mvanlierwalq any updates here? Did the fix work?

zssherman:
fixed in #1294 can be reopened if need be.

PR #1294, opened 2022-10-20
ENH: Check for sector scans and use ppi.

zssherman:


zssherman:
closes #1293

pyart/graph/radardisplay.py
```diff
@@ -134,7 +134,7 @@ def plot(self, field, sweep=0, **kwargs):
         plot_vpt : Plot a VPT scan
 
         """
-        if self.scan_type == 'ppi':
+        if (self.scan_type == 'ppi' or self.scan_type == 'sector'):
```
mgrover1, line 137: I like this approach!

mgrover1 (approved):

PR #1306, opened 2022-10-28
FIX: Add check for when instrument parameters is None.

zssherman:


mgrover1 (approved):
This seems like a reasonable fix! Thanks @zssherman

PR #1313, opened 2022-11-04
FIX: Phase processing code throwing deprecation warnings for scipy.ndimage.filters.convolve1d

rcjackson:
The phase processing code is throwing deprecation warnings for using scipy.ndimage.filters.convolve1d. It looks like that function is moving to scipy.ndimage.convolve1d. This fix would take care of the deprecation warning.


zssherman (approved):

zssherman:
Thanks Bobby for putting this together! Merging!

Issue #763, opened 2018-08-23
Add a method to apply a gatefilter to fields to mask the data


scollis:
At the moment the way I apply a gatefilter to mask (ie when I want to save) data is as follows:
```gf = pyart.filters.GateFilter(radar)
gf.exclude_below('reflectivity', 10)
nf = deepcopy(radar.fields['reflectivity'])
nf['data'] = np.ma.masked_where(gf.gate_excluded, nf['data'])
radar.add_field('filtered_refectivity', nf, replace_existing=True)```

It would be great to have a method that automates this...

something like:

radar.apply_filter(gatefilter, replace=False, fields=None)

If replace = False it will append 'filtered_'+field with areas where gatefilter.gate_exluded set to _FillValue
if replace = True it does it in-place
if fields = None it does it to all fields 

Inspired by Brenda Dolan

zssherman:
Self assigned, will look into this.

zssherman:
Working on this now

PR #1326, opened 2022-11-14
ADD: Adding add_filter function to radar.py.

zssherman:
This function applies a filter to fields and either replaces or adds new filtered fields.

Note, adding tests to this PR, just initial review.

Closes, #763

### DIFF
--- a/pyart/core/radar.py
+++ b/pyart/core/radar.py
@@ -788,6 +788,52 @@ class Radar(object):
         self.fields[field_name] = dic
         return
 
+    def add_filter(self, gatefilter, replace_existing=False,
+                   include_fields=None):
+        """
+        Updates the radar object with an applied gatefilter provided
+        by the user on fields within the radar object.
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


pyart/core/radar.py
```diff
@@ -788,6 +788,52 @@ def add_field(self, field_name, dic, replace_existing=False):
         self.fields[field_name] = dic
         return
 
+    def add_filter(self, gatefilter, replace_existing=False,
+                   include_fields=None):
+        """
+        Updates the radar object with an applied gatefilter provided
```
mgrover1, line 794: Can you add more here, stating that it masks values?

mgrover1 (commented):
One minor comment - otherwise, this looks great!

pyart/core/radar.py
zssherman: Sounds good!

zssherman:
@mgrover1 Added unit tests as well now for review.

Issue #1336, opened 2022-11-24
Error with gamic file having different bin numbers depending on elevation


tsbiscaro:
I am reading a GAMIC file with 600 gates in the first 5 elevations and 800 in the last 3. This results in an error as below:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/pyart/aux_io/gamic_hdf5.py", line 204, in read_gamic
    field_dic['data'] = gfile.moment_data(group, 'float32')
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/pyart/aux_io/gamicfile.py", line 142, in moment_data
    data[start:end+1, :sweep_data.shape[1]] = sweep_data[:]
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/numpy/ma/core.py", line 3387, in __setitem__
    _data[indx] = dval
ValueError: could not broadcast input array from shape (361,800) into shape (361,600)

The code is using the first sweep dimension (361, 600) to all sweeps, leading to this error when a sweep have a different azimuth/bin count than the first sweep. This is defined in gamicfile.py line 135.

ngates = int(self._hfile['/scan0/how'].attrs['bin_count'])

kmuehlbauer:
@tsbiscaro You might not get an answer today  🦃 .

So the reader would need to get the max bin number and use this for the creation of the array.

But this only works if the range resolution  is same in all sweeps.

If you only want to have a look at the data you might consider trying out

https://docs.openradarscience.org/projects/xradar/en/latest/

kmuehlbauer:
@tsbiscaro You might try the solution proposed in #1337. It's not as geeky as the PR number suggests, though.

tsbiscaro:
Thank you! Didn't know about xradar, I will take a look at it.

zssherman:
Fixed in #1337, closing

PR #1337, opened 2022-11-24
FIX: use max_num_gates for calculation of data shape

kmuehlbauer:
solves #1336 

But only if range resolution is same for all sweeps.

### DIFF
--- a/pyart/aux_io/gamicfile.py
+++ b/pyart/aux_io/gamicfile.py
@@ -40,6 +40,8 @@ class GAMICFile(object):
         self._scans = ['scan%i' % (i) for i in range(self.nsweeps)]
         self.rays_per_sweep = self.how_attrs('ray_count', 'int32')
         self.total_rays = sum(self.rays_per_sweep)
+        self.gates_per_sweep = self.how_attrs('bin_count', 'int32')
+        self.max_num_gates = max(self.gates_per_sweep)
         # starting and ending ray for each sweep
         self.start_ray = np.cumsum(np.append([0], self.rays_per_sweep[:-1]))
         self.end_ray = np.cumsum(self.rays_per_sweep) - 1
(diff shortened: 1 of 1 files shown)


zssherman (approved):
This looks good @kmuehlbauer! Thanks for putting this together!

tsbiscaro:
#1337 

That works for reading the data, but it will crash during the plot (I am trying to plot the first elevation):

 File "dual_pol.py", line 15, in <module>
    display.plot('corrected_reflectivity', 0, vmin=20, vmax=70., colorbar_label='dBZ', axislabels_flag=False)
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/pyart/graph/radardisplay.py", line 138, in plot
    self.plot_ppi(field, sweep, **kwargs)
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/pyart/graph/radardisplay.py", line 337, in plot_ppi
    pm = ax.pcolormesh(
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/matplotlib/__init__.py", line 1423, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/matplotlib/axes/_axes.py", line 6133, in pcolormesh
    X, Y, C, shading = self._pcolorargs('pcolormesh', *args,
  File "/home/tbiscaro/anaconda3/envs/radar/lib/python3.8/site-packages/matplotlib/axes/_axes.py", line 5661, in _pcolorargs
    raise TypeError('Dimensions of C %s are incompatible with'

A sample file can be downloaded at: https://1drv.ms/u/s!AjzwFqty9OKPiPwNlxYAGzNyHDoLfQ?e=WUk03I

kmuehlbauer:
Yes, what a pity, I've missed to fix the range as well. I'll add the needed changes.

Thanks for the data file. I can confirm, that the range resolution changes in your file between sweeps. That contradicts the current Py-ART CfRadial1-based datamodel. 

@zssherman @mgrover what's the current workaround in that case?

tsbiscaro:
Yes, the range varies too. I've contacted the radar operator to change this because it doesn't make sense to me, but I've sent the file anyway as a good test file for reading a really bad-formatted file...

kmuehlbauer:
I would not say it's bad formatted. It makes sense to decrease the range resolution with increasing elevation which results in better height resolution for higher angles. That's one reason CfRadial2 was invented, a data model also Py-Art will switch to, with the help of `xradar`.

PR #1338, opened 2022-11-28
FIX: check uniformity of range_step, get consistent range_step for gamic reader

kmuehlbauer:
follow-up to #1337


mgrover1 (approved):
Great! Thanks @kmuehlbauer

PR #1339, opened 2022-12-01
DOC: Update convective stratiform example

lauratomkins:
There were a couple formatting things that I wanted to update in my example. I also realized that the `axislabels_flag=False` argument in `pyart.graph.GridMapDisplay.plot_grid` doesn't work as intended with a Lambert Conformal projection (I'm not sure if this is already a known issue or not), so I updated to a different projection to make the figures easier to read.


mgrover1 (approved):
Good call moving the references down

PR #1356, opened 2022-12-13
MNT: Revert setuptools, to avoid unwanted behavior.

zssherman:
Need to look into the difference more on find_namespace_packages and find_packages. Trying to avoid similar install error behavior as seen in ACT.  I switched to avoid warnings during the feedstock testing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ import subprocess
 import glob
 from numpy import get_include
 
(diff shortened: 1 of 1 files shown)


zssherman:
@mgrover1 I believe this should fix your __init__.py pep8 errors in the formatting PR.

mgrover1 (approved):
Great! Thanks @zssherman

PR #1379, opened 2023-01-19
ENH: Small correction to GateMapper example for the API

jrobrien91:
Quick fix to the GateMapper example for the API.


jrobrien91:
credit goes to @rcjackson for catching this

mgrover1 (approved):
Great catch - waiting for tests to pass

PR #1380, opened 2023-01-23
REL: fix build for Python 3.11 wheels

neutrinoceros:
just noticed that the latest versions didn't include 311 wheels, despite #1303
This should fix it for next time.


mgrover1:
Good catch @neutrinoceros !

mgrover1 (approved):

PR #1381, opened 2023-01-24
FIX: A fix for vpt mismatch array error for matplotlib shading.

zssherman:


mgrover1 (changes requested):
Can you add a test for this function @zssherman

zssherman:
Ah yep, good point!

mgrover1 (approved):

PR #1394, opened 2023-03-06
drop pkg_resources as a runtime requirement

neutrinoceros:
`pkg_resources` is deprecated https://setuptools.pypa.io/en/latest/history.html#v67-5-0

Note to reviewers: it's not perfectly clear to me which versions of Python are still supported as various sources differ:
- from PyPI classifiers: 3.6-3.10
- from 1.14.5 wheels: 3.8-3.11
- from test matrix: 3.9-3.11

Are the classifiers are just out of date ? I did the maximal effort to keep compatibility with Python<3.8 here, but if 3.9 is already the minimal supported version this PR can be simplified.

### DIFF
--- a/pyart/__init__.py
+++ b/pyart/__init__.py
@@ -22,7 +22,7 @@ from os import environ as _environ
 if "PYART_QUIET" not in _environ:
     print(_citation_text)
 
-from pkg_resources import DistributionNotFound, get_distribution
+import importlib.metadata as _importlib_metadata
 
 # import subpackages
 # print out helpful message if build fails or importing from source tree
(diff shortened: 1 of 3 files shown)


neutrinoceros:
I see that ruff+pyupgrade is targeting Python 3.9+, so let's just run with it

zssherman:
@neutrinoceros Thank you for the pull request! I think at this point we are trying to follow conda-forge etc on current testing and releases. So usually the 3 newest python versions I think or whichever it is. I'll let @mgrover1 provide some input as well. But with that being said, this PR looks good to me! I think yeah we need to have a more universal Python standard for versions between things. And this looks like it would fix the classifiers for PyPI. But I'll check the wheel build as well.

zssherman (approved):

mgrover1 (approved):

PR #1395, opened 2023-03-06
CI: Remove python 3.8 and update documentation to reflect releases and CI testing.

zssherman:


mgrover1 (approved):
I agree here - it follows the associated scientific-python SPEC https://scientific-python.org/specs/spec-0000/

PR #1421, opened 2023-04-17
fix: memory leak in grid mappers

### LINKED ISSUE #1420: BUG: memory leak in grid mappers due to incomplete garbage collection
* Py-ART version: 1.15
* Python version: 3.10.6
* Operating System: Ubuntu 22.04

### Description
@mgrover1

As far as I can remember, there is a memory leak in the grid mappers of Pyart.

Sequential runs of `grid_from_radars` for example, that calls `map_gates_to_grid`, that calls `GatesToGridMapper` in `_gate_to_grid_map.pyx` compiled in C, iteratively fill the memory (RAM) to a point where the computer crashes.

If one wants to produce cartesian grids for every timestep of a day with different radars at a certain grid resolution and grid domain, it ultimately crashes because every grid produced is not entirely freed from memory.

### What I Did
By exploring the matter, it is highly probable that the garbage collector of Python (that attributes / freeds the memory of instances) doesn't freed all the instances that are in memory when using `map_gates_to_grid`.

This could mainly come from:
- instances (i.e variables) in the code that have references to themselves. Because of that, the reference count of the instance is never zero and the garbage collector always keeps that instance in memory.

## The Fix
The fix I've found it to simply call the Python garbage collector at the end of the `map_gate_to_grid` function in `gates_to_grid.py`.

```
import gc

def map_gates_to_grid(...):
    [...]
    if map_roi:
        roi_array = np.empty(grid_shape, dtype=np.float32)
        gatemapper.find_roi_for_grid(roi_array, roi_func)
        grids["ROI"] = roi_array

    gc.collect()
    return grids
```

Vforcell:
- [x] Closes #1420
- [x] Tests added (--> No additional tests needed)

### DIFF
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -2,7 +2,7 @@
 Generate a Cartesian grid by mapping from radar gates onto the grid.
 
 """
-
+import gc
 import warnings
 
 import numpy as np
@@ -185,6 +185,8 @@ def map_gates_to_grid(
         roi_array = np.empty(grid_shape, dtype=np.float32)
         gatemapper.find_roi_for_grid(roi_array, roi_func)
         grids["ROI"] = roi_array
+
+    gc.collect()
     return grids
 
 


zssherman (approved):

mgrover1 (approved):
Great! Thanks for the contribution here @Vforcell 😄

Vforcell:
@zssherman I see a linting error for import orders.

See : `pyart/map/gates_to_grid.py:6:1: I001 [*] Import block is un-sorted or un-formatted`

What is the convention in PyArt ? Do you use a Python dependency in top of black or flake8 ?

pyart/map/gates_to_grid.py
```diff
@@ -6,6 +6,7 @@
 import warnings
 
 import numpy as np
+import gc
```
mgrover1, line 9: Can you move this above warnings? It belongs in that import block.

mgrover1 (changes requested):
One change here

mgrover1:
> @zssherman @mgrover1 I see a linting error for import orders.
> 
> See : `pyart/map/gates_to_grid.py:6:1: I001 [*] Import block is un-sorted or un-formatted`
> 
> What is the convention in PyArt ?
> 
> **Edit:** is there a way to run the CI linting part locally on my Ubuntu 22.04 computer ?

Ya - you can run pre-commit install from the root Py-ART directory
```
pre-commit install
```

then
```
pre-commit run
```

This will also be applied every time you commit changes - running linting and fixing where possible

mgrover1 (approved):
Once the rest of the tests pass, this is good to go!

mgrover1:
Merging!

Vforcell:
@mgrover1 question : are fixes available directly through the conda installation ? Or does it need to be released and tagged first ?

mgrover1:
You can install the unreleased version using 
'''
pip install git+https://github.com/ARM-DOE/pyart
'''

You'll want to make sure uninstall the conda version first.

PR #1446, opened 2023-08-14
ENH: optional rectangular window dims for velocity texture analysis and bug fixes to the load_kazr method

isilber:


mgrover1 (changes requested):
This looks great @isilber can you add/modify the test(s) here?

isilber:
I added two tests for the texture method:
1. Same as the other tests verifying a value of 0, but using a rectangular filter dim input.
2. Consistency check of texture analysis using an input scalar and an equivalent tuple input applied to a noisy velocity field.

mgrover1 (approved):
This looks good to merge!

PR #1457, opened 2023-09-15
Update accessor.py

syedhamidali:
The function claims to return ```x, y, and z``` values, but it actually returns ```x, y, and x``` values. Is this discrepancy unintentional? If it's an error, please feel free to accept this pull request. If not, I apologize for any confusion arising from my pull request.


mgrover1 (approved):
Good catch - thanks for the fix here @syedhamidali !

PR #1477, opened 2023-10-20
ENH: xarray grid output

syedhamidali:
- [x] Implementation of https://github.com/ARM-DOE/pyart/discussions/1476#discussion-5735867

To streamline the process of extracting a subset from `pyart.core.grid` or the grid object, I propose to add a few more variables to the `xarray dataset` that mirrors the grid object. When this `xarray dataset` is written to a `NetCDF` file, it becomes equivalent to the original `pyart.core.grid` object.

How can this be utilized?

```python
import pyart

# Reading a grid from a netcdf file
grid = pyart.io.read_grid("file.nc")

# Converting the grid to an xarray dataset
ds = grid.to_xarray()

# Slicing the dataset based on x and y coordinates
ds = ds.sel(x=slice(5000, 70000), y=slice(5000, 70000), drop=True)

# Saving the sliced dataset to a new netcdf file
ds.to_netcdf("outfile.nc")
```

Subsequently, it is possible to read the modified dataset using `pyart` as follows:

```python
# Reading the modified dataset with pyart
grid_small = pyart.io.read_grid("outfile.nc")
```

This enhances the process of working with grid data, making it more convenient and efficient for users.

### DIFF
--- a/pyart/core/grid.py
+++ b/pyart/core/grid.py
@@ -316,11 +316,6 @@ class Grid:
 
         """
 
-        if not _XARRAY_AVAILABLE:
-            raise MissingOptionalDependency(
-                "Xarray is required to use Grid.to_xarray but is not " + "installed!"
-            )
-
         lon, lat = self.get_point_longitude_latitude()
         z = self.z["data"]
         y = self.y["data"]
@@ -348,6 +343,7 @@ class Grid:
                     data.attrs.update({meta: self.fields[field][meta]})
 
             ds[field] = data
+
             ds.lon.attrs = [
                 ("long_name", "longitude of grid cell center"),
                 ("units", "degree_E"),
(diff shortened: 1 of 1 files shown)


pyart/core/grid.py
```diff
@@ -326,47 +326,112 @@ def to_xarray(self):
...
+                        attr_value, dims=dims, attrs=get_metadata(attr_name)
+                    )
+
+        if "radar_time" in ds.variables:
+            ds.radar_time.attrs.pop("calendar")
+
+        if self.radar_name is not None:
+            radar_name = self.radar_name["data"]
+            ds["radar_name"] = xarray.DataArray(
+                np.array([b"".join(radar_name)]),
```
mgrover1, line 428: What is the reason for using a byte-encoded name here?

pyart/core/grid.py
```diff
@@ -326,47 +326,112 @@ def to_xarray(self):
...
+            attrs=_make_coordinatesystem_dict(self),
+        )
+
+        # write the projection dictionary as a scalar
+        projection = self.projection.copy()
+        # NetCDF does not support boolean attribute, covert to string
+        if "_include_lon_0_lat_0" in projection:
+            include = projection["_include_lon_0_lat_0"]
+            projection["_include_lon_0_lat_0"] = ["false", "true"][include]
+        ds["projection"] = xarray.DataArray(
```
mgrover1, line 387: I think this variable, along with the other origin_latitude, etc. variables belong as coordinates - can you add those to the list of coordinates? I think that is what is causing the failure in CI here.

mgrover1 (changes requested):

pyart/core/grid.py
syedhamidali: @mgrover1 My apologies for the late response; I've been tied up with exams recently and have another busy week ahead. I'll get to it as soon as possible. Thanks for your suggestion.

mgrover1: No worries! Take your time - just let me know when you are ready for another review!

pyart/core/grid.py
syedhamidali: Because some radar data use the following convention:
```python
print(grid.radar_name)
```
```python
{'long_name': 'Name of radar used to make the grid',
 'data': masked_array(data=[[b'D', b'O', b'W', b'8']],
              mask=False,
        fill_value=b'N/A',
             dtype='|S1')}
```
To make it look like `b'DOW8`, because if you read a `pyart` or `radx` generated `grid` using `xarray`, it shows up like `array([b'DOW8'], dtype='|S4')`. That's why, I am using byte-encoded name so the result looks like

```python
ds["radar_name"] = xarray.DataArray(
                np.array([b"".join(grid.radar_name)])
print(ds["radar_name"])
```
```python
array([b'DOW8'], dtype='|S4')
```

syedhamidali:
Hey @mgrover1,

Just wanted to check in on the status of #1477. I believe we’ve completed the function that allows us to directly write xarray grids using `ds.to_netcdf` method. This functionality should enable us to seamlessly read these grids back in using Py-ART’s grid reader. Let me know if you have any updates or if there are any additional steps required.

pyart/core/grid.py
mgrover1: @syedhamidali - did you address this feedback already?

mgrover1:
@syedhamidali - can you please review the previous comment to ensure that is addressed to fix the failing CI?

pyart/core/grid.py
syedhamidali: @mgrover1 Yes, added to coords

syedhamidali:
> @syedhamidali - can you please review the previous comment to ensure that is addressed to fix the failing CI?

Resolved in [this commit](https://github.com/ARM-DOE/pyart/pull/1477/commits/5eb6dd67258ae8a79148b6508a34c56402d0f7e0)

mgrover1:
Fantastic! Thanks @syedhamidali 👍

RBhupi:
@syedhamidali Are you still working on this PR? Could you please resolve the conflicts?

syedhamidali:
Hi @RBhupi, great to reconnect after some time. Since @rcjackson has already implemented the necessary changes, this PR is no longer needed. We can go ahead and close it.

RBhupi:
@syedhamidali Thank you!

PR #1486, opened 2023-11-07
ADD: function to rescale reflectivity to a precipitation rate

lauratomkins:
I added a function that I found myself needing a lot. It rescales radar reflectivity to precipitation rate. The user can change the coefficient and exponent to customize which Z-R or Z-S relationship they would like to use. Since the function will work with both Z-R and Z-S relationships, I named it as a generic "precipation rate". Please let me know if you have any feedback or any suggested changes.


zssherman:
Hi @lauratomkins thanks for the PR! A few things, could we move this function to the qpe module in retrievals? The failing tests I believe is due to math in the unit testing, I would try to use numpy.log etc instead?

lauratomkins:
@zssherman any ideas on why this is failing on some of these builds?

zssherman:
@lauratomkins Hmm yeah not too sure, I'll look into it, it's odd that it's only failing on certain Ubuntu builds, wondering if we need np.ma to handle the nans possibly? I can keep digging to see what I can find

zssherman:
Yeah seems we are getting nans somehow from the np.log10 on different systems, my guess was maybe a precision issue, but my attempted fix sadly did not work..

mgrover1:
@lauratomkins - I pushed some changes that should help the testing suite here. I think the key will be to use the `np.testing.assert_allclose` testing functionality which will ensure values are within a certain threshold, similar to the floor command you used here.

mgrover1:
@zssherman this fixed the issues! Once the tests pass, I think we can move ahead and merge! Thanks for all the hard work here @lauratomkins !!

mgrover1 (approved):
Great! A useful functionality for the community!

PR #1487, opened 2023-11-13
ENH: updated feature detection function and example

lauratomkins:
Hi all, I've renamed and updated the convective stratiform functions that I added a while back. I've been adapting these functions to detect features in winter storms so I removed most of the "convective" references in the code and renamed it to "feature detection" to make it more generic. I added included a few more features to the function itself and I updated the example (reorganized and added more examples). Let me know what changes I should make. Thanks!


mgrover1:
Since this is an API change (the function name is being deleted + changed), I would argue for keeping in the original function, passing the arguments to the new function, and adding a warning this will be deprecated in Py-ART 2.0. The major version usually indicates breaking changes... what are your thoughts here @zssherman ?

mgrover1 (changes requested):
As I mentioned in the comment, I think adding this as a new function, and keeping the original function in here until Py-ART 2.0 would make the most sense since this will break functionality. We can keep the `plot_convective_stratiform` example, and add the plot_feature_detection as well, adding a warning at the top of the plot_convective_stratiform that it will deprecated soon.

zssherman:
@mgrover1 I think that's a good path forward

lauratomkins:
@mgrover1 , I implemented the changes you requested. Does it look okay?

mgrover1 (approved):
This looks great - thank you for making those changes!

PR #1492, opened 2023-12-05
ADD: function to create CFADs

lauratomkins:
Hi folks,

I have another PR which adds functionality to create CFADs with radar data. I've included an example as well. Not sure if my functions are in the right place so please let me know if you have any feedback!

Thanks!


zssherman:
@lauratomkins Thanks for the PR! I will take a look shortly, Max will likely as well soon.

mgrover1 (changes requested):
This looks fantastic! My only suggestion here would be to pass in radar object as input for the cfad function, similar to the composite reflectivity function https://github.com/ARM-DOE/pyart/blob/main/pyart/retrieve/comp_z.py#L16

Then have a `field="reflectivity"` argument by default; this will make it easier for users to plug in their radar data, as well as being consistent with submodules.

lauratomkins:
@mgrover1 thanks for the suggestion. Ideally, I'd like this function to work with both radar and grid objects. I started writing the function in a way that would accept both, but I'm wondering if it makes more sense to have two functions, one for radar objects and one for grid objects? Do you have any other functions that can work on both types of objects that I can use as a reference? Or do you have any suggestions on the best way to handle this?

zssherman:
@lauratomkins Max is currently out at a deployment, why it's been slightly quiet with this PR. I'm wondering too, if we can just have a pyart object, without having to have one for the grid and one for a radar object. The field data is accessed I believe the same if we called radar.fields['reflectivity']['data'] and grid.fields['reflectivity']['data'] so might be able to make it still universal if we are going the route Max suggested while still maintaining one function. We can see what else Max thinks once he returns.

mgrover1:
I agree with @zssherman here; as long as the object is able to access `radar.fields[reflectivity_field]`  it should be good.

lauratomkins:
Apologies for the delay but I think I've updated it in a way that makes sense. Please let me know if it needs any more changes. Also, I'm not sure how to discard a commit, but I believe that my changes to the `plot_feature_detection.py` file is causing merging issues. I believe the issue that I was trying to correct was fixed so it can be discarded.

zssherman (approved):

pyart/retrieve/cfad.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # get altitude data
+    # first try to get altitude data from a radar object
+    try:
+        altitude_data = radar.gate_z["data"]
+    # if it fails, try to get altitude data from a grid object
+    except:
+        try:
+            altitude_data = radar.point_z["data"]
+        except:
```
zssherman, line 66: We might want to do a full error catching here like a ValueError or something similiar. @mgrover1 thoughts?

pyart/retrieve/cfad.py
```diff
@@ -0,0 +1,93 @@
...
+        List of bin edges for height values to use for CFAD creation.
+    field : str
+        Field name to use to look up reflectivity data. In the
+        radar object. Default field name is 'reflectivity'.
+    field_mask : array
+        An array the same size as the field array used to mask values.
+    min_frac_thres : float, optional
+        Fraction of values to remove in CFAD normalization (default 0.1). If an altitude row has a total count that
+        is less than min_frac_thres of the largest number of total counts for any altitude row, the bins in that
+        altitude row are masked.
```
zssherman, line 36: Space line between the headers of each section, Returns, references etc I think is needed for sphinx to generate correctly.

zssherman:
@lauratomkins I added just to more comments, besides that I think this is almost good to go! I'll let @mgrover1 do a final review on my suggestions and see if he has any more thoughts.

mgrover1 (approved):
This looks great! Thanks for making those changes.

zssherman (approved):

PR #1496, opened 2023-12-15
ADD: Transparency options for geotiff writer

qubitqualia:
Added options to the geotiff writer that allow RGB images to be created with a transparent background. An opacity setting was also added to control the overall transparency of the output image.

- [X ] Tests added
- [ X] Documentation reflects changes


zssherman:
@qubitqualia Thanks for the PR! Looks reasonable to me! Think just some PEP8 fixes are needed for the nonewline etc

qubitqualia:
Should be good now. This is my first pull request so thank you for your patience!

zssherman:
No worries! Rerunning the tests now

zssherman:
Looks like a few more PEP8 changes.

zssherman:
@qubitqualia I left a few more comments, and quick question, are we setting the default to transparent? Or are we setting it to normal prior to this PR and if a user wants to they can set it to transparent, I saw the transparent parameter set to true and wanted to make sure.

zssherman:
@mgrover1 I got the tests passing now, If you can do a review on this when you have time?

mgrover1 (approved):
This look good - thanks for the test here as well

zssherman (approved):

PR #1497, opened 2023-12-18
Update rainbow_wrl.py

HappyXiaoAnAn:
Prevent the program from crashing when the number of bins changes between sweeps.

In my rb5 dataset, the number of bins would change between sweeps. Therefore,
I adjusted the array size to match the maximum number of bins.
(I'm not sure whether this is appropriate or safe.)

### DIFF
--- a/pyart/aux_io/rainbow_wrl.py
+++ b/pyart/aux_io/rainbow_wrl.py
@@ -282,7 +282,7 @@ def read_rainbow_wrl(
         # all sweeps have to have the same range resolution
         for i in range(nslices):
             slice_info = rbf["volume"]["scan"]["slice"][i]
-            if slice_info['rangestep'] != common_slice_info["rangestep"]:
+            if slice_info["rangestep"] != common_slice_info["rangestep"]:
                 raise ValueError("range resolution changes between sweeps")
 
     r_res = float(common_slice_info["rangestep"]) * 1000.0
--- a/pyart/aux_io/rainbow_wrl.py
+++ b/pyart/aux_io/rainbow_wrl.py
@@ -282,7 +282,7 @@ def read_rainbow_wrl(
         # all sweeps have to have the same range resolution
         for i in range(nslices):
             slice_info = rbf["volume"]["scan"]["slice"][i]
-            if (slice_info['rangestep'] != common_slice_info["rangestep"]):
+            if slice_info['rangestep'] != common_slice_info["rangestep"]:
                 raise ValueError("range resolution changes between sweeps")
 
     r_res = float(common_slice_info["rangestep"]) * 1000.0
@@ -354,7 +354,7 @@ def read_rainbow_wrl(
             slice_info["slicedata"]["rawdata"],
             rays_per_sweep[i],
             nbins_sweep[i],
-            maxbin
+            maxbin,
         )
 
     if bfile.endswith(".vol") or bfile.endswith(".azi"):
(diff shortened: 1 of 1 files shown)


zssherman:
@HappyXiaoAnAn Thanks for the PR! Yeah, I'll have to think on whether it's how it should be handled or not. @mgrover1 thoughts? @mgrover1 when your not busy with the deployment stuff

mgrover1:
This looks good - we just need to fix the linting here!

HappyXiaoAnAn:
Thanks for the reply. I have fixed the linting. In addition, I have also added two properties, "meters_between_gates" and "meters_to_center_of_first_gate", required by the uf format file.

pyart/aux_io/rainbow_wrl.py
```diff
@@ -280,16 +284,18 @@ def read_rainbow_wrl(
     else:
         start_range = 0.0
     _range["data"] = np.linspace(
-        start_range + r_res / 2.0, float(nbins - 1.0) * r_res + r_res / 2.0, nbins
+        start_range + r_res / 2.0, float(maxbin - 1.0) * r_res + r_res / 2.0, maxbin
```
kmuehlbauer, line 287: @HappyXiaoAnAn @zssherman Heads up! This needs a check that range resolution is same for all sweeps. If not we have to raise an error as this can't be represented with the CfRadial1 data model.

HappyXiaoAnAn: Thanks for reminding! I would like to add some code to check the range resolution. Does this make sense?
```python
# range
if not single_slice:
    # all sweeps have to have the same range resolution
    for i in range(nslices):
        slice_info = rbf["volume"]["scan"]["slice"][i]
        if not (slice_info['rangestep'] == common_slice_info["rangestep"]):
            raise ValueError("range resolution changes between sweeps")
```

kmuehlbauer: That looks OK to me.

zssherman: @kmuehlbauer Good point! Thanks for the heads up!

mgrover1: That makes sense!! Thank you @HappyXiaoAnAn

HappyXiaoAnAn:
Hi @zssherman @mgrover1 , thank you for your comment. I have fixed the linting issues. As this is my first pull request on GitHub, I am not sure how to proceed with the remaining problem. Could you please guide me on what I should do next? Thank you very much.

mgrover1:
> Hi @zssherman @mgrover1 , thank you for your comment. I have fixed the linting issues. As this is my first pull request on GitHub, I am not sure how to proceed with the remaining problem. Could you please guide me on what I should do next? Thank you very much.

What else do you need help with? It looks great to me!

HappyXiaoAnAn:
Because there was another error during the CI workflow before I fixed the linting, I had thought that I missed something. However, everything works fine now, and there is no problem for me. Thank you.😊

zssherman:
@kmuehlbauer Wanted to check in with you, does this change look reasonable? I wanted to make sure this didn't break standards etc, but wasn't entirely sure...

mgrover1:
> @kmuehlbauer Wanted to check in with you, does this change look reasonable? I wanted to make sure this didn't break standards etc, but wasn't entirely sure...

He mentioned in the review above that it looked good to go, so let's go ahead and merge!

PR #1509, opened 2024-01-18
ENH: add 'zdist_factor' input parameter to 'map_gates_to_grid'.

isilber:
This parameter scales the distance in the z-dimension when calculating weights upon gridding. It is semi-equivalent to the scaling factors in the RoI class methods. An example of using this parameter would be setting a 'zdist_factor' value of 0.0 combined with an h_factor=0.0 (if calling DistBeamRoI) or z_factor=0.0 (if calling DistRoI), resulting in the exclusion of the z dimension in gridding weighting, which could serve as a potential solution for gridding a single PPI sweep from a single radar. In that case, the distance in the weighting function effectively serves as the distance of a given point from the PPI sweep's cone.

For example, using the latter example (h_factor=0.0 and zdist_factor=0.0), one can quickly resolve the issue in #1508 , gridding properly a single PPI sweep independent of the elevation angle, which by default (and as applicable to most cases) is manifested in the dependency on the 3D RoI value (here, setting a 2D dependence instead).

raw:
image
gridded:
image

- [X] Closes #1508 
- [X] Documentation reflects changes


pyart/map/gates_to_grid.py
```diff
@@ -30,7 +30,6 @@ def map_gates_to_grid(
     gatefilters=False,
     map_roi=True,
     weighting_function="Barnes2",
-    toa=17000.0,
```
mgrover1, line 33: Why remove the top of atmosphere constrant?

mgrover1 (commented):
One small question - also, can you please add an example in this PR to show users how to use this functionality/highlight the use-case?

pyart/map/gates_to_grid.py
isilber: That is the other contribution I wanted to separate. The toa variable was specified in this method but was not used/called even once, so I suspect that this is a residual from earlier releases, though I could be wrong.

isilber: @mgrover1, of course. That is a good idea.
I'm attaching an example notebook demonstrating the solution for this type of case (gridding of a single PPI sweep).
[zdist_exmaple.zip](https://github.com/ARM-DOE/pyart/files/14000249/zdist_exmaple.zip)

mgrover1: @isilber - can you add this as an example `.py` script? (see this as an example - https://github.com/ARM-DOE/pyart/blob/main/examples/mapping/plot_map_one_radar_to_grid.py), this will ensure it appears in the example gallery.

You can put it in the `/examples/plotting` directory, just make sure to call it something that starts with `plot_...` as that will ensure matplotlib sphinx builds it.

isilber: @mgrover1 Where should the example file be placed?

isilber:
@mgrover1, yes I understood that. I was referring to the accompanying data file

mgrover1:
@isilber - I added in the example, using the example data. Let me know what you think!

pyart/map/gates_to_grid.py
```diff
@@ -30,7 +30,6 @@ def map_gates_to_grid(
     gatefilters=False,
     map_roi=True,
     weighting_function="Barnes2",
```
mgrover1, line 32:
```suggestion
    weighting_function="Barnes2",
    toa=17000.0,
```

mgrover1: Let's keep this in. Other than that, this looks great, we can resolve this in a separate PR

mgrover1 (commented):
One minor comment - which you should be able to accept @isilber , other than that, this looks good to go!

isilber:
OK with that, but then `toa` should also be restored in `pyart/map/_gate_to_grid_map.pyx`

mgrover1 (approved):

PR #1526, opened 2024-03-07
FIX: Possible fix for upload failure for wheels.

zssherman:
When each os wheels zip is created, they are stored with macos.zip etc etc. I'm wondering if its having a hard time finding the wheels within each. I looked up some docs and looks like if there is multiple artifacts, might be able to download from each. Might be a solution...

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -94,8 +94,9 @@ jobs:
       - name: Download wheels
         uses: actions/download-artifact@v4
         with:
-          name: wheels-${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


zssherman:
@mgrover1 Yeah I wasn't too sure, i was scratching my head on this one, the failure occurs here:
```
      - name: Download wheels
        uses: actions/download-artifact@v4
        with:
          name: wheels
          name: wheels-${{ matrix.os }}
          path: dist
```

mgrover1 (approved):

mgrover1 (approved):

PR #1545, opened 2024-04-04
ENH: Dimension-dependent weighting factor flexibility in gridding

isilber:
The gridding and default RoI methods always had a Cartesian Z component weighting factor. Here, optional weighting factors for the gridding routine are added. This enhancement provides the basis for separately gridding single RHI sweeps, which will become highly useful once the radar implementation is finalized.
The code edits were tested against some data, including [existing examples](https://arm-doe.github.io/pyart/examples/mapping/plot_grid_single_sweep_ppi.html#sphx-glr-examples-mapping-plot-grid-single-sweep-ppi-py) and showed robust behavior that does not impact the default behavior.
Two related test scripts and the above-mentioned example were updated to reflect code changes.

- [x] Tests added

### DIFF
--- a/examples/mapping/plot_grid_single_sweep_ppi.py
+++ b/examples/mapping/plot_grid_single_sweep_ppi.py
@@ -81,7 +81,7 @@ plt.show()
 
 print(
     "===\n\nUsing 2D weighting by "
-    "setting h_factor and zdist_factor to 0.0, the max range looks OK now\n\n==="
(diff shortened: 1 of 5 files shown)


mgrover1:
@isilber - thanks for all your hard work here! Can you add a check/allow `h_factor` to be a single value? And if it is, broadcast that across all three dimensions? That way people's code will not break with once this is released.

Other than that, thanks for updating the tests and example to reflect these changes. Very nice work!

isilber:
@mgrover1 . This makes sense. Done. Thanks!

PR #1552, opened 2024-04-08
FIX: physically consistent RoI calculation for gridding; updated gridding input parameters to mitigate smearing

### LINKED ISSUE #1548: RoI calculation (in gridding routine) smear data too strongly, especially vs. z than vs. x and y (physically inconsistent)
### Description
The default RoI method `DistBeamRoI` currently calculates the RoI as follows:
```
roi = (self.h_factor[0] * ((z - z_offset) / 20.0) +
          sqrt((self.h_factor[1] * (y - y_offset))**2 +
                 (self.h_factor[2] * (x - x_offset))**2) *
          self.beam_factor)
```
where:
`self.beam_factor = tan(nb * bsp * PI / 180.)`
By default, `bsp=1.0` (virtual beam spacing) and `nb=1.5 ` (virtual beam width; effectively half a radar's 3-dB beam width).

**The main issue:**
This relates to the RoI calculation: Ideally, each dimension should be weighted equally, but currently, the z dimension has roughly twice the weight of the horizontal dimensions (0.05 vs. `tan(nb * bsp * PI / 180.) `= 0.026). 
This inconsistency results in more significant signal smearing in the z dim because the RoI increases faster in that dimension. To my knowledge, this approach does not have a physical basis or justification, but I could be wrong.
I understand that this was likely driven by the common use of PPIs (e.g., with NEXRAD) in order to receive a continuous gridded output, especially at large ranges, but this especially biases RHI scans that are commonly in use by ARM, and results in significant loss of information when using ARM radars.

**Other issues:**
1. The beam width calculation presumably considers a radar's beam width. The default value for `nb` is 1.5, meaning a beamwidth of 3 degrees. This is exaggerated even for NEXRAD, which has a beam width of ~1 degree (corresponding to `nb=0.5`) and of course ARM radars such as KaSACR (beam width of 0.33 degrees). This means that data are smeared inconsistent with radar characteristics, and information is lost in gridding.
2. The minimum RoI of 500 m produces too strong smearing, especially at small ranges. I understand, as mentioned previously, that this is likely to enable "capturing" more NEXRAD data points, which have a range resolution of 250 m, but again, this does not conform with other widely used precipitation and cloud radars used by the community.

We should have a solution conforming with Py-ART's primary objective (ostensibly supporting the radar community with an emphasis on ARM). Anyway, below I have an intermediate solution.

**Suggested solution (more physically consistent)**
1. RoI equation:
roi = sqrt(self.h_factor[0] * ((z - z_offset)**2 +
                self.h_factor[1] * (y - y_offset))**2 +
                self.h_factor[2] * (x - x_offset))**2) *
          self.beam_factor)
2. minimum RoI and beam width:
a. Interim solution:
 -` nb= 1. `(double NEXRAD/X-SAPR beam width; ~6 times Ka/W-SACR beam width; 1.4 times X-SACR,, 2.2 times X-SAPR)
 - `min_radius = 250.   # m`  or even better, `100 m` given that in short range beams have greater overlap and at long range, the min_radius has no influence as demonstrated below)
b. Optimal solution: radar-class based specifications

Below is an example from TRACER showing a "raw" (b1) RHI scan followed by gridding with the current default + RoI, and the suggested solution with the interim solution.    

image
Current default:
image
image
Suggested solution 
image
image

isilber:
This updated PR now includes ARM-specific min_radius option (automatically determined) to retain high-resolution information when operating on standard ARM radar data.
- [x] Closes #1548 
- [x] Tests?

### DIFF
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -97,7 +97,7 @@ def map_gates_to_grid(
 
     # set min_radius depending on whether processing ARM radars
     if min_radius is None:
-        min_radius = 250.0
+        min_radius = 500.0
         if "platform_id" in radars[0].metadata.keys():
             if np.any(
                 [
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -35,7 +35,7 @@ def map_gates_to_grid(
     constant_roi=None,
     z_factor=0.05,
     xy_factor=0.02,
-    min_radius=100.0,
+    min_radius=None,
     h_factor=(1.0, 1.0, 1.0),
     nb=1.0,
     bsp=1.0,
@@ -94,6 +94,15 @@ def map_gates_to_grid(
 
     if len(radars) == 0:
         raise ValueError("Length of radars tuple cannot be zero")
+
+    # set min_radius depending on whether processing ARM radars
+    if min_radius is None:
+        min_radius = 250.
+        if 'platform_id' in radars[0].metadata.keys():
+            if np.any([x in radars[0].metadata['platform_id'].lower()
+                       for x in ['sacr', 'sapr']]):
+                print("gridding an ARM radar; setting min_radius=100 m")
+                min_radius = 100.
 
     skip_transform = False
     if len(radars) == 1 and grid_origin_alt is None and grid_origin is None:
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -36,10 +36,10 @@ def map_gates_to_grid(
     z_factor=0.05,
     xy_factor=0.02,
     min_radius=500.0,
-    h_factor=1.0,
+    h_factor=(1.0, 1.0, 1.0),
     nb=1.5,
     bsp=1.0,
-    zdist_factor=1.0,
+    dist_factor=(1.0, 1.0, 1.0),
     **kwargs
 ):
     """
@@ -104,6 +104,12 @@ def map_gates_to_grid(
             grid_origin_alt = float(radars[0].altitude["data"])
         except TypeError:
             grid_origin_alt = np.mean(radars[0].altitude["data"])
+
+    # convert input h_factor and dist_factor from tuple or list to array
+    if isinstance(h_factor, (tuple, list)):
+        h_factor = np.array(h_factor, dtype="float32")
+    if isinstance(dist_factor, (tuple, list)):
+        dist_factor = np.array(dist_factor, dtype="float32")
 
     gatefilters = _parse_gatefilters(gatefilters, radars)
     cy_weighting_function = _detemine_cy_weighting_func(weighting_function)
@@ -175,7 +181,7 @@ def map_gates_to_grid(
             excluded_gates,
             roi_func,
             cy_weighting_function,
-            zdist_factor,
+            dist_factor,
         )
 
     # create and return the grid dictionary
--- a/pyart/map/gates_to_grid.py
+++ b/pyart/map/gates_to_grid.py
@@ -97,12 +97,15 @@ def map_gates_to_grid(
 
     # set min_radius depending on whether processing ARM radars
     if min_radius is None:
-        min_radius = 250.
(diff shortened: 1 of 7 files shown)


mgrover1:
@isilber - the tests are failing with the same radar object, which is a synthetic dataset generated within the testing module

https://github.com/ARM-DOE/pyart/blob/243e0c5269b59ffd3c68ee41174d7c452fbd2204/pyart/testing/sample_objects.py#L135

Please see the link above to find how to generate the same radar object 👍

mgrover1 (approved):
This looks good to me - thanks for the improvement here @isilber

pyart/map/grid_mapper.py
```diff
@@ -292,9 +292,9 @@ def map_to_grid(
     constant_roi=None,
     z_factor=0.05,
     xy_factor=0.02,
-    min_radius=500.0,
-    h_factor=1.0,
-    nb=1.5,
+    min_radius=None,
```
mgrover1, line 295: @isilber - why set this to None and silently fill it with 250m? I think it would be better to set the default here to 250m... consistent with the rest of the factors here, otherwise it might be confusing to the user what the default is.

pyart/map/gates_to_grid.py
```diff
@@ -95,6 +95,21 @@ def map_gates_to_grid(
...
+        try:
+            if "platform_id" in radars[0].metadata.keys():
+                if np.any(
+                    [
+                        x in radars[0].metadata["platform_id"].lower()
+                        for x in ["sacr", "sapr"]
+                    ]
+                ):
+                    min_radius = 100.0
+        except:
```
mgrover1, line 110: Do you know the specific error that would be thrown here? Ex. key not in dict?

mgrover1:
@isilber - just a couple more comments before we merge here!

pyart/map/gates_to_grid.py
isilber: 1. when metadata is not an attribute (can be fixed in the tests module)
2. when no 'platform_id' attribute in metadata

pyart/map/grid_mapper.py
isilber: @mgrover1,This touches some place for refactoring in this gridding module, especially given that `map_to_grid` is outdated and was essentially neglected for the much more updated `map_gates_to_grid`. (supported by a few Py-ART discussions I found during the process, but cannot pull at the moment) 
It is `None` because this method is most likely to be called from `grid_from_radars`. Without treating None internally, I suspect we put ourselves in a worse position for processing tractability, but I could be wrong. 
Also, each future default parameter update requires fewer places in the code to be updated as well, increasing code self-consistency.

pyart/map/gates_to_grid.py
mgrover1: So I think having an `AttributeError` check here instead of a blanket except would be useful to ensure we are catching the right warnings.

pyart/map/grid_mapper.py
mgrover1: `map_to_grid` is still supported https://github.com/search?q=repo%3AARM-DOE%2Fpyart+map_to_grid&type=pullrequests

Both `map_to_grid` and `map_gates_to_grid` take `min_radius` as arguments, and setting one to be None (with 250 meters in a try/except block) and the other 500 meters would be confusing from API standpoint - as mentioned before, it would be confusing for someone reading the docstring and not knowing what the default value is being set to.

pyart/map/gates_to_grid.py
isilber: Makes sense, @mgrover1. Done.

pyart/map/grid_mapper.py
isilber: I agree with you that setting 250 on one and 500 on the other will be confusing, which I was concerned about. 
In any case, I see your point of view and set explicit default values (I also added a note to the metadata concerning the default values when using ARM data).

mgrover1: Great! Thank you!!

PR #1562, opened 2024-04-15
changes to comp_z

### LINKED ISSUE #1559: Composite Z bugs (issue and fix)
In my original composite reflectivity code, I neglected at different sweeps can have different PRFs, which change the sensitivity and range. This created some odd artifacts like here: 

image

To get around this, I have adjusted the composite reflectivity code to include a quick np.where call to find all the sweeps with the same nyquist as the lowest sweep. This gets rid of the hard artifacts:

image

I have added a bool flag to turn this on/off (not sure why someone would want to turn this off, but its there if people do). 

I would guess in the future i should build it so you can select the specific nyquist, but for now this fix works.

dopplerchase:
Code for issue #1559, where composite reflectivity code carried artifacts through with poor quality radar data from velocity sweeps. I also made a change to the interpolator because it was flagged as depreciated (e.g., scipy.interpolate.interp2d). 

- [X] Closes #1559 

This is now open for comment from the maintainers! Happy to make changes as needed. 

Example usage: 

``compz = pyart.retrieve.composite_reflectivity(radar, field="reflectivity")`` 

image

The code is flexible, and allows you to still have the old functionality through the simple bool: 

``compz = pyart.retrieve.composite_reflectivity(radar, field="reflectivity",same_nyquist=False)``

image


zssherman (changes requested):

mgrover1:
@dopplerchase - is there anything we can do to help here?

dopplerchase:
@mgrover1 find me some more time? lol We need to think carefully about Ryan's comments noted on https://github.com/ARM-DOE/pyart/issues/1559

mgrover1:
no worries 😄 that is a good point!!

decadeneo:
I'm glad to hear that someone has encountered a similar bug. When directly using the function `pyart.retrieve.composite_reflectivity`, I found that the interpolated composite reflectivity values are all NaN. Upon diving into the source code, I was prompted to change the interpolation method. However, when attempting to modify it with alternative interpolation methods, I encountered an error indicating that the number of azimuth angles per layer in my radar data is inconsistent, thereby preventing a smooth interpolation process.

zssherman:
@mgrover1 I'm seeing the testing fail due to the depreciation of interp2d, we might need to do a PR just with the depreciation change. I can work on that.

RBhupi:
It is failing multiple test on my local machine.
Merging `main` into `composite_z_fix` commit to auto-triggers CI.

PR #1578, opened 2024-05-15
MNT: Update arm_scan_name to be optional. Also dep warning fix.

### LINKED ISSUE #1533: Resolving Scan Type Discrepancies in CfRadial RHI Data Visualization
* Py-ART version: 1.18.0
* Python version: 3.11.6
* Operating System: linux

### Description
I'm working with RHI data from a CfRadial format file. The data's `radar.sweep_mode` is 'rhi', yet the `radar.scan_type` is incorrectly marked as 'other'. This discrepancy causes errors when attempting to visualize the data using `pyart.graph.RadarDisplay(radar)`.

The issue originates from
https://github.com/ARM-DOE/pyart/blob/b54157f7e98243e914a586abae003278118492a2/pyart/io/cfradial.py#L200
and is related commit #1484 #1030.

My data's `scan_name` is 'PrecipRhi2', which fails to meet any criteria in the fallback types, resulting in the `scan_type` defaulting to 'other'.

Using scan_mode to determine scan_type seems inappropriate, as it leads to this type of error.

zssherman:
@mgrover1 My thought process for this was to have the arm_scan_name as an optional parameter as sweep_mode (from what im seeing in the cfradial document) contains the information we should be using by default. I'm leaving the option for arm_scan_name if the scan type is stored there instead. Idk if this is the best way forward for this? But figured I'd post the idea here.

- [x] Closes #1533

### DIFF
--- a/pyart/io/cfradial.py
+++ b/pyart/io/cfradial.py
@@ -56,6 +56,7 @@ def read_cfradial(
     exclude_fields=None,
     include_fields=None,
     delay_field_loading=False,
+    use_arm_scan_name=False,
     **kwargs,
 ):
     """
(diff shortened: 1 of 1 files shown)


mgrover1 (approved):
this logic seems reasonable!

PR #1586, opened 2024-05-22
ADD: Addition of above_toa_filter.

### LINKED ISSUE #442: The toa parameter is interpreted differently in the two gridding methods
The `toa` parameter control the height above which gates are excluded from the gridding algorithm.  This parameter is interpreted differently by the two gridding routines.  

In the 'map_gates_to_grid' the distance of the gate above the radar is added to the offset of the radar from the grid and compared against toa (toa is the height above grid origin).  

In 'map_to_grid' the distance of the gates above the radar is compared directly to toa (toa is the height above the radar).  

A uniform interpretation of toa should be used in both gridding methods.

zssherman:
This filter replaces toa parameter in gridding code and allows for the user to specify their own remove above toa filter by using the gatefilter for consistency between both gridding functions.

- [x] Closes #442 

### DIFF
--- a/pyart/filters/gatefilter.py
+++ b/pyart/filters/gatefilter.py
@@ -637,6 +637,16 @@ class GateFilter:
             marked = self._get_fdata(field) > value
         return self._merge(marked, op, exclude_masked)
 
+    def exclude_above_toa(
+        self, value, exclude_masked=True, op="or", inclusive=False
+    ):
+        """Exclude gates where a given field is above a given value."""
+        if inclusive:
+            marked = self._radar.gate_altitude["data"] >= value
+        else:
+            marked = self._radar.gate_altitude["data"] > value
+        return self._merge(marked, op, exclude_masked)
+
     def exclude_inside(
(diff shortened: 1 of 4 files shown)


zssherman:
@mgrover1 Question is, do we continue if gatefilter is None, remove anything above 17000.0 for gridmapper, or leave it fully up to the user

zssherman:
@mgrover1 Changes have been done for backwards compatibility, this should be good now for a final review.

zssherman:
@mgrover1 When you get a chance, this is good for a final review

mgrover1 (changes requested):
Can we keep the toa parameter in `map_gates_to_grid`, and remove when we release in 2.0?

zssherman:
@mgrover1 The only reason I removed it from the map_gates_to_grid but not grid mapper is that in map_gates_to_grid I did not see the parameter being used anywhere when I search for toa so I believe it was just there.

zssherman:
@mgrover1 Added it back in.

mgrover1:
the failing test is fixed in #1597

mgrover1 (approved):

PR #1603, opened 2024-06-25
FIX: Update comp z to remove interp2d.

zssherman:
Figured to fix this to fix the CI. While we find a solution for the comp z VCP nexrad sweep issue in #1559


zssherman:
@mgrover1 Good timing lol, closing in favor of #1604

Issue #1522, opened 2024-03-06
Incorporation of RadTraQ into Py-ART


AdamTheisen:
### Description

RadTraQ was initially developed for quality monitoring of vertically pointing radar data from the ARM program.  With the Py-ART move to xradar, it seems like it could be a good time to incorporate the remaining codes in RadTraQ into Py-ART as they are based on xarray.  There is some overlap already on the CFAD and column extraction but the remaining functions in RadTraQ could be worthwhile to pull in.

https://github.com/ARM-Development/RadTraQ

The end-goal would be to retire RadTraQ and focus future development efforts in Py-ART.

rcjackson:
@mgrover1 I will be working on integrating some of the RadTraQ features for PyART 2.0 this afternoon/tomorrow.

zssherman:
Think I accidently closed this. Still more in Radtraq that needs to be brought over.

PR #1630, opened 2024-08-29
ADD: ZDR bias code from RadTraQ

rcjackson:
- [x] Addresses part of #1522 
- [x] Tests added

I have adapted the ZDR bias code from RadTraQ into PyART with tests and the given example.


pyart/correct/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .attenuation import calculate_attenuation  # noqa
 from .attenuation import calculate_attenuation_philinear  # noqa
 from .attenuation import calculate_attenuation_zphi  # noqa
+from .bias_and_noise import correct_bias, correct_noise_rhohv, calc_zdr_offset  # noqa
```
zssherman, line 11: Looks like correct_noise_rhohv got duplicated in the conflict fix.

pyart/correct/bias_and_noise.py
```diff
@@ -144,6 +145,62 @@ def correct_bias(radar, bias=0.0, field_name=None):
...
+    gatefilter: PyART GateFilter
+        Gatefilter for filtering out data for calculating ZDR bias
+    height_range: tuple
+        The minimum and maximum heights in meters for the scan.
+    zdr_var: string or None
+        The name of the ZDR variable. Set to None to have PyART try to determine this automatically.
+
+    Returns
+    -------
+    obj : dict
```
zssherman, line 165: Can we rename obj to something else? Not really necessary, but more so i think obj as a class or pyart or xarray obj and not dict

zssherman (changes requested):
Just some small comments, besides that, looks good!

zssherman (approved):

zssherman:
Doc failure is unrelated to PR, fixed in mine.

zssherman:
@rcjackson I think we got two correct_bias now, there was one on the line below of the changed one.

mgrover1 (approved):
This looks great! Thanks for the hard work here @rcjackson @zssherman

PR #1631, opened 2024-08-29
FIX: Add local install of pyart to doc env file

mgrover1:
Add a local install of pyart to docs environment

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -41,3 +41,4 @@ dependencies:
       - pooch
       - versioneer
       - ablog
+      -e ..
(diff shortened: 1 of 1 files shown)


doc/environment.yml
```diff
@@ -41,3 +41,4 @@ dependencies:
       - pooch
       - versioneer
       - ablog
```
zssherman, line 43: should be:
- -e ..

zssherman: format was weird but double dashes with a space in between

mgrover1: fixed

mgrover1:
@zssherman - looks like tests are passing, I am going to merge to get the CI building properly

PR #1633, opened 2024-08-29
ENH: Max CAPPI

### LINKED ISSUE #597: Constant Altitude Plan Position Indicator (CAPPI)
It would be convenient to create a griddisplay class in line with the radardisplay class to be able to plot CAPPIs in km units.

In addition it may be convenient to be able to plot gridmapdisplay and radarmapdisplay in km units.

syedhamidali:
- [x] Closes #597
- [x] Tests added
- [x] Examples Added
- [x] Documentation reflects changes: Not sure

So, I have added the plot_maxcappi function to the gridmapdisplay, here is what it is supposed to do

```python
import pyart

grid = pyart.io.read_grid("GRID_KGRR20240119_163853_V06.nc")
display = pyart.graph.GridMapDisplay(grid)

display.plot_maxcappi('reflectivity', vmin=0, vmax=60, add_slogan=True)
```
image


zssherman:
I re-added back in Max as a reviewer as well as he made some change comments on the original PR, I'm reviewing as well.

pyart/graph/__init__.py
```diff
@@ -62,5 +62,7 @@
 from .radardisplay_airborne import AirborneRadarDisplay  # noqa
 from .radarmapdisplay import RadarMapDisplay  # noqa
 from .radarmapdisplay_basemap import RadarMapDisplayBasemap  # noqa
+from . import max_cappi  # noqa
```
zssherman, line 65: Is this duplicated with the line below?

pyart/graph/max_cappi.py
```diff
@@ -0,0 +1,443 @@
...
+        The map projection for the plot. Default is cartopy.crs.LambertAzimuthalEqualArea().
+    colorbar : bool, optional
+        Whether to include a colorbar in the plot. Default is True.
+    range_rings : bool, optional
+        Whether to include range rings at 50 km intervals. Default is True.
+    dpi : int, optional
+        DPI (dots per inch) for the plot. Default is 100.
+    show_progress : bool, optional
+        Whether to print progress messages. Default is False.
+    savedir : str, optional
```
zssherman, line 79: I would keep this consistent and leave it up to the use to use plt.savefig? @mgrover1 thoughts?

pyart/graph/max_cappi.py
```diff
@@ -0,0 +1,443 @@
...
+        Directory where the plot will be saved. If None, the plot is not saved.
+    show_figure : bool, optional
```
zssherman, line 81: Same here, maybe leave it up to the use for plt.show()?

pyart/graph/max_cappi.py
```diff
@@ -0,0 +1,443 @@
...
+        Whether to display the plot. Default is True.
+    **kwargs : dict, optional
+        Additional keyword arguments to pass to matplotlib's `pcolormesh` function.
+
+    Returns
+    -------
+    None
+        This function does not return any value. It generates and optionally displays/saves a plot.
+    Notes
```
zssherman, line 90: Space between Notes header and the line above.

pyart/graph/gridmapdisplay.py
```diff
@@ -1110,6 +1112,46 @@ def cartopy_coastlines(self):
             category="physical", name="coastline", scale="10m", facecolor="none"
         )
 
+    def plot_maxcappi(
```
zssherman, line 1115: More curiosity, but reason for having the code in its own file and then here as well?

pyart/graph/max_cappi.py
syedhamidali: Yeah, you are right, dropping it now.

pyart/graph/max_cappi.py
syedhamidali: Actually the it doesn't take the fig and ax arguments since its fig canvas (Main, upper, right and colorbar axes) is manually built, that is why. What do you suggest?

pyart/graph/__init__.py
syedhamidali: Yeah, I guess it is not needed.

pyart/graph/gridmapdisplay.py
syedhamidali: It was a long code, didn't wanted to overcrowd the gridmapdisplay.py script

syedhamidali: What do you suggest?

zssherman: I think it would be fine being in the gridmapdisplay, or either or, we can get @mgrover1 thoughts on it as well.

syedhamidali: I’m open to merging it back if that’s preferred. Let's see what @mgrover1 has to say!

syedhamidali:
I’ve addressed the major issues causing CI failures. The only one remaining is build-deploy-site / deploy-website. I’m not sure what that involves, so you might need to handle it.

zssherman:
@syedhamidali No worries on the build doc fail, I have a fix for it in my PR

mgrover1 (approved):
This looks great to me - I like the idea of moving a bulk of it to its own submodule, and it being used with the gridmapdisplay.

PR #1638, opened 2024-08-30
FIX: Minor fix in plot_maxcappi

syedhamidali:
This fix addresses an issue in the plot_maxcappi function where radar_name could appear either as a list of byte strings or a simple string.

### DIFF
--- a/pyart/graph/max_cappi.py
+++ b/pyart/graph/max_cappi.py
@@ -284,7 +284,6 @@ def plot_maxcappi(
         labelbottom=False,
     )
 
-    # Retrieve instrument name
     # Initialize an empty list to store the processed radar names
     full_title = []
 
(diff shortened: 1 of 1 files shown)


pyart/graph/max_cappi.py
```diff
@@ -285,13 +285,33 @@ def get_coord_or_attr(ds, coord_name, attr_name):
     )
 
     # Retrieve instrument name
-    instrument_name = ds.attrs.get("instrument_name", "N/A")[:4]
+    # instrument_name = ds.attrs.get("instrument_name", "N/A")[:4]
```
zssherman, line 288:
```suggestion
```

zssherman: Can we remove old code comment.

zssherman (changes requested):

zssherman:
@syedhamidali Thanks for the PR! One comment I added

pyart/graph/max_cappi.py
syedhamidali: Sure, please go ahead.

PR #1639, opened 2024-09-02
FIX: Minor fix in plot_maxcappi

syedhamidali:
This fix addresses an issue in the plot_maxcappi function where radar_name could appear either as a list of byte strings or a simple string.


syedhamidali:
@mgrover1 It already has a test. Or am I missing anything?

zssherman (approved):

PR #1643, opened 2024-09-04
FIX: Incorrect field plotted in cloud mask example

zssherman:
Fixes the incorrect field being plotted in the cloud mask example


mgrover1 (approved):
Good catch! Looks good to me

PR #1660, opened 2024-10-08
TST: Update test for windows builds, remove try statement for geotiff.

zssherman:


zssherman:
@mgrover1 This should be good for review now. Windows 3.9 is stalled, but I believe that is fixed in the other PR.

mgrover1 (approved):
Thanks for putting this together - this looks good to me!

PR #1663, opened 2024-10-08
MNT: Add warnings for IO readers for 2.0 to push the use of xradar.

zssherman:


mgrover1 (approved):
We will need to add a follow-up PR to the user guide explaining this more in detail, but the warnings here look good!

mgrover1 (approved):

PR #1681, opened 2024-11-05
FIX: Ensure backwards compatibility with datatree/new versions of xarray

juhi24:
xarray-datatree has been merged to xarray>=2024.10.0. Since xarray version is not pinned, this will cause ImportErrors unless addressed. This is PR is an attempt towards a phased transition to using `xarray.datatree`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12"]
         os: [macos, ubuntu, windows]
 
     steps:
--- a/continuous_integration/environment-ci.yml
+++ b/continuous_integration/environment-ci.yml
@@ -12,7 +12,7 @@ dependencies:
   - wradlib
   - cartopy
   - cvxopt
-  - xarray<=2024.9.0
+  - xarray>=2024.10.0
   - metpy
   - pytest-cov
   - pytest-mpl
@@ -25,7 +25,7 @@ dependencies:
   - shapely
   - ruff
   - mda-xdrlib
-  - xradar<=0.7.0
+  - xradar>=0.8.0
   - pooch
   - versioneer
   - black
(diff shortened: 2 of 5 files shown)


neutrinoceros (commented):
This looks like a step in the right direction, however if neither `xarray-datatree` nor `datatree` is a hard dependency to `pyart`, then there should probably be an abstract way to specify which flavor users want (e.g. `pip install arm-pyart[xarray-datatree]`, `pip install arm-pyart[datatree]`). This would preserve downstream from having to specify the exact requirements themselves, though it doesn't address the fact that the default install is broken.
I would recommend having `xarray-datatree` as a hard dependency and be done with it, but if that is not possible I would still recommend crafting a custom error message to explain the situation when *both* import fail

juhi24:
You probably meant packages `xarray` and `xarray-datatree`, providing modules `xarray.datatree` and `datatree`, respectively. Package called `datatree` is unrelated. The package `xarray-datatree` is deprecated in favour of `xarray`, which now includes the datatree functionality. I would recommend against providing a flavour based on a package that is no longer maintained.

Currently, `pyart` seems to have a hard dependency on (the latest) `xradar` which, in turn, depends on `xarray >= 2024.10.0`. Therefore, `pyart` should either use `xarray.datatree` or pin to earlier versions of `xradar` and `xarray`. Otherwise, the default installation will remain broken.

kmuehlbauer:
From my perspective it would be reasonable to:

1. pin xradar<=0.7.0 and xarray<=2024.9.0 and do a release
2. fix anything which has to be fixed
3. pin xradar>=0.8.0 and xarray>=2024.10.0 and do a another release

That's one way to not break users existing code. Users might then decide which version to use. 

The old xarray-datatree and the new datatree from xarray differ in quite some places and would need users to adapt their code (depending on what they've imported from there). I'm not that familiar with pyart's usage of xradar, but good chance that correct pinning will already do the trick.

mgrover1:
Thank for the PR here @juhi24 !! I have opened another PR (#1682) which should help toward fixing this issue. I think we should:
- Merge that one
- Cut a new release
- Fix the failing CI on this PR, and merge this
- Release again with the updated xarray/xradar releases

mgrover1 (approved):

PR #1696, opened 2024-11-22
FIX: Update to use f-format strings. Remove glpk version check until we have a working version

zssherman:
Removed glpk version check as well as it wasn't doing anything, can be replaced by pyglpk once there is a working version.


mgrover1 (approved):

zssherman:
@mgrover1 Ill dig into this more monday, I forgot that I have to check if the format is a decimal string or integer, which is breaking the mdv reader

mgrover1 (approved):

PR #1704, opened 2024-12-10
ENH: `read_mmcr` method to read ARM MMCR b1-level data files

isilber:
A method to robustly read ARM MMCR b1-level data files (the datastream available through the ARM Archive).
The method returns a `dict` with keys representing each of the operated MMCR modes (available in the data files) and values representing the corresponding `Radar` objects. An optional input parameter (`mode_to_extract`) enables only extracting a single mode, in which case a single `Radar` object is returned.

The method is written in a similar format to the `read_kazr` method (which has some clutter in it) under the same `aux_io` module.
That module lacks unit tests (as also noted in its `__init__.py` script) and therefore I didn't include unit tests.

- [ x] Documentation reflects changes

### DIFF
--- a/pyart/aux_io/arm_vpt.py
+++ b/pyart/aux_io/arm_vpt.py
@@ -271,14 +271,14 @@ def read_mmcr(
         values the corresponding `Radar` object.
         If `mode_to_extract` is specified returning the `Radar` object corresponding to the that
         mode.
-        
+
     """
 
     if mode_names is None:
         mode_names = {
-            1: {"name": "bl", "prt": 1.0 / 68e-6},                  
-            2: {"name": "ci", "prt": 1.0 / 126e-6},                  
-            3: {"name": "ge", "prt": 1.0 / 106e-6},                  
+            1: {"name": "bl", "prt": 1.0 / 68e-6},
+            2: {"name": "ci", "prt": 1.0 / 126e-6},
+            3: {"name": "ge", "prt": 1.0 / 106e-6},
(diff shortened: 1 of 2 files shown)


pyart/aux_io/arm_vpt.py
```diff
@@ -218,3 +218,269 @@ def read_kazr(
...
+    ----------
+    filename : str
+        Name of NetCDF file to read data from.
+    field_names : dict, optional
+        Dictionary mapping field names in the file names to radar field names.
+        Unlike other read functions, fields not in this dictionary or having a
+        value of None are still included in the radar.fields dictionary, to
+        exclude them use the `exclude_fields` parameter. Fields which are
+        mapped by this dictionary will be renamed from key to value.
+    additional_metadata : dict of dicts, optional
```
zssherman, line 246: I would be fine with removing this here and in the function above and just setting the call below to None if not used.

zssherman (commented):
I left one comment, besides that looks good. @mgrover1 can give this a review as well, when you have time?

pyart/aux_io/arm_vpt.py
zssherman: Yeah we can remove both in a future PR.

isilber:
I just added options to read dual-pol channels' MMCR data (limited number of data files).
I also changed the PRT specifications to be read from the loaded file instead of hardwired handbook values

mgrover1 (approved):
This looks great! Thanks @isilber for putting this together

zssherman:
Nope, merge away! Thanks again @isilber !

PR #1707, opened 2024-12-13
FIX: Update sigmet reader to handle 1byte kdp, that needs to be divided by wavelength.

zssherman:
See discussion here:
https://github.com/ARM-DOE/pyart/discussions/1706

### DIFF
--- a/pyart/io/_sigmetfile.pyx
+++ b/pyart/io/_sigmetfile.pyx
@@ -214,7 +214,7 @@ cdef class SigmetFile:
             nyquist = wavelength_cm / (10000.0 * 4.0 * prt_value)
             data['WIDTH'] *= nyquist
             # scale 1-byte KDP by the wavelength
-	    if 'KDP' in self.data_type_names:
+	if 'KDP' in self.data_type_names:
             # The IRIS Programmer's Manual indicates 1-byte differential phase format
             # data should be divided by the wavelength in cm (section 4.3.12).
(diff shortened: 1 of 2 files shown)


pyart/io/_sigmetfile.pyx
```diff
@@ -213,7 +213,12 @@ cdef class SigmetFile:
             prt_value = 1. / self.product_hdr['product_end']['prf']
             nyquist = wavelength_cm / (10000.0 * 4.0 * prt_value)
             data['WIDTH'] *= nyquist
-
+            # scale 1-byte KDP by the wavelength
+	    if 'KDP' in self.data_type_names:
```
zssherman, line 217:
```suggestion
	if 'KDP' in self.data_type_names:
```

pyart/io/_sigmetfile.pyx
```diff
@@ -213,7 +213,12 @@ cdef class SigmetFile:
             prt_value = 1. / self.product_hdr['product_end']['prf']
             nyquist = wavelength_cm / (10000.0 * 4.0 * prt_value)
             data['WIDTH'] *= nyquist
-
+            # scale 1-byte KDP by the wavelength
```
zssherman, line 216:
```suggestion
        # scale 1-byte KDP by the wavelength
```

mgrover1 (approved):
This looks great to me - thanks for the fix here @zssherman

PR #1715, opened 2024-12-29
add lonboard example

### LINKED ISSUE #1653: DOC: lonboard example
I have recently become acquainted with the [lonboard](https://github.com/developmentseed/lonboard) plotting library and successfully plotted radar data using it, with the goal of throwing together a little interactive interface using [panel](https://panel.holoviz.org/).

However, the only functions of pyart used are `pyart.io.read` and `pyart.core.Radar.get_gate_lat_lon_alt` which are both well-documented at this point.

I was curious if y'all thought an example involving lonboard/panel would be appropriate here or not.

thanks!

wx4stg:
I have created a jupyter notebook that follows the "plot nexrad reflectivity" example, but uses lonboard as the plotting library rather than matplotlib, and documents the differences in plotting with lonboard. lonboard requires a notebook environment to plot (or at least, something that can display HTML, a [panel](https://panel.holoviz.org) server would work too), I've included an `ipynb` file. This bucks the trend as most other examples are .py files, so if this is unacceptable please let me know in the discussion and we can work out the best way to include this.

Additionally, there are a few dependent libraries that are required for this to work. lonboard, as well as arro3 which handles the Arrow table data structures used by lonboard. I have no idea what this means for CI but I suspect everything will fail, at least for the initial commit.

Thanks to @kylebarron for ELI5'ing/fixing up an example to help me learn apache arrow tables, work on lonboard, and for the excellent talk at SciPy 2024 that @mgrover1 and I attended and were inspired by. 

- [ ] Closes #1653


mgrover1 (changes requested):
Thanks for the contribution here @wx4stg !! One slight change - can you organize the imports in alphabetical order, without any lines in between? The linter is picking up on these

kylebarron:
> as well as arro3 which handles

In terms of your input data, you can also use pyarrow to construct the data structures if you prefer, but lonboard will use arro3 internally.

wx4stg:
I think that should fix it...

mgrover1 (changes requested):
I think this would be better as a .py script here... happy to help convert the code here!

wx4stg:
Ha, I completely forgot about this. It's been a semester since I've thought about lonboard but I need to re-familiarize myself with it for some personal projects so I'll take a look (maybe this weekend) if y'all don't beat me to it.

RBhupi:
@wx4stg Can you make changes requested in the following comment?

> I think this would be better as a .py script here... happy to help convert the code here!

RBhupi (approved):
@zssherman Please do as discuses. 

Keep it a notebook and move it to the dedicated notebook folder.

PR #1755, opened 2025-05-21
FIX: Fixes mismatch shape issue.

zssherman:
Fixes issue of mismatch shape x and y for cfad. If no mask is provided, will check for existing mask in field data and apply it to the altitude data, similar to if a mask is provided by the user.


zssherman:
@mgrover1 Yep! Sorry I kinda forgot about this one

zssherman:
@mgrover1 test has been added

Issue #1531, opened 2024-03-19
Get the beam width using the aux_io.read_sinarame_h5 function 


evmpython:
* Py-ART version: 1.18.0
* Python version: 3.10.13
* Operating System: Ubuntu 22.04.3 LTS

### Description
Hi all. I am trying extract the beam width from HDF5 from Braziliam radar using the function “aux_io.read_sinarame_h5”. But the “instruments_parameters” is empty.  I am using Google Colab. 

When the files is open with hdfview software, the radar file have this information (see below). How Can I get the  beam width in this file? Follow one HDF5 radar file as example. 

[2024021622400300dBZ.vol.zip](https://github.com/ARM-DOE/pyart/files/14652868/2024021622400300dBZ.vol.zip)

### What I Did

!pip install -q condacolab
import condacolab
condacolab.install()
!mamba install -q -c conda-forge arm_pyart

import pyart  
filename = '2024021622400300dBZ.vol.h5'
radar = pyart.aux_io.read_sinarame_h5(f'/content/{filename}')
radar.instrument_parameters.keys()

AttributeError                            Traceback (most recent call last)
[<ipython-input-1-9d68e307ffea>](https://localhost:8080/#) in <cell line: 9>()
      7 radar = pyart.aux_io.read_sinarame_h5(f'/content/{filename}')
      8 
----> 9 radar.instrument_parameters.keys()

AttributeError: 'NoneType' object has no attribute 'keys'
![WhatsApp Image 2024-03-18 at 20 09 26](https://github.com/ARM-DOE/pyart/assets/100850499/92792971-a028-4354-a948-04734478d7d6)

zssherman:
@evmpython The sinararme reading in py-art at the moment is in aux_io, so intrument parameters defaults to none as it does not currently have a method in the code that pulls it. These files i'm familiar with, but if its store in some way in h5py, we can always add that in.

evmpython:
Also i have done test with pyart.aux_io.read_odim_h5. But the same error persist.

mrugna:
Sorry for reopening this issue :)

I created a fresh install using 1.18.7 and found this modification so I'll try to bring a little bit of context.

This sinarame function was made a long time ago as a separate version of the ODIM H5 (aux) reader because it was meant to support the files created from argentine radar hardware ("sinarame" kind of stands for Argentina national radar network). This had to be done in a separate function because of some missing parameters in the H5 file which is not complete or somewhat compliant.

[...]

PR #1766, opened 2025-06-06
FIX: Corrected instrument parameters check for sinarame H5 files.

mrugna:
This fixes the error pointed at #1531 that appears when reading SINARAME H5 files.


zssherman (approved):

zssherman:
Thanks for the PR @mrugna , this looks good to me. I'll let @mgrover1 give it a review as well.

PR #1810, opened 2025-11-22
WHL: upgrade cibw, migrate static configuration to pyproject.toml, enable cp313 wheels

neutrinoceros:
I noticed that the latest version didn't have wheels for Python 3.13 or 3.14, so I did a bit of housecleaning to enable them.
The reason I'm also making an attempt at enabling musllinux wheels is that the documented reason for disabling them is outdated: numpy has been supporting this target for some time now.
Out of scope here, but note that Cython>=3.1 is now able to generate Limited API-compliant C code, which means it should be relatively easy to both improve the future-portability of your wheels and reduce the overall number of wheels you need to produce for each release. I'd be happy to contribute this too if desired.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -24,8 +24,8 @@ jobs:
         os: [
           ubuntu-latest,
           windows-latest,
-          macos-14,
-          macos-13,
+          macos-latest,
+          macos-15-intel,
         ]
       fail-fast: false
 


.github/workflows/wheels.yml
```diff
@@ -28,8 +24,8 @@ jobs:
         os: [
           ubuntu-latest,
           windows-latest,
-          macos-14,
-          macos-13,
+          macos-latest,
+          macos-15-intel,
```
neutrinoceros: the important bit here is to move away from `macos-13` as the image is deprecated (and possibly already discontinued). While I'm at it, I also bumped the amr64 image to match how ubuntu and windows are specified.

pyproject.toml
```diff
@@ -40,3 +40,20 @@ select = [
     # Pyupgrade
     "UP",
 ]
+
+[tool.cibuildwheel]
+build-verbosity = 1
+skip = [
+    "cp38-*",
+    "cp39-*",
+    "cp314t-*",
```
neutrinoceros: given that both 3.13 and 3.14 wheels were missing I assumed that free-threading wasn't supported yet. Note that cibuildwheel 3.2 and newer just enable it by default for 3.14, not 3.13 (because it was marked as experimental for that version), so, to be conservative, and because it's potentially much more work to add, I just disabled this one explicitly.

pyproject.toml
```diff
@@ -40,3 +40,20 @@ select = [
     # Pyupgrade
     "UP",
 ]
+
+[tool.cibuildwheel]
+build-verbosity = 1
+skip = [
+    "cp38-*",
+    "cp39-*",
```
neutrinoceros: note that it shouldn't be necessary to list versions you *do* (`build`) or *do not* (`skip`) support here; the modern approach is to rely soly on `project.requires-python`, but the `project` table is entirely missing here (the project metadata is instead 100% dynamically generated at build time via `setup.py`). While out of scope here, I could easily contribute a bit of modernization in that area too.

zssherman:
Hi @neutrinoceros, thank you for the PR! A few of us are out for the holiday, but can give this a review soon! We greatly appreciate this PR to update the wheels etc!

neutrinoceros:
Absolutely no rush on my end, enjoy your holiday !

.github/workflows/wheels.yml
```diff
@@ -45,17 +41,9 @@ jobs:
...
       - name: Build wheels for CPython
-        uses: pypa/cibuildwheel@v3.2.1
+        uses: pypa/cibuildwheel@v3.3.0
         with:
           output-dir: dist
-        env:
-          CIBW_BUILD: "cp310-* cp311-* cp312*"
-          CIBW_SKIP: "*-musllinux_*"  #  numpy doesn't have wheels for musllinux so we can't build some quickly and without bloating
-          CIBW_ARCHS_LINUX: "x86_64"
-          MACOSX_DEPLOYMENT_TARGET: "10.9" # as of CIBW 2.9, this is the default value, pin it so it can't be bumped silently
```
neutrinoceros: note that I intentionally didn't replicate this line because the *correct* target depends on the Python version being built against, and it's much easier to rely on cibuildwheel's default for this.

zssherman (approved):
Looks good to me, thanks @neutrinoceros !

PR #1816, opened 2025-12-23
FIX: Fix to exclude_fields to still work when include_fields is None.

rcjackson:
This is a minor fix that will fix the error that is thrown when include_fields is None and exclude_fields is not None. With this behavior, I assume that we are including every field that is not in exclude_fields (like the documentation states is the assumption with include_fields) for pyart.io.read_cfradial.


zssherman (approved):

zssherman:
Good catch @rcjackson , merging!

Issue #1820, opened 2026-01-12
MDV grid writer crashes on broken datetime tz-awareness


milly-troller:
I'm trying to write a grid to MDV file as

```
pyart.io.write_grid_mdv(test_file_fname, grid)
```

which crashes with the following;

```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], [line 2](vscode-notebook-cell:?execution_count=7&line=2)
      1 test_file_fname = 'l3_test.mdv'
----> [2](vscode-notebook-cell:?execution_count=7&line=2) pyart.io.write_grid_mdv(test_file_fname, grid)

File /opt/conda/envs/202601-l3ai/lib/python3.11/site-packages/pyart/io/mdv_grid.py:111, in write_grid_mdv(filename, grid, mdv_field_names, field_write_order)
    109 d["max_ny"] = ny
    110 d["max_nz"] = nz
--> 111 td = datetime.datetime.now(datetime.timezone.utc) - datetime.datetime(
    112     1970, 1, 1, 0, 0
    113 )
    114 d["time_written"] = int(
    115     round(td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / 10**6
    116 )
    118 # sensor location using radar_ attribute or origin_ if not available

TypeError: can't subtract offset-naive and offset-aware datetimes
```

and the longer I look at the lines [in main](https://github.com/ARM-DOE/pyart/blob/main/pyart/io/mdv_grid.py#L111-L113);

```
    td = datetime.datetime.now(datetime.timezone.utc) - datetime.datetime(
        1970, 1, 1, 0, 0
    )
```

the more confident I am the data I'm trying to save is completely irrelevant; this line, alone, uses no inputs and crashes on its own; it's plainly evident that it's subtracting a TZ-unaware date from a TZ-aware one.

The code from before [#1556](https://github.com/ARM-DOE/pyart/pull/1556) probably worked in Python 2? But I don't see how the new version ever could have worked in either Python 2, which afaik lacks `datetime.timezone.utc` or Python 3, where the timezone awareness mismatch should always crash?

Changing the line to

```
    td = datetime.datetime.now(datetime.timezone.utc) - datetime.datetime(
        1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc
    )
```

seems correct, and fixes the issue in my case.

zssherman:
[...]

PR #1825, opened 2026-01-14
Fix timezone handling in datetime calculation

milly-troller:
Fixes otherwise inevitable crash, as per https://github.com/ARM-DOE/pyart/issues/1820

Shouldn't affect any changes in normal operation, because without the fix, the function in unusable. It also doesn't involve any data coming into the function, just local, current time and epoch 0.


zssherman (approved):

zssherman:
@milly-troller Thanks for the PR! Running the tests now

zssherman:
@milly-troller Tests are passing! Merging, thanks for putting the fix in!

PR #1846, opened 2026-04-10
DOC: Adding correct reference to ARM 

jrobrien91:
- updating docs and message to users upon import correcting the ARM reference


zssherman (approved):

zssherman:
@jrobrien91 Good catch, thank you! I'll merge this once the tests pass

PR #1853, opened 2026-05-13
ENH: Update radar object for optional airborne variables.

zssherman:
Added in missing variables from airbourne datasets such as pitch_change_rate and eastward_wind


rcjackson:
Can we update the unit tests accordingly with a mock airborne radar object?

rcjackson (changes requested):
Could we also update the unit tests to cover the addition of these new variables?

zssherman:
@rcjackson Yep!

rcjackson (approved):

PR #1861, opened 2026-06-11
ENH: Update to trusted publisher

zssherman:


zssherman:
@rcjackson Do you have a second for a quick review?

rcjackson:
Looks good! Fewer stored tokens is always better.

PR #1870, opened 2026-07-10
CI: Remove python 3.11 and add 3.14

zssherman:


syedhamidali:
I think cartopy isn't available for 3.14 yet

zssherman:
@syedhamidali There is a build for python 3.14 on their feedstock, though their last release was before then, it is having issues with codecov for python 3.14. I am still able to generate the map background with python 3.14

RBhupi:
Cartopy is testing for python-version '3.14' in their CI/CD. Their build is failing for some other reason.

tests/graph/test_gridmapdisplay.py
```diff
@@ -5,13 +5,20 @@
...
 # in baseline_images directory. Currently this test only determines if files
 # can be created, not that they are correct.
+import sys
 
 import matplotlib.pyplot as plt
 import pytest
 
 import pyart
 
+skip_cartopy_py314 = pytest.mark.skipif(
```
RBhupi, line 15: Maybe revisit skips once Cartopy has resolved the Python 3.14 issues.

RBhupi (approved):
Looks good to me. 
Thanks!

PR #1879, opened 2026-08-07
Fix get_nexrad_location mutating shared NEXRAD_LOCATIONS table

Yuxiang-Ren-HUB:
### Problem

`get_nexrad_location()` converts the station's elevation from feet to
meters by writing the result back into the dict it just read from:

    loc = NEXRAD_LOCATIONS[station.upper()]
    loc["elev"] = loc["elev"] * 0.3048
    return loc["lat"], loc["lon"], loc["elev"]

`loc` is not a copy -- it's a reference into the shared, module-level
`NEXRAD_LOCATIONS` dict, so the conversion is applied permanently.
Calling `get_nexrad_location()` a second time for the same station
applies the conversion again, silently corrupting the result:

    >>> get_nexrad_location("KTLX")[2]
    369.7224          # 1213 * 0.3048, correct
    >>> get_nexrad_location("KTLX")[2]
    112.69138752      # converted twice, wrong

Any code path that calls this function more than once for the same
station -- e.g. processing multiple files/sweeps from the same radar
in one process -- silently gets a wrong altitude after the first call.

### Fix

Store the converted elevation in a local variable and return that,
instead of writing back into the shared table. `NEXRAD_LOCATIONS` is
now never mutated by this function.

### Testing

Added `tests/io/test_nexrad_common.py` (previously no dedicated test
file existed for this module):
- `test_get_nexrad_location_known_station`: sanity-checks the returned
  lat/lon/elev for a known station.
- `test_get_nexrad_location_repeated_calls_are_stable`: regression test
  for this bug -- calls the function 3 times for the same station and
  asserts the elevation doesn't change, and that the underlying table
  entry stays at its original (feet) value.

Both pass locally.

### DIFF
--- a/pyart/io/nexrad_common.py
+++ b/pyart/io/nexrad_common.py
@@ -26,18 +26,20 @@ def get_nexrad_location(station):
     """
     loc = NEXRAD_LOCATIONS[station.upper()]
 
-    # Convert from feet to meters for elevation units. Read into a local
-    # variable and return a new value rather than writing back into
-    # loc["elev"] -- loc is a reference into the shared, module-level
-    # NEXRAD_LOCATIONS dict, so mutating it in place applied this
-    # conversion permanently. A second call for the same station then
-    # converted the already-converted value again, silently corrupting
-    # the elevation (e.g. KTLX: 1213 ft -> 369.72 m on the first call,
-    # then -> 112.7 m on the second).
-    elev_m = loc["elev"] * 0.3048
+    # The bundled table stores elevation in feet.  This function has always
+    # converted the entry in place, leaving the shared, module-level
+    # NEXRAD_LOCATIONS dict holding meters; that side effect is kept here
+    # because downstream code may read the table directly after calling this
+    # function.  The entry is now tagged with its units so the conversion is
+    # applied at most once.  Previously a second call for the same station
+    # re-applied it and silently corrupted the value (KTLX: 1213 ft ->
+    # 369.72 m on the first call, then -> 112.69 m on the second).
+    if loc.get("units") != "m":
+        loc["elev"] = loc["elev"] * 0.3048
+        loc["units"] = "m"
 
-    return loc["lat"], loc["lon"], elev_m
-
+    return loc["lat"], loc["lon"], loc["elev"]
+    
 
 # Locations of NEXRAD locations was retrieved from NOAA's
 # Historical Observing Metadata Repository (HOMR) on
--- a/pyart/io/nexrad_common.py
+++ b/pyart/io/nexrad_common.py
@@ -38,6 +38,7 @@ def get_nexrad_location(station):
 
     return loc["lat"], loc["lon"], elev_m
 
+
 # Locations of NEXRAD locations was retrieved from NOAA's
 # Historical Observing Metadata Repository (HOMR) on
 # 2014-Mar-27. http://www.ncdc.noaa.gov/homr/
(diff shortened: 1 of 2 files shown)


Yuxiang-Ren-HUB:
Friendly ping on this one — it looks like no CI has run yet. I believe workflows need a maintainer to approve them for a first-time contributor.
Happy to rebase or adjust anything if that would help. Thanks!

scollis:
Hey! Patience please! We are updating our contributors guide right now. We are also considering requiring submitters to know who they are as purely AI generated PRs will not be allowed. A real person must review code. We're not saying this is a pure AI generated PR but you do not have much of a GitHub history (as you said) and we need to ensure you are not an agent. If you could let us know who you are in real life that would speed things up.

scollis:
Great! And greetings! And thanks for the AI disclosure. We'll get to this soon.

Yuxiang-Ren-HUB:
Thanks for the reply, and for taking the time to check.
I'm a graduate student in atmospheric science in China and I use Py-ART
in my own radar work, which is how I ran into this. get_nexrad_location
was writing the converted elevation back into the shared NEXRAD_LOCATIONS
dict, so a second call for the same station came back wrong (KTLX:
369.72 m, then 112.69 m). I reproduced it, wrote the fix and the tests,
and ran them locally before opening the PR.
I'm a real person, not an agent. I did use an AI assistant while writing
the patch, but the bug, the fix and the tests are ones I checked myself.

Yuxiang-Ren-HUB:
Thank you, much appreciated.

pyart/io/nexrad_common.py
```diff
@@ -26,10 +26,17 @@ def get_nexrad_location(station):
...
-    loc["elev"] = loc["elev"] * 0.3048
+    # Convert from feet to meters for elevation units. Read into a local
+    # variable and return a new value rather than writing back into
+    # loc["elev"] -- loc is a reference into the shared, module-level
+    # NEXRAD_LOCATIONS dict, so mutating it in place applied this
+    # conversion permanently. A second call for the same station then
+    # converted the already-converted value again, silently corrupting
+    # the elevation (e.g. KTLX: 1213 ft -> 369.72 m on the first call,
+    # then -> 112.7 m on the second).
+    elev_m = loc["elev"] * 0.3048
```
RBhupi: I think the behavior  `table values changes` should be preserved as some code in the wild may be relying on it. We may add units to know if the height is already changed in meters or not. If yes then conversion will return the values but will not change it back in the table again. This will preserved the behavior.

Yuxiang-Ren-HUB:
Thanks @RBhupi — good point, and I hadn't considered the compatibility angle. Keeping the in-place conversion is clearly the safer choice, so I've reworked it that way: the conversion is now guarded by a `units` tag so it runs at most once, and the `return` line is back to the original. Tests updated to match, with a fixture that restores KTLX so the suite doesn't become order-dependent.
One thing I'd like your call on: I tag lazily, so the entry gains a `units` key at runtime and the dict shape changes after the first lookup. The alternative is adding `"units": "ft"` to every entry in the table up front — constant shape, but a much bigger diff. Happy to switch if you prefer that.
Thanks again for taking the time to look at this, I appreciate it.

Yuxiang-Ren-HUB:
One thing I'm unsure about on my end … I wasn't sure whether that's something I should be handling differently, or whether it needs a maintainer to enable. … I just wanted to check I hadn't left something undone on my side.

zssherman:
Hi @Yuxiang-Ren-HUB , apologies! And no worries on your end in regards of code changes etc! We are just finishing up our contributor AI usage policy, once we get that merged in, we will give this another review and be good from there!

Cheers

Yuxiang-Ren-HUB:
@zssherman，Thanks so much, really appreciate it!

